PR #4, opened 2016-08-03
Mount docker postgres volume on host

mihaibogdan10:
I think this should be part of the default settings. Otherwise, taking sentry down with
```
docker-compose down
```
means we just lost all the data in postgres, including users data and events gathered so far.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/onpremise/4)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,8 @@ services:
 
   postgres:
     image: postgres:9.5
+    volumes:
+      - ./data/postgres:/var/lib/postgresql/data
 
   web:
     extends: base
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
       SENTRY_POSTGRES_HOST: postgres
       SENTRY_EMAIL_HOST: smtp
     volumes:
-      - ./data:/var/lib/sentry/files
+      - ./data/sentry:/var/lib/sentry/files
 
   smtp:
     image: tianon/exim4


mattrobenolt:
I guess I expected to leave this as an exercise for the reader.

Would you want to change the default mount for the other images then too? Since this will conflict with `./data` being mounted for all sentry containers.

Maybe we do, say `./data/sentry` for sentry, and `./data/postgres`, etc for other services. That way each is partitioned away.

I also never expected someone to just run this `docker-compose` file without changing to suit their needs. :)

mihaibogdan10:
@mattrobenolt That's fair. In my case, it just slipped my mind that I should mount postgres, and went ahead and started sentry, created accounts for a few members of my team, etc.
Then I had to take it down to change the smtp settings to use our AWS one, at which point I lost all the data.

I'm just saying it's easy to forget to mount that volume and do some config work through the UI.

mattrobenolt:
Awesome, thanks!

Side note, do you know how to make this work with docker volumes instead? I think that'd be a more recommended method, but I've never used them with composer.

mihaibogdan10:
I'm not sure what you mean by that. Docker-compose is simply used to define multiple docker containers and the way they link together. The volumes defined in **docker-compose.yml** translate into definitions for docker volumes (special folders that get shared between the container and the host).

TL;DR We're already using docker volumes, we're just using them indirectly through docker-compose.

mattrobenolt:
Sorry, I mean actual data volumes such that you get from `$ docker volume create <name>`.

mihaibogdan10:
This is from the [docker-compose docs](https://docs.docker.com/compose/compose-file/): 

> Likewise, network and volume definitions are analogous to docker network create and **docker volume create**

PR #5, opened 2016-08-05
docker-compose usage

maguowei:
1. [variable-substitution](https://docs.docker.com/compose/compose-file/#variable-substitution)
2. use depends_on, remove links.It seems links no need. ref: https://docs.docker.com/compose/link-env-deprecated/

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/onpremise/5)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,20 @@ Official bootstrap for running your own [Sentry](https://getsentry.com/) with [D
  * Docker 1.10.0+
  * Compose 1.6.0+ _(optional)_
 
+## Usage
+
+```bash
+# generate secret key
+$ docker-compose run --rm web config generate-secret-key
+$ export SENTRY_SECRET_KEY='your-secret-key'
+
+# init or upgrade
+$ docker-compose run --rm web upgrade
+
+# run in the background
+$ docker-compose up -d
+```
+
 ## Resources
 
  * [Documentation](https://docs.getsentry.com/on-premise/server/installation/docker/)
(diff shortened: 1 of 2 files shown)


Bessonov:
@mattrobenolt 

Just merged this PR in master locally and tested. Works for me.

Bessonov:
@mattrobenolt 
That is not right step. It's OK to write "This docker-compose.yml is meant to be just an example", regardless the repo is described as "Sentry On-Premise setup". Not "Example for Sentry On-Premise setup". It's OK to write "Official bootstrap for running your own Sentry with Docker." and handle it as stepdaughter. But don't provide a setup to run it without modifications and close a PR to get that is very frustrating.

mattrobenolt:
docker-compose in general is a very opinionated thing. No docker-compose file is going to work for everyone in every situation. The whole point of this is to help form a foundation, not a one-click install and boom, everything is magic. Our official documentation does not mention docker-compose and we don't recommend it. We don't use it ourselves and I personally have very limited experience with the tool to even begin to give proper recommendations here. Any of the instructions we've provided for native docker can very easily be rolled into you coming up with your own docker-compose file if that's a deployment method you choose.

If you stand back, to me, including docker-compose as a method is equivalent to us also providing a file to run on Mesos/Marathon and Nomad, and instructions for K8, Google Container Service, Carina, etc. It's a method for deployment of the whole stack. Your stack is going to be very different from anyone else's.

Issue #2, opened 2016-07-27
docker-compose ?


JeanFred:
Hello,

I see there is a `docker-compose.yml`. Using docker-compose to run Sentry sounds great ; however it is not mentioned at all in the documentation at https://docs.getsentry.com/on-premise/server/installation/docker/.

Is docker-compose support just experimental, or would you recommend it for new installs?

Thanks!

Freundschaft:
docker-compose up worked for me but before that you need to run
docker-compose run --rm web upgrade in order to create db and user

hernandev:
running it on docker cloud would also be great, running without Dockerfile

mattrobenolt:
It's kinda experimental. I don't fully advocate it since I don't personally know it well enough and last I heard, docker-compose was not intended to be used for production, just small dev environments.

Can it be used now to coordinate large multi machine deployments?

hernandev:
docker cloud, formerly tutum is using it's own sintax but it's 99% compatible with docker-compose, rancher uses docker-compose along with a additional file, so looks like the standard now, anyway, rancher provides a catalog entry for sentry, so it's just one click install, and i'm not using docker-cloud anymore, altough i did got it working on docker-cloud, just adjusting a few things and always keeping the data volumes so when containers are rebuilt data is not lost

maguowei:
add docker-compose usage #5

fredvd:
The sentry 'on premise' documentation is quite confusing at the moment. Running everything under python is deprecated according to the official docs, but the docker documentation is only listing manual steps to individually start and link containers. 

with 'on premise' I understand running something to use it, so some degree of production setup.  the quickest way to production is using the docker-compose.yml, but if it's still experimental it should be clearly marked as such. Also see #9

InAnimaTe:
I'm actually making some changes to the README for this repo and will submit a PR in the next day or two. I'm going to make it clear that this is a somewhat experimental and explain the manual steps needed before a `docker-compose up` will work as expected. We've been running this for a the past few months so it isn't exactly `not production` ready, it just needs a little more explanation (and love ;)

As for the on-prem docs @fredvd is talking about (these), is there any way I can contribute to those @mattrobenolt ?

mattrobenolt:
@InAnimaTe Of course: https://github.com/getsentry/sentry/tree/master/docs/installation/docker :)

InAnimaTe:
Beautiful. I'll have a PR for that too ;)

[...]

PR #12, opened 2016-10-14
Enhanced README with Instructions!

InAnimaTe:
This is a PR adding instructions around utilizing the contents of this repository to lift a fully working Sentry install on Docker. It fills some gaps that the current docker related docs don't cover.

This is a bit more informative than #5 and covers some of the concerns in #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ be fine stopping and removing the containers without worry.
 
 If you'd like to protect your Sentry install with SSL/TLS, there are
 fantastic SSL/TLS proxies like [HAProxy](http://www.haproxy.org/) 
-and [Nginx](https://www.nginx.com/).
+and [Nginx](http://nginx.org/).
 
 ## Resources
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,86 @@ Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docke
  * Docker 1.10.0+
  * Compose 1.6.0+ _(optional)_
 
+## Up and Running
+
+Assuming you've just cloned this repository, the following steps 
+will get you up and running in no time!
+
+1. `mkdir -p data/{sentry,postgres}` - Make our local database and sentry config directories.
+    This directory is bind-mounted with postgres so you don't lose state!
+2. `docker-compose run web config generate-secret-key` - Generate a secret key.
+    Add it to `docker-compose.yml` in `base` as `SENTRY_SECRET_KEY`.
+3. `docker-compose run web upgrade` - Build the database.
+    Use the interactive prompts to create a user account.
+4. `docker-compose up -d` - Lift all services (detached/background mode).
+5. Access your instance at `localhost:9000`!
+
+Note that as long as you have your database bind-mounted, you should
+be fine stopping and removing the containers without worry.
+
+## Backing up postgres
+
+Following with the trend of containers, you could even add something like
+[this](https://github.com/InAnimaTe/docker-postgres-s3-archive) to 
+backup postgres to an AWS S3 bucket:
+
+```
+  postgresqlbackup:
+    image: inanimate/postgres-s3-archive:9.5
+    restart: always
+    links:
+        - postgres:postgres 
+    environment:
+        - "AWS_ACCESS_KEY_ID=PUTACCESSIDHERE"
+        - "AWS_SECRET_ACCESS_KEY=PUTSECRETKEYHERE"
+        - "BUCKET=s3://awesomebackupsbucket/sentry"
+        - "SYMMETRIC_PASSPHRASE=hahacanthaxme"
+        - "NAME_PREFIX=sentry-database-backup"
+        - "PGHOST=postgres"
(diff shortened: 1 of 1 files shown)


mattrobenolt:
I don't know if I like this super opinionated approach to the database backups and how to run a TLS proxy.

Neither of those things are projects we'd want to officially endorse in order to run Sentry.

mattrobenolt:
I guess I'd say, this feels more like an exercise for a blog post, rather than, "How to run Sentry." Since neither of those are core components and there are many opinions on how to do those things correctly, neither of which I'd feel comfortable putting an official stamp on, since that means we're endorsing and condoning.

Not necessarily that I think they're _bad_. I just don't agree this is the place for that.

InAnimaTe:
I do see what you mean. Just fyi, I'm not trying to push one thing or another, I just want to help people get ideas flowing for how they can better utilize other projects to enhance their Sentry installs based on the DevOp toolsets of modern application delivery.

Here's a couple ideas on how to change those around:

We can either do something like (a short message prior to those sections):

> The following are just examples of projects which can be used to enhance your local installation of Sentry. There are a variety of options and you are encouraged to find the correct one for you ;)

Rr I can just drop those sections down to something much more simplified (instead of the recommendation of a certain project and conf. details), we would have:

* If you are interested in running Sentry in a SSL/TLS configuration, there are fantastic proxy tools like HAProxy, Nginx, and Caddy.

I don't mind either way ;)

mattrobenolt:
> Just fyi, I'm not trying to push one thing or another, I just want to help people get ideas flowing for how they can better utilize other projects to enhance their Sentry installs based on the DevOp toolsets of modern application delivery.

Yeah, I understand. :) I would love to have more opinionated documentation out there, but as it is, this should be the core. However others want to scale out or "productionize" their installs is up to them. To me, this falls under the same as us suggesting say, "Run this on Carina." Sure, that works, but it will also work other ways.

> Or I can just drop those sections down to something much more simplified (instead of the recommendation of a certain project and conf. details), we would have:

> If you are interested in running Sentry in a SSL/TLS configuration, there are fantastic proxy tools > like HAProxy, Nginx, and Caddy.

I'm ok with a light recommendation around TLS, just because I feel that's important. I also have no idea what Caddy is, so I'd never mention that in our docs. HAProxy and nginx are both fine.

And definitely remove the database backup thing, because that's just normal infrastructure common sense. You want to backup your database, so figure it out. There's plenty of documentation on it and many different ways to accomplish it. It's also very likely that they're _not_ running Postgres inside of a container as well.

InAnimaTe:
^^ ;)

README.md
```diff
@@ -7,6 +7,29 @@ Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docke
...
  * Compose 1.6.0+ _(optional)_
 
+## Up and Running
+
+Assuming you've just cloned this repository, the following steps 
+will get you up and running in no time!
+
+1. `mkdir -p data/{sentry,postgres}` - Make our local database and sentry config directories.
+    This directory is bind-mounted with postgres so you don't lose state!
+2. `docker-compose run web config generate-secret-key` - Generate a secret key.
```
JeanFred: Shouldn’t we use `--rm` to avoid leaving containers lying?

JeanFred:
I’m also running a production instance of Sentry, using roughly the same steps as this PR suggests.

In case this helps, here is mine (orchestrated using Ansible)
```
    - name: Build the Docker image
      command: docker-compose build

    - name: Run database migrations
      command: docker-compose run --rm web upgrade --noinput

    - name: Run missing Hipchat migrations
      command: docker-compose run --rm web django migrate sentry_hipchat_ac

    - name: Docker-compose up
      command: docker-compose up -d
```

README.md
InAnimaTe: Ahh yes! Just committed the addition. Thanks for pointing that out!

README.md
```diff
@@ -7,6 +7,29 @@ Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docke
...
+5. Access your instance at `localhost:9000`!
+
+Note that as long as you have your database bind-mounted, you should
+be fine stopping and removing the containers without worry.
+
+## Securing Sentry with SSL/TLS
+
+If you'd like to protect your Sentry install with SSL/TLS, there are
+fantastic SSL/TLS proxies like [HAProxy](http://www.haproxy.org/) 
+and [Nginx](https://www.nginx.com/).
```
mattrobenolt: Link to `nginx.org` plz. nginx.com is the commercial entity, and nginx.org is the open source.

InAnimaTe: Done

InAnimaTe:
As you can tell, I'm completely noob to this new fancy review system Github has.

InAnimaTe:
So I've made the requested changes and this has been sitting for over a week here. Is there anything else I'm supposed to do?

mattrobenolt:
Sorry about that, we have a lot of repositories. 😛

PR #14, opened 2016-10-28
Fix cron die

NamPNQ:


mattrobenolt:
I'm not sure what this is trying to fix, but this definitely won't do anything.

PR #15, opened 2016-10-28
Updates to docker compose setup

leifcr:
This removes the base, as the previouse setup run both base + web, which led to running the web service twice.

It uses depends_on and links as references on the worker and cron, since they are the same for web, worker and cron.

### DIFF
--- a/docker-compose-base.yml
+++ b/docker-compose-base.yml
@@ -1,0 +1,14 @@
+version: '2'
+services:
+  sentry_base:
+    build: .
+    environment:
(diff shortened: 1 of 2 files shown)


mattrobenolt:
This was meant to be closed with https://github.com/getsentry/onpremise/commit/177ed0efa8df77a7e76ba557708127eddfd0d8ef

I'm open to accepting some changes, but this seems to over-complicate the problem and far exceeds the intentions of the included file.

PR #16, opened 2016-11-04
Upgrade to use sentry 8.10

JeanFred:
8.10 was released on 2016-11-01


mattrobenolt:
I always forget this one since I can't bump it until it's pushed to Docker Hub, which is out of my control. :(

JeanFred:
@mattrobenolt Thanks for merging :)

PR #17, opened 2016-11-07
Update deprecated filestore settings

gentlecat:


JeanFred:
I encountered the same thing as well and made the same changes :)

mattrobenolt:
Ah, technically, this is ok as-is, but I'll update. Thanks for the nudge!

mattrobenolt:
This was more properly synced with https://github.com/getsentry/onpremise/commit/8186ee21e8f62eb195242269608f3105b4f0c3a2

gentlecat:
Thanks!

PR #19, opened 2016-11-10
Use docker volumes for persistent data.

### LINKED ISSUE #18: Failure to run on OSX (permission denied)
When attempting to generate the secret I get an error:
```
$ docker-compose run --rm web config generate-secret-key
Starting onpremise_postgres_1
chown: changing ownership of ‘/var/lib/sentry/files’: Operation not permitted
```

I would guess this is because of the way the data volumes are mounted from the Mac and don't allow the permission change.

buckett:
Using docker volumes makes this more portable than mounting volumes from the host OS.

Fixes #18

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,11 @@
 # instructions
 
 version: '2'
+
+volumes:
+  sentry-data:
+  postgres-data:
+
 services:
   base:
     build: .
@@ -14,7 +19,7 @@ services:
       SENTRY_POSTGRES_HOST: postgres
       SENTRY_EMAIL_HOST: smtp
     volumes:
-      - ./data/sentry:/var/lib/sentry/files
+      - sentry-data:/var/lib/sentry/files
 
   smtp:
     image: tianon/exim4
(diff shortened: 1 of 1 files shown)


mattrobenolt:
I'm not entirely sure I agree with this nor am I entirely sure I'd want to advocate using docker volumes 100% of the time. This is meant to be more of a guideline and meant for you to adapt to your infrastructure. So I'm trying hard not to dictate these minor implementation details.

Now regretting even attempting to supply a `docker-compose.yml`. I might remove it entirely and leave it as an exercise for the reader.

buckett:
I was just suggesting it as as long as you're on a version of docker that supports it it seems to be more predictable.  If nothing else if someone gets the same error they might at least be able google it, end up at the issue and see that they might be able to workaround it.

mattrobenolt:
> I was just suggesting it as as long as you're on a version of docker that supports it it seems to be more predictable.

I personally disagree that it's more predictable. :) But ultimately these decisions are personal and have pros and cons in each situation and environment. We run docker in production for sentry.io and much prefer mounting volumes from disk rather than using docker volumes, for example.

mattrobenolt:
I will say though that tinkering with stuff locally, I'd prefer a docker volume. Where as in a production deployment, I would not.

trurl123:
This PR really helps me! I could not install Sentry without it.

PR #30, opened 2017-02-10
Add note to README to escape $ chars

gqmelo:
When using latest docker-compose I've got "Invalid interpolation format" because the secret key I added to docker-compose.yml had $. It was not obvious what was the problem.


mattrobenolt:
I'm going to just close this because we resolved this on master by not generating keys like this anymore to work around this issue. So the next version won't be affected by this and I'll never remember to remove this.

PR #34, opened 2017-03-10
support for github enterprise SSO

QingsongYao:
we need to install a python library in order to enable SSO provider for Github Enterprise
also need to expose two variables which set the correct domain name and api url for Github Enterprise.


mattrobenolt:
I don't want this here. This is something you'd do in your own personal install. The purpose of this repository is to be forked and modified for your own needs. :)

PR #51, opened 2017-08-08
Add restart policy unless-stopped

guyskk:
- make sentry auto start when the system started
- make sentry auto start when killed


mattrobenolt:
Can I suggest `unless-stopped` instead?

guyskk:
@mattrobenolt `unless-stopped` looks better, I will use it.

PR #58, opened 2017-10-13
Add example with nginx configuration.

andriisoldatenko:
I think it's better to have working version with nginx also.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -1,0 +1,18 @@
+upstream web {
+    server web:9000;
+}
+
+server {
+    listen 80;
+    charset utf-8;
+
+    location / {
+        try_files $uri @proxy_to_app;
+    }
+    location @proxy_to_web {
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Host $http_host;
+        proxy_redirect off;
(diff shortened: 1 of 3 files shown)


nginx/nginx.conf
```diff
@@ -0,0 +1,18 @@
+upstream web {
```
mattrobenolt: You don't need to define this upstream at all.

mattrobenolt:
The only push back I'd have here is that more typically than not, you probably also want TLS, and this pushes people moreso to running just plain port 80. I was hoping to leave this bit as an excercise for the reader since there are lots of different ways to do it. Maybe you just use an ELB or something like that. If you deploy on systems like Kubernetes, you can usually just use a different "router" load balancer in which case this nginx here isn't too valuable.

andriisoldatenko:
@mattrobenolt Yes my main goal to add SSL also.

mattrobenolt:
Sure. And I'd just argue that's a bit out of the scope of what we should dictate here. There are lots of TLS things to configure, how to get certificates, etc. And again, would like to support cases where they already have a load balancer or some other TLS termination service which is pretty common here.

andriisoldatenko:
@mattrobenolt I'll close PR

PR #60, opened 2017-10-16
Add Procfile and instructions for running on Aptible

adrian-gomez:


sentry.conf.py
```diff
@@ -304,3 +304,20 @@
...
     BITBUCKET_CONSUMER_SECRET = env('BITBUCKET_CONSUMER_SECRET')
+
+## Custom hint modifications
+
+# Use 3 web workers
+SENTRY_WEB_OPTIONS = {
+    'workers': 3,  # the number of web workers
+}
+
+# Force ssl when using ssl
```
mattrobenolt: I definitely disagree with these changes in the config file. We don't not support `django-sslify` nor do we want this installed as a default which we, the Sentry team, don't use ourselves.

sentry.conf.py
```diff
@@ -304,3 +304,20 @@
...
+if env('SENTRY_USE_SSL', False):
+    MIDDLEWARE_CLASSES = (
+        'sslify.middleware.SSLifyMiddleware',
+    ) + MIDDLEWARE_CLASSES
+
+# Restrict registration of new users
```
mattrobenolt: These should not be dictated by default. The point of including the config here is so you can edit it yourself and apply whatever changes you want.

sentry.conf.py
```diff
@@ -304,3 +304,20 @@
 if 'BITBUCKET_CONSUMER_KEY' in os.environ:
     BITBUCKET_CONSUMER_KEY = env('BITBUCKET_CONSUMER_KEY')
     BITBUCKET_CONSUMER_SECRET = env('BITBUCKET_CONSUMER_SECRET')
+
+## Custom hint modifications
+
+# Use 3 web workers
```
mattrobenolt: This isn't needed either.

adrian-gomez:
@mattrobenolt sorry about this PR, it was to be opened on our fork but it was opened upstream by mistake.

PR #62, opened 2017-10-31
Feature/add ssl support

vananabu:


mattrobenolt:
I really don’t want to bundle this by default. I’d rather we just document it as an example. There are too many ways to do this, and it’s  out of the scope of Sentry itself. See https://github.com/getsentry/onpremise/pull/58 for similar discussion.

vananabu:
I know, I chose the wrong project when was creating pr. sorry.

PR #73, opened 2018-02-16
docker compose version increased

### LINKED ISSUE #72: docker compose version
**issue**

https://github.com/getsentry/onpremise/blob/a514caf906fa22a66eaf6532a721a9393177d2ff/docker-compose.yml#L12

> Value 'unless-stopped' is not expected here

**solution**

'unless-stopped' was added in version 3
https://github.com/getsentry/onpremise/blob/a514caf906fa22a66eaf6532a721a9393177d2ff/docker-compose.yml#L9

sseypt:
Fixes #72

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@
 # See docs.getsentry.com/on-premise/server/ for full
 # instructions
 
-version: '2'
+version: '3'
 services:
   base:
     restart: unless-stopped


sseypt:
extends is not supported in v3 anymore (https://github.com/moby/moby/issues/31101)
https://github.com/getsentry/onpremise/blob/a514caf906fa22a66eaf6532a721a9393177d2ff/docker-compose.yml#L45

PR #74, opened 2018-02-19
Allow configuration of S3 as the filestore backend

aflatter:
Hello sentry folks,

I didn't like to store files in a local directory, so I added a way to configure the filestore backend using environment variables. I would be very happy if you'd accept these changes, because it would free me from maintaining my own Docker images. For reference: I created a [automated build](https://hub.docker.com/r/tevim/sentry-onpremise-s3) on Docker Hub, which will be updated by our team.

A last remaining issue for me is that the Sentry's base Dockerfile specifies `/var/lib/sentry/files` as a volume. Would it be possible to change this and to leave configuration of the volume to the user?

Thanks

### DIFF
--- a/sentry.conf.py
+++ b/sentry.conf.py
@@ -17,7 +17,12 @@
 #  SENTRY_REDIS_DB
 #  SENTRY_MEMCACHED_HOST
 #  SENTRY_MEMCACHED_PORT
+#  SENTRY_FILESTORE_BACKEND
 #  SENTRY_FILESTORE_DIR
+#  SENTRY_FILESTORE_BUCKET_NAME
+#  SENTRY_FILESTORE_ACCESS_KEY
+#  SENTRY_FILESTORE_SECRET_KEY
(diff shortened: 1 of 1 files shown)


dcramer:
I'm +1 on this change (in general, not the actual implementation)

mattrobenolt:
This needs to first be done in https://github.com/getsentry/docker-sentry, since these files are just mirrored from there.

And to address the comment about the `VOLUME`, this is just standard practice. You don't have to use this `VOLUME`, and you can mount in whatever you want and just configure it differently, thought there's not a lot of reason to do that.

So for now, since I can't accept the change in here, I'm going to close this and we can bring it over into `docker-sentry` repo instead.

janraasch:
Hi there, I took the changeset and moved it to a pr on docker-sentry: https://github.com/getsentry/docker-sentry/pull/127 so we can continue the conversation over there.

PR #78, opened 2018-03-09
Switch from "links" to "depends_on" in docker-compose.yml

tivaliy:
Now in Docker all containers can reach each other while they are in the same network. docker-compose transparently creates default network. Also `links` functionality is deprecated. In order to not break starting order of services we need to replace links with `depends_on`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,7 +43,7 @@ services:
   web:
     restart: unless-stopped
     extends: base
-    links:
+    depends_on:
       - redis
       - postgres
       - memcached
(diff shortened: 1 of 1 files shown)


mattrobenolt:
I think doing this will break assumptions made though. In the Sentry container, we read environment variables from the linked containers. And Docker networking doesn’t share environment variables.

tivaliy:
Am I miss something? As far as I see we use `extends` of `base` to share sentry environment variables across containers.

mattrobenolt:
Lots of configuration like this: https://github.com/getsentry/docker-sentry/blob/master/8.22/sentry.conf.py#L43

revolunet:
should have been fixed by #105

JanMalte:
This issue should be closed as this is fixed now

BYK:
Closing per comments above.

PR #105, opened 2018-07-13
Docker compose 3.4

revolunet:
- upgrade docker compose to 3.4
      - use `depends_on`
      - reuse a `default`  block
      - use docker named volumes

and some readme fixes


mattrobenolt:
Going to assume this works fine. 🥇

PR #110, opened 2018-07-22
Use default parameter of env function

joren485:
`sentry.conf.py` uses `or` to set default values for environment variables, but [the `env` function has a `default=''` parameter](https://github.com/getsentry/sentry/blob/2ea999f948e88bf6fb606753ddaf79af8cf5085e/src/sentry/conf/server.py#L34). This PR uses that parameter instead of `or`.

I left the statements with two `or` uses the same, because I believe
```
env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER')
or 'postgres'
```
is more readable than
```
env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER', 'postgres')
```

### DIFF
--- a/sentry.conf.py
+++ b/sentry.conf.py
@@ -61,10 +61,7 @@ if postgres:
                 or ''
             ),
             'HOST': postgres,
-            'PORT': (
-                env('SENTRY_POSTGRES_PORT')
-                or ''
-            ),
+            'PORT': env('SENTRY_POSTGRES_PORT'),
             'OPTIONS': {
                 'autocommit': True,
             },
@@ -97,9 +94,9 @@ redis = env('SENTRY_REDIS_HOST') or (env('REDIS_PORT_6379_TCP_ADDR') and 'redis'
 if not redis:
     raise Exception('Error: REDIS_PORT_6379_TCP_ADDR (or SENTRY_REDIS_HOST) is undefined, did you forget to `--link` a redis container?')
 
-redis_password = env('SENTRY_REDIS_PASSWORD') or ''
-redis_port = env('SENTRY_REDIS_PORT') or '6379'
-redis_db = env('SENTRY_REDIS_DB') or '0'
+redis_password = env('SENTRY_REDIS_PASSWORD')
+redis_port = env('SENTRY_REDIS_PORT', '6379')
+redis_db = env('SENTRY_REDIS_DB', '0')
 
 SENTRY_OPTIONS.update({
     'redis.clusters': {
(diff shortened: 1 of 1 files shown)


joren485:
@glensc 
I did not give any examples.

The code in the description is to show that I did not change statements with more than one `or`, because I believe they are more readable.

glensc:
yes, you did gave example. in PR body:

```
I left the statements with two or uses the same, because I believe

env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER')
or 'postgres'
is more readable than

env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER', 'postgres')
```

but i think it should be swapped:

```
I left the statements with two or uses the same, because I believe

env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER', 'postgres')
env('SENTRY_DB_NAME')

is more readable than

or env('POSTGRES_ENV_POSTGRES_USER')
or 'postgres'
```

glensc:
you write in your example that SNIPPET1 is better than SNIPPET2, but then your example seem to have SNIPPET1 and SNIPPET2 swapped, because looking at your diff, you consider SNIPPET2 better than SNIPPET1.

in your SNIPPET2 you use the `default` parameter, but in SNIPPET1 you don't. and PR title says it's better to use `default` parameter.

also updated previous post, poor copy paste: https://github.com/getsentry/onpremise/pull/110#issuecomment-421755245

joren485:
Thanks for your clear explanation.

I consider the use of the `default` parameter better than the use of one `or` statement .
I, however, do not consider the use of the `default` parameter better than the use of more than one `or` statement.

In the following two examples, `Example 1` has one `or` and `Example 2` has two `or`.

##### Example 1

I believe that 
```
env('SENTRY_MEMCACHED_PORT', '11211')
```
is more readable than
```
env('SENTRY_MEMCACHED_PORT')
or '11211'
``` 

##### Example 2

I do not believe that
```
env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER', 'postgres')
``` 
is more readable than
```
env('SENTRY_DB_NAME')
or env('POSTGRES_ENV_POSTGRES_USER')
or 'postgres'
```

mattrobenolt:
I would like to point out that this is a change in behavior. It’s not just a design choice.

If the environment variable exists, but is empty, it’s not going to hit the “default” code branch.

PR #115, opened 2018-08-06
data doesn't mounted to data directory

mofelee:
We should put data configuration back or update readme.

https://github.com/getsentry/onpremise/pull/105

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ x-defaults: &defaults
     SENTRY_POSTGRES_HOST: postgres
     SENTRY_EMAIL_HOST: smtp
   volumes:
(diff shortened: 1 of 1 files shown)


revolunet:
Why not volumes ?

"If you are developing new Docker applications, consider using named volumes instead"

https://docs.docker.com/storage/bind-mounts/

mofelee:
@revolunet 
I have no inclination in whether or not to use named volumes. If you want to use it, I think we should to modify the readme. 😄 

image

revolunet:
Thanks, indeed

mattrobenolt:
I would much rather just update the README and maintain using volumes.

PR #127, opened 2018-09-26
Allow url_prefix to be set via environment variable

aloyr:
This adds the option to configure settings.url-prefix via an environment variable on the container.

I am experimenting with sentry.io for my employer and made this change to our stack as well as other changes pertaining to LDAP setup.


bariman:
LGTM

mattrobenolt:
This can't be accepted in this repository, and would have to go through https://github.com/getsentry/docker-sentry

PR #137, opened 2018-11-11
Use alpine images

McSneaky:


McSneaky:
Reverted it back to regular image

I have been using Postgres Alpine image for quite some time in production and have not noticed any errors  / anomalies. 
But since I have not run any tests I can't be 100% sure it works the same as non Alpine version

PR #139, opened 2018-11-16
Upload example of systemd service file

adw0rd:


mattrobenolt:
This doesn't really belong here, and on top of it, I would never suggest blasting docker-compose through a systemd service file. That seems very niche and would not be encouraged.

PR #167, opened 2019-04-15
feat: Improve configuration to be less Docker-specific

dcramer:
This removes auto-binding of various values (specified in config.yaml) when they're not actually configured. It ensures that these values can then be configured from the web UI as Sentry has intended.

See also https://github.com/getsentry/docker-sentry/issues/163#issuecomment-483011252

### DIFF
--- a/sentry.conf.py
+++ b/sentry.conf.py
@@ -40,6 +40,7 @@ from sentry.conf.server import *  # NOQA
 
 import os
 import os.path
+import six
 
 CONF_ROOT = os.path.dirname(__file__)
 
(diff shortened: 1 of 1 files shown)


dcramer:
I haven't actually tested this out yet, just wanted to get this up before I get my day started.

sentry.conf.py
```diff
@@ -250,46 +251,52 @@
...
-email = env('SENTRY_EMAIL_HOST') or (env('SMTP_PORT_25_TCP_ADDR') and 'smtp')
-if email:
-    SENTRY_OPTIONS['mail.backend'] = 'smtp'
-    SENTRY_OPTIONS['mail.host'] = email
-    SENTRY_OPTIONS['mail.password'] = env('SENTRY_EMAIL_PASSWORD') or ''
-    SENTRY_OPTIONS['mail.username'] = env('SENTRY_EMAIL_USER') or ''
-    SENTRY_OPTIONS['mail.port'] = int(env('SENTRY_EMAIL_PORT') or 25)
-    SENTRY_OPTIONS['mail.use-tls'] = env('SENTRY_EMAIL_USE_TLS', False)
-else:
-    SENTRY_OPTIONS['mail.backend'] = 'dummy'
```
dcramer: ive intentionally removed this behavior as email is required for a production sentry install, and you can configure the smtp settings in the UI

dcramer:
Also still not 100% clear if I should PR this or just patch the 9.1 config in docker-sentry, but I assumed this was the source of truth.

sentry.conf.py
```diff
@@ -250,46 +251,52 @@
...
+# enable more "smart" configuration.
+
+ENV_CONFIG_MAPPING = {
+    'SENTRY_EMAIL_PASSWORD': 'mail.password',
+    'SENTRY_EMAIL_USER': 'mail.username',
+    'SENTRY_EMAIL_PORT': ('mail.port', int),
+    'SENTRY_EMAIL_USE_TLS': ('mail.use-tls', bool),
+    'SENTRY_EMAIL_HOST': 'mail.host',
+    'SENTRY_SERVER_EMAIL': 'mail.from',
+    'SENTRY_ENABLE_EMAIL_REPLIES': 'mail.enable-replies',
```
acdha: I had to make this be a boolean:

```suggestion
    'SENTRY_ENABLE_EMAIL_REPLIES': ('mail.enable-replies', bool),
```

sentry.conf.py
```diff
@@ -250,46 +251,52 @@
...
+
+    # If you're using mailgun for inbound mail, set your API key and configure a
+    # route to forward to /api/hooks/mailgun/inbound/
+    'SENTRY_MAILGUN_API_KEY': 'mail.mailgun-api-key',
+
+    'SLACK_CLIENT_ID': 'slack.client-id',
+    'SLACK_CLIENT_SECRET': 'slack.client-secret',
+    'SLACK_VERIFICATION_TOKEN': 'slack.verification-token',
+
+    'SECRET_KEY': 'system.secret-key',
```
acdha:
```suggestion
    'SENTRY_SECRET_KEY': 'system.secret-key',
```

dcramer: In this case this is actually SECRET_KEY (Django's setting)

dcramer: oh wait actually this is the env name, let me double check everything now ;)

acdha: I can say that these were the only two things which prevented me from from running the container. I'm still debugging my SAML2 configuration, though, and haven't looked at anything else.

markotitel:
Hi,
I've tried to build sentry image using this branch. But docker logs tells `!! Configuration error: ValueError: invalid literal for int() with base 10: ''`

dcramer:
@markotitel thanks! Might have to make the value checks look for empty before coercing.

dcramer:
@markotitel feel free to give the branch another try now

gebv:
Maybe forgot `SECRET_KEY` rename to `SENTRY_SECRET_KEY`?

dcramer:
@gebv should be in there -- make sure you do a hard reset as ive forced pushed any fixes

sentry.conf.py
```diff
@@ -250,46 +251,54 @@
...
 
-#####################
-# SLACK INTEGRATION #
-#####################
-slack = env('SLACK_CLIENT_ID') and env('SLACK_CLIENT_SECRET')
-if slack:
-    SENTRY_OPTIONS['slack.client-id'] = env('SLACK_CLIENT_ID')
-    SENTRY_OPTIONS['slack.client-secret'] = env('SLACK_CLIENT_SECRET')
-    SENTRY_OPTIONS['slack.verification-token'] = env('SLACK_VERIFICATION_TOKEN') or ''
+def bind_env_config(config=SENTRY_OPTIONS, mapping=ENV_CONFIG_MAPPING):
```
markstory: Do you need to worry about mutable default arguments here?

dcramer: what do you mean?

markstory: Nevermind, it isn't a relevant issue for this function.

dcramer:
Expanded config to include new github + vsts vars.

sentry.conf.py
```diff
@@ -250,46 +263,63 @@
...
-    SENTRY_OPTIONS['mail.enable-replies'] = env('SENTRY_ENABLE_EMAIL_REPLIES', False)
+# Docker's environment configuration needs to happen
+# prior to anything that might rely on these values to
+# enable more "smart" configuration.
+
+ENV_CONFIG_MAPPING = {
+    'SENTRY_EMAIL_PASSWORD': 'mail.password',
+    'SENTRY_EMAIL_USER': 'mail.username',
+    'SENTRY_EMAIL_PORT': ('mail.port', int),
+    'SENTRY_EMAIL_USE_TLS': ('mail.use-tls', bool),
```
mattrobenolt: We need to use `sentry.utils.types:Bool` otherwise this coersion is going to yield a very unexpected behavior and a change from what it was before. Doing `env('...', False)` with `False` as the default would cause it to use this `Bool` type becuase it gets passed through `sentry.utils.types:type_from_value`.

sentry.conf.py
```diff
@@ -276,12 +277,13 @@
 ENV_CONFIG_MAPPING = {
     'SENTRY_EMAIL_PASSWORD': 'mail.password',
     'SENTRY_EMAIL_USER': 'mail.username',
-    'SENTRY_EMAIL_PORT': ('mail.port', int),
-    'SENTRY_EMAIL_USE_TLS': ('mail.use-tls', bool),
+    'SENTRY_EMAIL_PORT': ('mail.port', Int),
```
dcramer: was this broken before then? i was re-implementing the type casting that was present

mattrobenolt: The `Int` one wasn't, but `Bool` was. I just did `Int` to do them both the same. Coersion happens by default in `env()` when y ou pass a default. And these were doing like, `env('SENTRY_EMAIL_USE_TLS', False)` which inferred the type as Bool, to do the correct coersion.

The problem here with lowercase `bool` is that it doesn't make a lot of sense for env vars. When converting a string to a bool, `bool('0')` for example, is True. Whereas my `Bool('0')` is False. Among other expected values.

dcramer: :+1:

PR #178, opened 2019-04-30
fixed issue 177 - system.secret-key must be set

csbarker:
https://github.com/getsentry/onpremise/issues/177


tobiasmcnulty:
In our tests we needed the secret key both in the Docker container environment and config.yml...Not entirely sure why...

csbarker:
@tobiasmcnulty Yeah seems so. I've updated the README to tell people to update both files. Its not the greatest solution but at least this will allow them to run an instance by following the instructions.

jwillmer:
This is a workaround, not a solution. Why does the system not recognise (some) the `.env` file variables anymore?

mattrobenolt:
I agree. Because we broke it. I’m going to figure out why and fix it.

PR #179, opened 2019-05-02
Update Dockerfile for sentry 9.1.1

obfuscoder:


mattrobenolt:
I would personally rather not pin versions explicitly like this. It doesn't really matter, and new installs are going to get latest `9.1.x`. This pin works so latest patches get pulled at the time of doing `docker-compose build --pull`. I'd argue that to get much value out of this repo, you need to fork it anyways, in which case, pinning every explicit version instead of a fuzzy version is not very useful.

mattrobenolt:
I've added additional clarifications here: https://github.com/getsentry/onpremise/commit/89ec146ee137d3fd47df6154d68b8e92e9892467

PR #183, opened 2019-05-07
Fix: 'mail.enable-replies' type

jwillmer:
Exception: 'mail.enable-replies': got <type 'str'>, expected boolean


BYK:
Confirmed that this is an issue and the propose fix solves it. Thanks a lot @jwillmer!

PR #192, opened 2019-06-11
fix(email): Add `SENTRY_EMAIL_LIST_NAMESPACE` env setting

BYK:
Port of getsentry/docker-sentry#170


.env.example
```diff
@@ -1,3 +1,4 @@
 # Run `docker-compose run web config generate-secret-key`
 # to get the SENTRY_SECRET_KEY value.
 SENTRY_SECRET_KEY=
+SENTRY_EMAIL_LIST_NAMESPACE=localhost
```
mattrobenolt: I don’t think I’d add it in here since it’s such a rare setting.

BYK: Where should we put the defaults tho as the conf file doesn't really allow that for ENV mappings?

mattrobenolt: I’m not sure. But it needs to fallback to the same default option value. I forget all the moving pieces in play here honestly. But we can’t require every value to be explicitly set since it duplicates a lot of efforts.

PR #209, opened 2019-07-02
docs(readme): Update readme for the install script

BYK:
Follow up to #207.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,39 +1,35 @@
-# Sentry On-Premise
+# Sentry On-Premise [![Build Status][build-status-image]][build-status-url]
 
 Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docker](https://www.docker.com/).
 
 ## Requirements
 
  * Docker 1.10.0+
- * Compose 1.17.0+ _(optional)_
- 
- ## Minimum Hardware Requirements:
- 
- * You need at least 3GB Ram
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,39 +1,35 @@
...
-    Docker volumes have to be created manually, as they are declared as external to be more durable.
-2. `cp -n .env.example .env` - create env config file
-3. `docker-compose build` - Build and tag the Docker services
-4. `docker-compose run --rm web config generate-secret-key` - Generate a secret key.
-    Add it to `.env` as `SENTRY_SECRET_KEY`.
-5. `docker-compose run --rm web upgrade` - Build the database.
-    Use the interactive prompts to create a user account.
-6. `docker-compose up -d` - Lift all services (detached/background mode).
-7. Access your instance at `localhost:9000`!
+ * Compose 1.17.0+
```
mattrobenolt: I’d still leave the note that this is optional. Running Sentry does not have some hidden dependency on Docker Compose, and what’s provided is still an example. Ideally users use whatever scheduler they want.

BYK: I'd say when we have an "official" repo, it is better to reduce the surface area we cover. What we guarantee here in the repo is the flow with `docker-compose` which should be okay for local or small deploys. I'm sure folks who want to use something fancier wouldn't be bothered by this part :)

PR #211, opened 2019-07-08
feat(nightlies): Add ability to build from nightlies

BYK:
This patch adds the optional IMAGE env variable to set the base
image and adds nightly builds to Travis CI.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -5,9 +5,9 @@ TEST_USER='test@sentry.io'
 TEST_PASS='test123TEST'
 COOKIE_FILE=$(mktemp)
 declare -a TEST_STRINGS=(
-    '"isAuthenticated":true,'
-    '"username":"test@sentry.io",'
-    '"isSuperuser":true,'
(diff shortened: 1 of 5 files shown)


test.sh
```diff
@@ -5,9 +5,9 @@ TEST_USER='test@sentry.io'
 TEST_PASS='test123TEST'
 COOKIE_FILE=$(mktemp)
 declare -a TEST_STRINGS=(
-    '"isAuthenticated":true,'
-    '"username":"test@sentry.io",'
-    '"isSuperuser":true,'
+    '"isAuthenticated":true'
+    '"username":"test@sentry.io"'
```
mattrobenolt: Do we maybe wanna put a less potentially real email? I'm not sure where this goes, but also not sure if there's value in there being a real domain either.

BYK: This is used in the test script and flow. Happy to change it to something more random in a follow-up PR.

PR #237, opened 2019-08-08
fix(uwsgi): Fix broken uWSGI config

BYK:
This patch introduces the following fixes to uWSGI config which
was not set up correctly to be used publicly (not behind a router,
load balancer, proxy like nginx etc.):

 - Use `http` option for a front-proxy
 - Set protocol back to `uwsgi` due to above
 - Prevent forced socket binding to `uwsgi` due to protocol change
 above (https://git.io/fj7Lw)
 - Turn on "keep-alive" support

Should fix getsentry/sentry-cli#40.

### DIFF
--- a/sentry.conf.py
+++ b/sentry.conf.py
@@ -261,6 +261,7 @@ SENTRY_WEB_PORT = 9000
 SENTRY_WEB_OPTIONS = {
     'http': '%s:%s' % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
     'protocol': 'uwsgi',
+    # This is need to prevent https://git.io/fj7Lw
     'uwsgi-socket': None,
     'http-keepalive': True,
     'add-header': 'Connection: Keep-Alive',
--- a/sentry.conf.py
+++ b/sentry.conf.py
@@ -259,9 +259,15 @@ if env('SENTRY_USE_SSL', False):
 SENTRY_WEB_HOST = '0.0.0.0'
 SENTRY_WEB_PORT = 9000
 SENTRY_WEB_OPTIONS = {
+    'http': '%s:%s' % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
+    'protocol': 'uwsgi',
+    'uwsgi-socket': None,
+    'http-keepalive': True,
+    'add-header': 'Connection: Keep-Alive',
+    'memory-report': False,
+    'static-gzip-all': True,
     # 'workers': 3,  # the number of web workers
 }
-
 
 
 ##########
--- a/sentry.conf.py
+++ b/sentry.conf.py
@@ -264,9 +264,7 @@ SENTRY_WEB_OPTIONS = {
     # This is need to prevent https://git.io/fj7Lw
     'uwsgi-socket': None,
     'http-keepalive': True,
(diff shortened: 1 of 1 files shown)


sentry.conf.py
```diff
@@ -259,11 +259,18 @@
...
 SENTRY_WEB_PORT = 9000
 SENTRY_WEB_OPTIONS = {
+    'http': '%s:%s' % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
+    'protocol': 'uwsgi',
+    # This is need to prevent https://git.io/fj7Lw
+    'uwsgi-socket': None,
+    'http-keepalive': True,
+    'add-header': 'Connection: Keep-Alive',
+    'memory-report': False,
+    'static-gzip-all': True,
```
mattrobenolt: What's this do and how does it apply to us? We serve static files from the python app. And there's already logic there to serve pre-gzipped files if they exist, but I don't think we pre-gzip them at build time for anything on-premise.

sentry.conf.py
```diff
@@ -259,11 +259,18 @@
 SENTRY_WEB_HOST = '0.0.0.0'
 SENTRY_WEB_PORT = 9000
 SENTRY_WEB_OPTIONS = {
+    'http': '%s:%s' % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
+    'protocol': 'uwsgi',
+    # This is need to prevent https://git.io/fj7Lw
+    'uwsgi-socket': None,
+    'http-keepalive': True,
+    'add-header': 'Connection: Keep-Alive',
```
mattrobenolt: You don't need to add this header because with http/1.1 keepalive is implied. If you _don't_ want keepalive, you need an explicit `Connection: close`.

sentry.conf.py
```diff
@@ -259,11 +259,18 @@
...
+    'memory-report': False,
```
mattrobenolt: What's this doing here?

sentry.conf.py
BYK: Ah, my bad. I got confused about the serving logic. Will remove.

sentry.conf.py
BYK: You are right, that said I was simply following the docs here: https://uwsgi-docs.readthedocs.io/en/latest/HTTP.html#http-keep-alive

Will remove and see if it still works (I think it would based on our discussions on Slack)

sentry.conf.py
BYK: I figured we don't need memory reports on on-premise, do we?

BYK:
If we agree on these, I'll migrate some of them to https://github.com/getsentry/sentry/blob/34a554c14b68285d8a8eb6c5c4c56dfc1db9a83a/src/sentry/runner/commands/devserver.py#L81-L96 (especially keep alive and protocol fixes.

sentry.conf.py
mattrobenolt: I don't see why we'd disable it though. tbh not entirely sure what it does.

BYK: We explicitly set it to `true` in our default config and I have no idea why. Maybe we should reconsider that decision then. Will do in a follow-up diff.

PR #238, opened 2019-08-09
Give error message if environment variable SENTRY_IMAGE is not set

stumbaumr:
If SENTRY_IMAGE is not set docker-compose will just give a warning and default to a blank string and continue.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ x-defaults: &defaults
   build:
     context: .
     args:
-      SENTRY_IMAGE: ${SENTRY_IMAGE}
+      SENTRY_IMAGE: ${SENTRY_IMAGE:?err}
   depends_on:
     - redis
     - postgres


BYK:
Thanks for sending this PR @stumbaumr! This syntax is not supported on the compose v3 file format, which is the one we are using. For this reason, I won't be able to merge your patch. If this syntax was supported, I'd have opted for the default value fallback and point that tho 9.1 or the 10 image though.

stumbaumr:
'''
root@sentry01(dc1.prd):~/getsentry/onpremise (git:master)# grep version docker-compose.yml
version: '3.4'
root@sentry01(dc1.prd):~/getsentry/onpremise (git:master)# grep "?err" docker-compose.yml
        SENTRY_IMAGE: ${SENTRY_IMAGE:?err}
root@sentry01(dc1.prd):~/getsentry/onpremise (git:master)# docker-compose up -d --scale worker=2
ERROR: Missing mandatory value for "build" option in service "web": err
root@sentry01(dc1.prd):~/getsentry/onpremise (git:master)#
'''

No idea, why it works here? Maybe docker-compose does not care about it's documentation...

stumbaumr:
Maybe best to wait for the outcome from this:
https://github.com/docker/docker.github.io/issues/9263

BYK:
I'm hoping to remove this anyway as the user experience is not great. There was another issue for the lack of this feature's support and I don't think it's gonna get love any time soon.

PR #265, opened 2019-10-25
Declare ARG in Dockerfile.

mfb:
Resolves [Warning] One or more build-args [SENTRY_IMAGE] were not consumed.


mfb:
Or you might prefer to declare the default value in the ARG statement 🤷 :)

BYK:
Thanks a lot!

PR #285, opened 2019-11-27
fix(tracing): Force-disable self-tracing on on-premise

BYK:


dashed:
For posterity, this option was added in https://github.com/getsentry/sentry/pull/14055

dashed:
@BYK I haven't run onpremise before, so I'm assuming this works 😄

PR #288, opened 2019-11-30
ref(kafka): Simpler kafka listener setup

BYK:


docker-compose.yml
```diff
@@ -76,10 +75,10 @@ services:
     image: 'confluentinc/cp-kafka:5.1.2'
     environment:
       KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
-      KAFKA_LISTENERS: 'INTERNAL://0.0.0.0:9093,EXTERNAL://0.0.0.0:9092'
-      KAFKA_ADVERTISED_LISTENERS: 'INTERNAL://kafka:9093,EXTERNAL://kafka:9092'
-      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: 'INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT'
-      KAFKA_INTER_BROKER_LISTENER_NAME: INTERNAL
+      KAFKA_LISTENERS: 'EVERYONE://kafka:9092'
```
mattrobenolt: In this case, you don’t even need a special name. Just use PLAINTEXT.

PR #290, opened 2019-12-03
fix(kafka): Disable support metrics in Kafka & ZK

BYK:
Disables the annoying automatic support metrics for Kafka and ZK while defaulting all logs to WARN level, instead of some being WARN and some ERROR.


BYK:
>why though, we don't do any of these tunings.

Just prevents annoying logs and call to home every day: https://docs.confluent.io/current/proactive-support.html

mattrobenolt:
jfc

@JTCunning @b1naryth1ef we should make sure this is disabled for everything if we can. This is pretty scummy.

PR #291, opened 2019-12-03
fix(volumes): Fix creation of random volumes at every run

BYK:
Plugs all the holes (`VOLUME` defs) in the images we use so they at least persist across runs/restarts for things like random logs or SMTP spool. This also fixes symbolicator_cleanup service not having access to symbolicator data.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,9 @@ services:
   smtp:
     << : *restart_policy
     image: tianon/exim4
+    volumes:
+      - 'sentry-smtp:/var/spool/exim4'
+      - 'sentry-smtp-log:/var/log/exim4'
   memcached:
     << : *restart_policy
     image: 'memcached:1.5-alpine'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -154,3 +163,8 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
+  sentry-secrets:
```
mattrobenolt: Is this valid yaml? I'd expect you to need to do like, `sentry-smtp: {}`

BYK: it works so it is valid I guess? :) 🤷‍♂

mattrobenolt: I don't like it.

TIL

```yaml
foo:
bar:
```

is the same as

```json
{
  "foo": null,
  "bar": null
}
```

BYK:
>Do you actually h ave to prefix them all? I feel like docker-compose already adds a prefix to everything to avoid conflicts.

It does and uses the directory name as the prefix (`onpremise` in this case) which is not very descriptive. I'll address this by adding an explicit name to the project in a later PR.

docker-compose.yml
BYK: Just saw this sorry, I can change if you think it is confusing?

mattrobenolt: It's fine. I just don't like it because it appears to be invalid syntax. But honestly, most of the yaml syntax you're using looks like invalid syntax. So this is minimal.

BYK: >But honestly, most of the yaml syntax you're using looks like invalid syntax. 

Lol. If we have a style guide or something, happy to follow that. I think my editor auto formats this using prettier.

PR #292, opened 2019-12-04
feat(upgrade): Add automatic Snuba migration for upgrades

BYK:
Upgrades existing events from last `$SENTRY_EVENT_RETENTION_DAYS`
to Snuba automatically. Relies on ~~getsentry/sentry#15934~~.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -55,10 +55,8 @@ if [ "$RAM_AVAILABLE_IN_DOCKER" -lt "$MIN_RAM" ]; then
     exit -1
 fi
 
-echo ""
-ensure_file_from_example $SENTRY_CONFIG_PY
-ensure_file_from_example $SENTRY_CONFIG_YML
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -90,6 +65,14 @@ echo "Created $(docker volume create --name=sentry-kafka)."
 echo "Created $(docker volume create --name=sentry-clickhouse)."
 echo "Created $(docker volume create --name=sentry-symbolicator)."
 
+# Ensure nothing is working while we install/update
+docker-compose down
```
tonyo: Do we really need "down" (removes volumes, images, etc.), and not "stop" here?

BYK: Good question! I'd rather recreate containers but I think that automatically happens after a `docker-compose build` when needed so I think `stop` is better here. Updating.

PR #295, opened 2019-12-05
fix(data): Use `/data` for persisting files and file caches

BYK:
Moves `sentry-data` volume to `/data` mount point and sets all file-based storage settings to `/data/files`, `/data/dsym-cache` etc. accordingly. See https://github.com/getsentry/sentry/blob/50ac5506669043dd939dd0e44fe4cfb00377ff1d/src/sentry/options/defaults.py#L45-L54

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,9 +21,7 @@ x-sentry-defaults: &sentry_defaults
   environment:
     SNUBA: 'http://snuba-api:1218'
   volumes:
-    - 'sentry-data:/var/lib/sentry/files'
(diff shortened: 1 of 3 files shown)


BYK:
>How does this shared volume work across multiple containers?

Just like a shared network mount: if you're not careful you may step on each other's toes.

lzz2017:
What should we do if we want to connect to the database locally

BYK:
>What should we do if we want to connect to the database locally

This issue is not related to the database.

Issue #279, opened 2019-11-22
Allow configuring web port as env var


larsnystrom:
Most of the stuff I need to change in docker-compose can be done by just adding a `docker-compose.override.yml` file (like adding `http_proxy` env vars and switching smtp image). But it is not possible to override the port specification of the `web` service.

I think the cleanest solution would be to change 

```
ports:
  - "9000:9000/tcp"

```
to

```
ports:
  - "${SENTRY_WEB_PORT}:9000/tcp"
```

in `docker-compose.yml`.

What do you think?

My reason for changing the port is simply to restrict to `127.0.0.1:9000`, so that it can only be accessed from localhost.

BYK:
We are moving away from environment variables to config files and this setting can be changed here: https://github.com/getsentry/onpremise/blob/master/sentry/sentry.conf.example.py#L151

Would that satisfy your need?

larsnystrom:
I don't think that will work, because I'm talking about the binding of host ports to container ports, and I think that setting will just change the port used inside the container? I think environment variables is actually the only way to make this configurable, because `docker-compose up` doesn't even have a command line argument for binding ports.

BYK:
Whoops, right sorry my mind was elsewhere I guess. Sounds like a reasonable suggestion, would you like to send a PR?

That said I think the alternative can be using a `docker-compose.override.yml` file and override it there. What do you think?

PR #306, opened 2019-12-27
ref(env): Add SENTRY_WEB_PORT env var (#279)

larsnystrom:
Implements #279 

This should hopefully not be a breaking change for anyone. The `.env` definition is not necessary as there is a default value encoded in `docker-compose.yml`, but I added it to `.env` to make it more obvious it can be configured.

Possible issues:
 * You may not want to expose a port for the web interface at all, for example if you are running the interface behind a TLS-proxy which is part of the same docker network. This PR does not cater to that use case. The best you can do is set the port to `127.0.0.1:9000` which would make it only available to localhost.
 * I have not fully tested this, since I'm on a train with a flaky internet-connection, so you might want to just validate that it works because Murphy's law is a PITA.


BYK:
I think this looks pretty great, thanks a lot!

Two things:

>but I added it to .env to make it more obvious it can be configured.

I'd rather remove this as you also updated the README. We may document this on https://docs.sentry.io/ later on.

>I have not fully tested this, since I'm on a train with a flaky internet-connection, so you might want to just validate that it works because Murphy's law is a PITA.

How do you feel about extending the [.travis.yml](https://github.com/getsentry/onpremise/blob/master/.travis.yml) file for using a different port as an additional test as the [test script](https://github.com/getsentry/onpremise/blob/master/test.sh) supports the `SENTRY_TEST_HOST` environment variable, just for this purpose :)

>You may not want to expose a port for the web interface at all, for example if you are running the interface behind a TLS-proxy which is part of the same docker network. This PR does not cater to that use case. The best you can do is set the port to 127.0.0.1:9000 which would make it only available to localhost.

Those people can probably remove that port declaration altogether and call it a day. I like your PR being tiny and focused while delivering a lot of value for interested folks.

@mattrobenolt - any thoughts/feedback?

mattrobenolt:
I think there are too many ways to terminate TLS, so we should be open ended enough to support anything. Whether an nginx container is run inside the Docker network, or a Kubernetes Ingress, or something like an ELB, they all have different requirements.

BYK:
@mattrobenolt - I agree and I don't think the current approach breaks anything related to that, right? It just adds an option, not for TLS termination but for customization.

mattrobenolt:
Correct. I was just answering your question. I don’t think this change is bad. 

I do think using this environment variable conflicts with the environment variable for the actual web server listening port. In some situations it’s likely better to change the actual port it’s listening on instead of just port forwarding which is pretty clumsy.

BYK:
>I do think using this environment variable conflicts with the environment variable for the actual web server listening port.

Yup, it does. @larsnystrom how about we change the variable to `SENTRY_ONPREMISE_PORT`? /cc @mattrobenolt 

>In some situations it’s likely better to change the actual port it’s listening on instead of just port forwarding which is pretty clumsy.

The current approach would definitely create some confusion around the initial setup of `system.uri-prefix` so yeah, maybe we should just go with the full change and forward this into the containers properly. @larsnystrom are you ready to dive deeper? :D

larsnystrom:
Hi,

I just wanted to say that I won't have the time to implement this anytime soon. I have enough work on my plate right now and I would rather spend my free time with my family.

If I were to implement this I would remove the `ports` section from `docker-compose.yml` altogether. Instead I would add some step to the `install.sh` script which could ask if you wanted to expose a port on the host, and if so add a `docker-compose.override.yml` with port forwarding. It could be done interactively or by default if running non-interactively. This would let people who don't deploy using docker-compose to deal with the port forwarding the way they prefer, while still keeping it simple for new users.

I don't think this feature should touch the actual server port inside the container. This should be only about port forwarding for docker-compose users. Changing the port inside the container is a different topic imho.

But yeah, I'm closing this now. Hopefully someone else has the time to make to this work.

PR #307, opened 2019-12-27
feat(features): Enable more features for on-premise

BYK:
The removed feature lines are due to these flags no longer existing in the codebase.

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -178,7 +178,6 @@ SENTRY_WEB_OPTIONS = {
 # Features #
 ############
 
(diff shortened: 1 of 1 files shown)


BYK:
I know incidents and discover2 are in beta phase but I assume we can enable it for on-premise users for more feedback. What do you think?

The flip-side would be enabling beta features that may change significantly for a less-frequently updated environment so I'm open going both ways. /cc @mitsuhiko @dcramer

dcramer:
No alpha features for self hosted. They’re usually very buggy. It’s fine on saas since we patch frequently but not ok for self hosted.

BYK:
Alright then, closing this one out. I'll do the flag clean ups in a different PR.

PR #309, opened 2019-12-30
fix(post-process): Add missing post-process-forwarder

### LINKED ISSUE #287: Mail Event-Notifications not working
So my problem is that the alerts are not send out via mail.
Even after adding the **Send notification via Mail**
image

User Feedback, Invites and Test Mails are working fine.

When looking in `docker logs sentry_worker_1 --follow` only mails for the user feedback get added.

I have used the default smtp server.

The only change I have made was adding my mail address to the config. And I had to add the hostname for the smtp container, because mail got rejected with no FQDN or IP set error.
After that I run `docker-compose build`.

```yaml
services:
  smtp:
    << : *restart_policy
    image: tianon/exim4
    hostname: ***hostname of my server**
```

Am I missing something? 

I am using `sentry, version 10.0.0.dev0`

BYK:
We were not running the post-process forwarder, causing post-process to not run which covers all plugin and rule/alert work.

Fixes #287.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -144,7 +144,8 @@ services:
     command: run worker
   post-process-forwarder:
     << : *sentry_defaults
-    command: run post-process-forwarder --commit-batch-size 3
+    # Increase `--commit-batch-size 1` below to deal with high-load environments.
+    command: run post-process-forwarder --commit-batch-size 1
   sentry-cleanup:
     << : *sentry_defaults
     image: sentry-cleanup-onpremise-local
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -142,6 +142,9 @@ services:
   worker:
     << : *sentry_defaults
     command: run worker
+  post-process-forwarder:
+    << : *sentry_defaults
+    command: run post-process-forwarder --commit-batch-size 3
   sentry-cleanup:
     << : *sentry_defaults
     image: sentry-cleanup-onpremise-local


docker-compose.yml
```diff
@@ -142,6 +142,9 @@ services:
   worker:
     << : *sentry_defaults
     command: run worker
+  post-process-forwarder:
+    << : *sentry_defaults
+    command: run post-process-forwarder --commit-batch-size 3
```
mattrobenolt: You might have trouble with batch size of 3. I’d do 1 by default for onpremise.

BYK: Tested locally, seems to work fine with 1 event. Also, if I change this, should I change it for `snuba-replacer` above?

mattrobenolt: I’m not really sure. From my understanding this wouldn’t flush until the minimum batch size. This was specifically an issue with the consumers in sentry. In snuba, there’s another flag for a min batch duration. So you could say batch size of 1000, or 1 second. So even if you don’t get 1000 messages, it’ll flush after 1 second. Last I knew about post-process-forwarder, it didn’t have the time component. So if you have a size greater than 1, it’s possible you don’t fill the batch, and it’ll never flush. This may have changed since though.

BYK: Just checked the code and still no time-component so I'm reducing this to `1` as you suggested. We may wanna add the time component soon.

mattrobenolt: I agree. I’ve been asking for it for months. :)

PR #315, opened 2020-01-03
Add configurable SENTRY_WEB_EXPOSED_PORT env variable in stable release

jaumebecks:
As a follow up from this PR https://github.com/getsentry/onpremise/pull/306, could be possible to include this similar solution to the stable version?

I also have the need to customize the exposed port of the sentry web container due to some incompatibilities with XDebug, which uses the same 9000 port. I'd be nice to have the customization of this variable also in the stable release :)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
 # Run `docker-compose run web config generate-secret-key`
 # to get the SENTRY_SECRET_KEY value.
 SENTRY_SECRET_KEY=
+SENTRY_WEB_EXPOSED_PORT=9000
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,7 +51,7 @@ services:
   web:
     <<: *defaults
     ports:
-      - '9000:9000'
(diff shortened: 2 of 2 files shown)


BYK:
Hey, thanks for the PR!

I think this is fine but if we also try to address [my other comment](https://github.com/getsentry/onpremise/pull/306#issuecomment-569786899):

>The current approach would definitely create some confusion around the initial setup of `system.uri-prefix` so yeah, maybe we should just go with the full change and forward this into the containers properly. @larsnystrom are you ready to dive deeper? :D

We'd end up having multiple variables to control. Are you okay trying to propagate `SENTRY_WEB_PORT` into the containers and changing the port inside and outside the container this way for full consistency?

BYK:
> could be possible to include this similar solution to the stable version?

This will be the stable version _very_ soon :)

BYK:
Oh, wait you opened this PR towards `stable`. Sorry, that ain't changing so closing this :)

PR #317, opened 2020-01-03
fix(migration): Move Snuba bootstrapping before DB upgrade

BYK:
Since we are [moving the eventstream into a proper db migration](https://github.com/getsentry/sentry/pull/16226), we need Snuba and its friends ready at the time of `sentry upgrade` command. This patch does exactly that.

Unblocks https://github.com/getsentry/sentry/pull/16226

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,6 +101,16 @@ $dc build --force-rm
 echo ""
 echo "Docker images built."
 
+echo "Boostrapping Snuba..."
+$dc up -d kafka redis clickhouse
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -101,6 +101,16 @@ $dc build --force-rm
 echo ""
 echo "Docker images built."
 
+echo "Boostrapping Snuba..."
```
lynnagara:
```suggestion
echo "Bootstrapping Snuba..."
```

BYK: lol, thanks! :D

PR #327, opened 2020-01-08
feat(config): Have better GitHub config examples

BYK:
We need the `GITHUB_APP_ID` and `GITHUB_API_SECRET` configs in `sentry.conf.py` for GitHub auth whereas we need the `github-app.*` values in `config.yml` for the GitHub integration. The ones in `sentry.conf.py` are redundant but since `config.yml` is loaded _after_ `sentry.conf.py` I added this `__get__` hack so the users only need to set the values in one place.


sentry/sentry.conf.example.py
```diff
@@ -209,13 +209,24 @@
     }
 )
 
+# Hacky stuff
```
mattrobenolt: 🤔 I'd be scared of this causing an issue in the future if we were to convert these into real options.

I'm also don't actually see how this works at all.

```python
>>> class A(object):
...   def __get__(self, instance, owner):
...     return 'foo'
...
>>> foo = A()
>>> foo
<__main__.A object at 0x10ca97ef0>
>>> str(foo)
'<__main__.A object at 0x10ca97ef0>'
```

BYK: >I'd be scared of this causing an issue in the future if we were to convert these into real options.

I don't think we can drop support for the old configs easily. Even then, this would simply be harmless.

>I'm also don't actually see how this works at all.

The trick is, although we define these like module locals, they are actually bound to a dynamically created class instance: the `settings` instance from `django.conf`. So in reality, we are doing
```py
>>> class A(object):
...   def __get__(self, instance, owner):
...     return 'foo'
...
>>> class B(object):
...  foo = A()
>>> bar = B()
>>> bar.foo
```

which then triggers `__get__` and makes this work (tested locally already).

sentry/sentry.conf.example.py
```diff
@@ -209,13 +209,24 @@
...
+class SentryOptionsMapper(object):
+    def __init__(self, option_name):
+        self._option_name = option_name
+
+    def __get__(self, instance, owner):
+        if not hasattr(self, '_option_val'):
+            from django.conf import settings
+            self._option_val = settings.SENTRY_OPTIONS.get(self._option_name)
+        return self._option_val
```
scefali: @BYK Wondering why we need caching here. If something in `settings.SENTRY_OPTIONS` changes for any reason, we won't see that change.

mattrobenolt: It's immutable after boot. And this import isn't free. In theory, this is just memoization.

sentry/sentry.conf.example.py
mattrobenolt: lol dang, I see. Yeah. I'd still like to see this just done inside the initializer rather than this hack as a one-off for onpremise.

sentry/sentry.conf.example.py
scefali: @mattrobenolt good point about the import, I agree this way makes sense

BYK:
>I feel like this is more suitable to just handle in here: getsentry/sentry:src/sentry/runner/initializer.py@master
>
>And do it correctly instead of a hack here. We have a few places where we convert to/from options for things.

I agree. I'd even go further to make the actual auth plugin read the values from options. That said that only fixes the builds after this, which doesn't include the tagged v10 release. If I go there, I'd also love to load `config.yml` before `sentry.conf.py` which would make many other things easier.

Are you okay doing both: merge this PR and fix this upstream in getsentry/sentry at the same time? I expect fixing this over at getsentry/sentry take a bit more time anyway so this will cover us, "stop the bleeding", in the meantime.

BYK:
Closing in favor of getsentry/sentry#16550.

BYK:
Don't know why I closed this. I'll just remove the hacky parts and the rest still applies.

Issue #358, opened 2020-01-22
Prevent installation on old AMD servers ?


Amphaal:
Hi !

I tried to install Sentry on an small homeserver of mine, a kinda rusty HPE ProLiant MicroServer, runing a Turion II Neo N40L CPU. The issue here is that this CPU does not handle SSE 4.2 instructions, and thus cannot run a classic image of Clickhouse (https://clickhouse.yandex/docs/en/operations/requirements/).

Trying to install Sentry via ./install.sh in this context leads to confusing container errorlogs and timeouts during installation, which let you clueless on why the whole process failed...

As memory capacity is checked at script startup, can we also check if SSE 4.2 is handled ?

BYK:
This is a good suggestion, I'll see if we can check this. Thanks!

BYK:
Btw. if you are willing to give this a shot, be my guest please :)

Amphaal:
There you go !

PR #361, opened 2020-01-22
Check SSE4.2 compatibility before installation

Amphaal:
As asked by @BYK on #358, something like that can do the trick ;)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -65,7 +65,7 @@ fi
 #SSE4.2 required by Clickhouse (https://clickhouse.yandex/docs/en/operations/requirements/) 
 SSE_42_COMPAT="$(grep -c "sse4_2" /proc/cpuinfo)"
 if (($SSE_42_COMPAT == 0)); then
-    echo "FAIL: The CPU your machine is running on does not handle SSE 4.2 instructions, and thus cannot use Sentry 10"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -62,6 +62,13 @@ if [ "$RAM_AVAILABLE_IN_DOCKER" -lt "$MIN_RAM" ]; then
     exit -1
 fi
 
+#SSE4.2 required by Clickhouse (https://clickhouse.yandex/docs/en/operations/requirements/) 
+SSE_42_COMPAT="$(grep -c "sse4_2" /proc/cpuinfo)"
```
BYK: This will fail on a macOS so I think we should do something similar to the memory detection command above:
```bash
SUPPORTS_SSE42=$(docker run --rm busybox grep -c sse4_2 /proc/cpuinfo);
```

install.sh
```diff
@@ -62,6 +62,13 @@ if [ "$RAM_AVAILABLE_IN_DOCKER" -lt "$MIN_RAM" ]; then
...
+if (($SSE_42_COMPAT == 0)); then
+    echo "FAIL: The CPU your machine is running on does not handle SSE 4.2 instructions, and thus cannot use Sentry 10"
```
BYK:
```suggestion
    echo "FAIL: The CPU your machine is running on does not support the SSE 4.2 instruction set, which is required for one of the services Sentry uses (Clickhouse). See https://git.io/JvLDt for more info."
```

PR #371, opened 2020-01-30
build(clickhouse): Bump clickhouse version to match Snuba

BYK:


BYK:
>I assume this is fine to drop in without doing anything to upgrade the data or anything.

Yup, same assumption here.

PR #390, opened 2020-02-21
Make secret key generation idempotent

prutseltje:
Issue: I'm deploying sentry onpremise via ansible. After I make changes to the configuration I re-run the install script to update the configuration. But every run of the install script will re-generate the secret key, and therefore trigger the (update) install handler every ansible run.

Solution: Only generate a secret key if the default key has not been changed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -90,8 +90,8 @@ ensure_file_from_example $SENTRY_CONFIG_PY
 ensure_file_from_example $SENTRY_CONFIG_YML
 ensure_file_from_example $SENTRY_EXTRA_REQUIREMENTS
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -91,13 +91,15 @@ ensure_file_from_example $SENTRY_CONFIG_YML
 ensure_file_from_example $SENTRY_EXTRA_REQUIREMENTS
 
 echo ""
```
BYK: I think this `echo` should go into the `if` block too :)

prutseltje: Yes it should, I've changed it

PR #408, opened 2020-03-11
feat(snuba): Remove redundant UWSGI settings

BYK:
These became obsolete after getsentry/snuba#825

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,7 @@ x-snuba-defaults: &snuba_defaults
     DEFAULT_BROKERS: 'kafka:9092'
     REDIS_HOST: redis
     UWSGI_MAX_REQUESTS: '10000'
+    UWSGI_DISABLE_LOGGING: 'true'
 services:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -34,15 +34,7 @@ x-snuba-defaults: &snuba_defaults
     CLICKHOUSE_HOST: clickhouse
     DEFAULT_BROKERS: 'kafka:9092'
     REDIS_HOST: redis
-    # TODO: Remove these after getsentry/snuba#353
     UWSGI_MAX_REQUESTS: '10000'
-    UWSGI_DISABLE_LOGGING: 'true'
```
mattrobenolt: This is up to you if you want logging or not, but default is going to bring back logging. So if you want to keep logging disabled, then keep this one. I personally thing the logging is useful in production instances.

BYK: I'll keep logging disabled for now as it may become noisy and cause people's disks to fill up.

PR #421, opened 2020-03-30
feat(relay): Add Relay to onpremise installation

RaduW:
This PR adds the necessary services necessary to have Sentry working with Relay message ingestion.


docker-compose.yml
```diff
@@ -152,6 +153,30 @@ services:
       args:
         BASE_IMAGE: 'sentry-onpremise-local'
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
+  reverse_proxy:
+    ports:
+      - '9000:80/tcp'
+    image: "nginx:1.16.1"
+    volumes:
+      - type: bind
```
BYK:
```suggestion
      - type: bind
        read_only: true
```

Let's make the config folder read-only.

docker-compose.yml
```diff
@@ -152,6 +153,30 @@ services:
...
+        source: ./reverse_proxy_config/nginx.conf
+        target: /etc/nginx/nginx.conf
```
BYK:
```suggestion
        source: ./nginx
        target: /etc/nginx
```

relay_config/config.yml
```diff
@@ -0,0 +1,16 @@
+---
+relay:
+  upstream: "http://web:9000/"
+  host: 0.0.0.0
+  port: 3000
+logging:
+  level: TRACE
```
BYK:
```suggestion
  level: WARN
```

Let's not trace as it would be too noisy and consume a lot of disk space.

relay_config/config.yml
```diff
@@ -0,0 +1,16 @@
...
+  enable_backtraces: false
+processing:
+  enabled: true
+  kafka_config:
+    # Original kafka configuration was 9093 but it seems that the
+    # compose configures kafka to listen at 9092 ?
+    # - {name: "bootstrap.servers", value: "kafka:9093"}
```
BYK: Let's remove this comment :) It is what it is on on-premise.
```suggestion
```

relay_config/credentials.json
```diff
@@ -0,0 +1,5 @@
+{
+  "secret_key": "xWWsHUXG6RxHS9L7QWPrP0aDgVLKSa7vcE4Ypa-Vv44",
```
BYK: We should definitely add this file to `.gitignore` and generate it at install time via `install.sh`.

docker-compose.yml
```diff
@@ -152,6 +153,30 @@ services:
...
+    depends_on:
+      - web
+      - relay
+  relay:
+    image: "us.gcr.io/sentryio/relay:latest"
+    ports:
+      - '3000:3000/tcp'
```
BYK: I don't think we need to expose any ports from `relay` to outside world as it goes through nginx?
```suggestion
```

docker-compose.yml
```diff
@@ -152,6 +153,30 @@ services:
...
+    command: 'run --config /etc/relay'
+    volumes:
+      - type: bind
+        source: ./relay_config
+        target: /etc/relay
+    depends_on:
+      - kafka
+      - redis
+      - web
```
BYK: Why do we depend on `web` service? I don't think we do.
```suggestion
```

docker-compose.yml
```diff
@@ -152,6 +153,30 @@ services:
...
+      - relay
+  relay:
+    image: "us.gcr.io/sentryio/relay:latest"
+    ports:
+      - '3000:3000/tcp'
+    command: 'run --config /etc/relay'
+    volumes:
+      - type: bind
+        source: ./relay_config
+        target: /etc/relay
```
BYK:
```suggestion
        read_only: true
        source: ./relay
        target: /etc/relay
```

reverse_proxy_config/nginx.conf
```diff
@@ -0,0 +1,67 @@
...
+pid        /var/run/nginx.pid;
+
+
+events {
+    worker_connections  1024;
+}
+
+
+http {
+    include       /etc/nginx/mime.types;
```
BYK: Do we really need this file?

reverse_proxy_config/nginx.conf
```diff
@@ -0,0 +1,67 @@
...
+    default_type  application/octet-stream;
+
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
```
BYK: We may want to not keep access logs as they may consume too much disk space.

reverse_proxy_config/nginx.conf
```diff
@@ -0,0 +1,67 @@
...
+
+    sendfile        on;
+    #tcp_nopush     on;
```
BYK: Let's not have commented-out code. Do we need this on or off?

reverse_proxy_config/nginx.conf
```diff
@@ -0,0 +1,67 @@
...
+
+    keepalive_timeout  65;
+
+    #gzip  on;
```
BYK: Same as above. Sentry should already be doing gzip/compression so I need to ask @mattrobenolt how this should be configured with the pre-gzipped static assets. I think we need to point Nginx there or just keep letting uwsgi do its magic.

reverse_proxy_config/nginx.conf
```diff
@@ -0,0 +1,67 @@
...
+        server relay:3000;
+    }
+
+    upstream sentry {
+        server web:9000;
+    }
+
+    server {
+        listen 80;
+        resolver 127.0.0.11 ipv6=off;
```
BYK: Can you add a comment about this specific resolver IP address: where does it come from?

docker-compose.yml
```diff
@@ -152,6 +153,30 @@ services:
       args:
         BASE_IMAGE: 'sentry-onpremise-local'
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
+  reverse_proxy:
```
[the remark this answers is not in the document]
jan-auer: @RaduW 👍 I think that might also be a more descriptive name for for the dev setup.

docker-compose.yml
jan-auer: Relay requires this for authentication and to pull project configs.

BYK: We chatted on this with @RaduW -- Unless this is a hard-requirement for the service to run, we'd rather not list it as when we need to run a relay command like `docker-compose run --rm relay credentials`, it will then try to bring up the Sentry web service, which depends on Postgres, Kafka, Snuba etc. which is something we don't really need.

docker-compose.yml
RaduW: ok

relay_config/config.yml
RaduW: no logging configuration by default, 
leave only a commented out configuration so the user can easily change it to whatever.

relay_config/config.yml
RaduW: yep

docker-compose.yml
RaduW: you are right

docker-compose.yml
RaduW: Removed web dependency as we do retries for authentication with web so it is not critical that web is already up by the time relay tries to authenticate with it.

reverse_proxy_config/nginx.conf
RaduW: agreed

reverse_proxy_config/nginx.conf
RaduW: ok, still left the error_log though

docker-compose.yml
RaduW: ok

docker-compose.yml
RaduW: ok

reverse_proxy_config/nginx.conf
RaduW: ok

relay_config/credentials.json
RaduW: now generated at installation time

jan-auer:
Please see changes in https://github.com/getsentry/sentry/pull/18390: 
- There's a new envelope endpoint (introduced in https://github.com/getsentry/relay/pull/552) that needs to be forwarded.
- We need to introduce a size limit of at least `32m` for the chunk-upload endpoint to work.
- We should probably reconsider size limits of other endpoints, or just generally increase it to some higher value to get those out of the way.

If you'd like me to, I can make the changes to this PR.

jan-auer:
Another option would be to make a wildcard forward like in https://github.com/getsentry/sentry/pull/18433. By definition, all of these endpoints need to be handled by Relay, so we can as well forward them. The benefit of this is that Relay responds with `404` on non-existing endpoints, while Sentry incorrectly responds with `403`.

.travis.yml
```diff
@@ -14,5 +14,9 @@ script:
   - ./install.sh
   - docker-compose run --rm web createuser --superuser --email test@example.com --password test123TEST
   - docker-compose up -d
+  - docker container ls
```
BYK, line 17: Please revert all the changes in this file as they seem for debugging purposes.

docker-compose.yml
```diff
@@ -161,6 +162,28 @@ services:
       args:
         BASE_IMAGE: 'sentry-onpremise-local'
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
+  nginx:
+    ports:
+      - '9000:80/tcp'
+    image: "nginx:1.16.1"
```
BYK, line 168: Should we go with `1.16` so we'd get patch releases automatically? Also the latest version seems to be `1.17`.

install.sh
```diff
@@ -164,6 +166,35 @@ if [ "$SENTRY_DATA_NEEDS_MIGRATION" ]; then
...
 
+
+if [ ! -f "$RELAY_CREDENTIALS_JSON" ]; then
+    echo ""
+    echo "Generating Relay credentials..."
+
+    $dcr --user $(id -u) relay --config /etc/relay credentials generate --overwrite
+    chmod a+r $RELAY_CREDENTIALS_JSON
+    # display the contents of the relay directory (for debug purposes)
+    ls -al ./relay
```
BYK, line 177: Let's remove these :)
```suggestion
```

install.sh
```diff
@@ -164,6 +166,35 @@ if [ "$SENTRY_DATA_NEEDS_MIGRATION" ]; then
...
+    CREDENTIALS=$(sed -n 's/^.*"public_key":[[:space:]]"\([a-zA-Z0-9_-]*\)".*$/\1/p' "$RELAY_CREDENTIALS_JSON")
```
BYK, line 178: I think the space here should be followed by a `*` as it is optional in JSON (or can be repeated). Also note the `*` -> `+` change for the actual credentials value matcher.
```suggestion
    CREDENTIALS=$(sed -n 's/^.*"public_key":[[:space:]]*"\([a-zA-Z0-9_-]+\)".*$/\1/p' "$RELAY_CREDENTIALS_JSON")
```

install.sh
```diff
@@ -164,6 +166,35 @@ if [ "$SENTRY_DATA_NEEDS_MIGRATION" ]; then
...
+    CREDENTIALS="SENTRY_RELAY_WHITELIST_PK = [\"$CREDENTIALS\"]"
+
+    if grep -xq SENTRY_RELAY_WHITELIST_PK "$SENTRY_CONFIG_PY"; then
+        echo "FAIL: SENTRY_RELAY_WHITELIST_PK already exists in $SENTRY_CONFIG_PY, please replace with:"
+        echo ""
+        echo "  $CREDENTIALS"
+        echo ""
```
BYK, line 185: Maybe we should print this to stderr instead?
```suggestion
        >&2 echo "FAIL: SENTRY_RELAY_WHITELIST_PK already exists in $SENTRY_CONFIG_PY, please replace with:"
        >&2 echo ""
        >&2 echo "  $CREDENTIALS"
        >&2 echo ""
```

install.sh
```diff
@@ -164,6 +166,35 @@ if [ "$SENTRY_DATA_NEEDS_MIGRATION" ]; then
...
+        exit 1
+    fi
+
+     echo "" >> "$SENTRY_CONFIG_PY"
+     echo "$CREDENTIALS" >> "$SENTRY_CONFIG_PY"
+
+     tail -n 5 "$SENTRY_CONFIG_PY"
```
BYK, line 192: Please remove debug lines.
```suggestion
```

install.sh
```diff
@@ -164,6 +166,35 @@ if [ "$SENTRY_DATA_NEEDS_MIGRATION" ]; then
...
+else
+     echo ""
+     echo "Relay credentials exist will NOT generate new ones."
+fi
```
BYK, line 196: We don't say anything when we don't generate a secret key (because it already exists). Let's be consistent and say nothing if we do nothing.
```suggestion
fi
```

nginx/nginx.conf
```diff
@@ -0,0 +1,62 @@
...
+            proxy_pass http://relay;
+        }
+        location ~ ^/api/\d+/security/$ {
+            proxy_pass http://relay;
+        }
+        location ~ ^/api/\d+/csp-report/$ {
+            proxy_pass http://relay;
+        }
+        location ~ ^/api/\d+/events/[\w-]+/attachments/$ {
+            proxy_pass http://relay;
```
BYK, line 56: Let's replace this with the wildcard version from #8433 as @jan-auer suggested.

sentry/sentry.conf.example.py
```diff
@@ -220,3 +220,4 @@
 
 # BITBUCKET_CONSUMER_KEY = 'YOUR_BITBUCKET_CONSUMER_KEY'
 # BITBUCKET_CONSUMER_SECRET = 'YOUR_BITBUCKET_CONSUMER_SECRET'
+
```
BYK, line 223: Revert please?

nginx/nginx.conf
```diff
@@ -0,0 +1,62 @@
...
+
+    upstream sentry {
+        server web:9000;
+    }
+
+    server {
+        listen 80;
+        # use the docker DNS server to resolve ips for relay and sentry containers
+        resolver 127.0.0.11 ipv6=off;
+
```
BYK, line 36: Let's also squeeze in the max upload size limit change @jan-auer mentioned somewhere around here:
```suggestion
        client_max_body_size 32M;
```

.travis.yml
RaduW: I think they should stay, when the build is green you don't really need to look at the log. When there is a problem you want to see what is going on from the log without needing to do another commit. Since the logging commands do not add any significant time overhead  I would vote for leaving them where they are.

docker-compose.yml
RaduW: I don't have an opinion on this one, so it is your call. 
I was told to 'fix' the image versions for external dependencies so we get consistent/repeatable environments.

install.sh
RaduW: As mentioned before I think they don't hurt but I can remove them if you insist :-).

install.sh
RaduW: ok

nginx/nginx.conf
RaduW: Yes we should, I double checked in Sentry and I couldn't find any urls ^/api/[1-9]\d*/... the only ones close are /api/0/...

sentry/sentry.conf.example.py
RaduW: Added a space because the script inserts a new line at the end of the file with the known Relay keys (and it looks bad if it comes immediately after the comment).

install.sh
RaduW: same comment as above (about adding additional logs to the script).

.travis.yml
BYK: I'm okay doing `docker-compose logs` in the `after_failure` handler in TravisCI but simply listing the containers with every run is not very useful. Also, we should be keeping the PR to the point, which is implementing Relay so please revert these changes.

docker-compose.yml
BYK: We are okay fixing the images to minor versions which we do with `postgres` and `redis` already so let's change this to `1.16` to be consistent with others.
```suggestion
    image: "nginx:1.16"
```

install.sh
BYK: They do hurt as they output unnecessary information to everyone, who are usually not interested in this information or be able to understand what this means.

install.sh
BYK: No extra and random log outputs please. Also keeping the PR focused per my other comment.

sentry/sentry.conf.example.py
BYK: Okay

install.sh
RaduW: Sorry, I really don't get your point about not sending a message when not generating a secret key. We generate the whole configuration as a unit that includes ( id, public key and secret key). 
To me the message is very informative and make a lot of sense. 

Having a credentials.json already existing at install time would be unusual. 
Maybe the user has already run install.sh previously and is running it again in the hope that everything will be reset. In this case we should mention that we have found some existing keys and we are not going to mess with them again ( so if he has been messing with the Relay keys, re running the script will not help him).

install.sh
RaduW: Did you test it on Mac? 
Maybe we have a different version of sed installed because it doesn't work on my Mac ( I had quite a few problems with the Mac sed) it seems to only recognise a very limited regex set `+` doesn't seem to work. 

I could have used `[a-zA-Z0-9_-][a-zA-Z0-9_-]*` but it didn't seem to be worth my while. 

I added the optional number of spaces suggested `[[:space:]]*` but since this is the output of a program generated file I don't think we really need to support all possible correct json files ( which we don't anyway, since we don't support spaces before `:` or the key and the value being spread on two different lines.

install.sh
BYK: We do not output anything when we don't generate a secure key: https://github.com/getsentry/onpremise/blob/c20956527c5560625bb8b83fd605ee0b742a4ccf/install.sh#L93-L102

I'm saying we should be consistent with that behavior and simply don't output anything if we don't do anything. Remember, this maybe useful for _you_ as a developer but this repo is for end users who don't necessarily know or care about Relay or its credentials.

install.sh
BYK: >Did you test it on Mac?

Nope as I'm using Windows + Linux. Seems like `+` requires the extended regexp set which needs different flags for Linux and MacOS (`-r` and `-E` respectively). So instead we can do `[a-zA-Z0-9_-]\{1,\}`.

I think this is important because it expresses our intent more clearly. I'd prefer a fixed length if the credentials string has a fixed length like a hash for precision.

BYK: >I added the optional number of spaces suggested [[:space:]]* but since this is the output of a program generated file I don't think we really need to support all possible correct json files ( which we don't anyway, since we don't support spaces before : or the key and the value being spread on two different lines).

Good point, we should probably add a `[[:space:]]*` before the colon too. The output of the program is JSON and the formatting of the JSON file can change with a newer Rust version, a different JSON library or some other config change in that program so I think trying to cover these basics are important.

docker-compose.yml
RaduW: ok

.travis.yml
BYK: Created https://github.com/getsentry/onpremise/pull/443

jan-auer:
❤️ 🎉  Thanks both!

kagansari:
You may want to review this. I got permisson denied error on `/etc/relay` during installation

jan-auer:
@BYK Seems like we missed this since there is no cloudbuild check configured. Is there a way to have a GH check on both `getsentry/sentry` and `getsentry/onprem`?

BYK:
@kagansari yeah, sorry for the disruption we are aware there are issues under certain circumstances. We relied on our local environment and then Travis CI for this which both passed. Should be resolved today.

PR #422, opened 2020-03-30
fix(clickhouse): Bind a proper tmpfs for clickhouse tmp

BYK:
Defined/documented here: https://github.com/ClickHouse/ClickHouse/blob/621bcaf2eba65d7352ac3bbc22ad1e77965977aa/dbms/programs/server/config.xml#L115

Follow up to #414.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -99,6 +99,8 @@ services:
     volumes:
       - 'sentry-clickhouse:/var/lib/clickhouse'
       - 'sentry-clickhouse-log:/var/log/clickhouse-server'
+      - type: tmpfs
+        target: '/var/lib/clickhouse/tmp'
   snuba-api:
     << : *snuba_defaults
   snuba-consumer:


mattrobenolt:
Why? Whats this solve and seems unrelated to #414.

BYK:
https://docs.docker.com/storage/
>Writing into a container’s writable layer requires a storage driver to manage the filesystem. The storage driver provides a union filesystem, using the Linux kernel. *This extra abstraction reduces performance as compared to using data volumes, which write directly to the host filesystem.*

https://docs.docker.com/storage/storagedriver/
>Storage drivers allow you to create data in the writable layer of your container. The files won’t be persisted after the container is deleted, and *both read and write speeds are lower than native file system performance.*

The only downside I see with this is `tmpfs` only being available on Linux so we may want to swap that with a good-old regular volume mount.

mattrobenolt:
This doesn’t answer my question at all. What does this have to do with logging? And why do we need to explicitly bind tmpfs? If data is written into the container, it’s already temporary since it’s ephemeral to the container.

BYK:
>What does this have to do with logging?

This has to do with writing data into the container filesystem, which is considered bad practice due to many reasons - performance and storage size being the top concerns. Portability is also an issue.

>And why do we need to explicitly bind tmpfs? If data is written into the container, it’s already temporary since it’s ephemeral to the container.

We don't really _need_ to do this as it has been working fine. But using a tmpfs mount makes the I/O on this path more efficient and also does not consume space on the host machine forever (until the container is deleted, which is quite rare).

We may further add a size limit to this mount to make it more useful and safe.

mattrobenolt:
> This has to do with writing data into the container filesystem, which is considered bad practice due to many reasons - performance and storage size being the top concerns. Portability is also an issue.

I don’t agree with any of this. Your change doesn’t solve any of this. Nor are these actually problems.

> We don't really need to do this as it has been working fine. But using a tmpfs mount makes the I/O on this path more efficient and also does not consume space on the host machine forever (until the container is deleted, which is quite rare).

tmpfs is still going to consume the same space. If anything, this might actually push it into memory since typically a true tmpfs is memory, not filesystem. I have to imagine there’s a reason Clickhouse is using their own tmp folder and not actual /tmp. You already have /var/lib/clickhouse (the parent) mounted with a real disk, so IO isn’t a concern.

> We may further add a size limit to this mount to make it more useful and safe.

I would only assume this is a bad idea. If ClickHouse needs the working space and it can’t get it, it’ll probably error.

BYK:
Alright, all fair points so closing the PR 😃

PR #429, opened 2020-04-03
redis - maxmemory 2gb default; allkeys-lru policy; disabled save

petrprikryl:
Hi, I was having this problem with Redis https://stackoverflow.com/questions/19581059/misconf-redis-is-configured-to-save-rdb-snapshots. And I have resolved it using this configuration recommendations https://docs.sentry.io/server/performance/#redis. So maybe it should be set as default?


petrprikryl:
My problem was that on virtual machine where our Sentry is running wasn't enough space on disk so Redis couldn't make save. I have disabled the saving on Redis and then enlarged the disk on virtual machine so the problem have disappeared. But it is question anyway if the data lose here will have dramatic impact on Sentry. If so then I agree that the docs should be updated and you can close the PR :-).

mattrobenolt:
It does have a dramatic impact, but that's subjective to you if you care or not. It's not going to prevent you from using the product, but it's not something we can recommend or include out of the box. This is very much you choosing to accept data loss from simply restarting the process. One that I wouldn't recommend by default for our solution.

BYK:
Closing as we are not planning to merge this. Thanks again @petrprikryl.

PR #430, opened 2020-04-06
feat: Instruct users to migrate TSDB

untitaker:
Tested this in a Ubuntu VM. The output of `date` is not too pretty but at least localized (so D/M vs M/D is not confusing)

## What is the TSDB migration?

We're effectively deprecating all TSDB backends but `sentry.tsdb.redissnuba.RedisSnubaTSDB`. We cannot reasonably support any other backend due to the fact that we would have to reimplement each of the backends in Relay, which is written in a different language. Also, like with deprecating mysql support, we don't really have the capacity to support things we do not use ourselves.

## Migration

`install.sh` should rewrite your configuration automatically and define a cutover date such that no data is lost. Before the cutover date, data is written to two backends at once, Redis and Snuba, and read from one, Redis. After the cutover date, event-related metrics will be read from Snuba which matches what we have on sentry.io.

## Manual migration guide for TSDB

In case `install.sh` is unable to migrate your files you will be given basic instructions on the console that essentially tell you to completely delete all TSDB config and paste the new, standard one. If for some reason you cannot say goodbye to your existing TSDB config, please create a new issue in this repo and cc @untitaker on it.


filesnate:
IMO, the instructions in the script are confusing. 

Remove SENTRY_DSB and SENTRY_DSB_OPTIONS:
```
SENTRY_TSDB = "sentry.tsdb.redissnuba.RedisSnubaTSDB"
```
And replace with
```
SENTRY_TSDB = "sentry.tsdb.redissnuba.RedisSnubaTSDB"
```
In both case, the value for SENTRY_DSB is identical.

What happens after 90 days?  Why not remove it directly?  Are there other implications?  Has something replaced this functionality?

(The normal docs still reference tsdb, so should they be updated as well)
https://docs.sentry.io/server/tsdb/
https://hub.docker.com/r/getsentry/snuba

untitaker:
This message only shows if the existing value in the config is *not* RedisSnubaTSDB.

RedisSnubaTsdb basically dual-writes for 90 days (to redis and snuba) while reading from Redis, and after the specified timestamp it starts reading from Snuba which should now have all relevant data.

filesnate:
You are correct, I missed the `!` in the check.

I think the `#Remove` could be made more obvious by stating that ONLY the `SENTRY_DSB_OPTIONS` setting needs to be removed.

```
"  # Automatic switchover 90 days after $(date). SENTRY_TSDB_OPTIONS  line can be removed afterwards."
```

untitaker:
But that's not right @filesnate , you need to get rid of both options, or at least the values of them, and replace them with this.

untitaker:
@BYK this is ready to review again after your recent comments.

@filesnate holding off of documenting this as development docs are going through an overhaul atm

.gitignore
```diff
@@ -75,6 +75,7 @@ data/
 
 # custom Sentry config
 sentry/sentry.conf.py
+sentry/*.bak
```
BYK, line 78: Nit: this should probably come after `sentry/config.yml` to make the comment above relevant.

install.sh
```diff
@@ -71,6 +71,45 @@ if (($SUPPORTS_SSE42 == 0)); then
     exit 1
 fi
 
+replace_tsdb() {
+    if (
+        [ -f "$SENTRY_CONFIG_PY" ] &&
```
BYK, line 76: You should not need to check this as we ensure the file exists before proceeding?

install.sh
```diff
@@ -71,6 +71,45 @@ if (($SUPPORTS_SSE42 == 0)); then
...
+        ! grep -xq 'SENTRY_TSDB = "sentry.tsdb.redissnuba.RedisSnubaTSDB"' "$SENTRY_CONFIG_PY"
+    ); then
+        tsdb_settings="SENTRY_TSDB = \"sentry.tsdb.redissnuba.RedisSnubaTSDB\"
+
+# Automatic switchover 90 days after $(date). Can be removed afterwards.
+SENTRY_TSDB_OPTIONS = {\"switchover_timestamp\": $(date +%s) + (90 * 24 * 3600)}"
+
+        if grep -q 'SENTRY_TSDB_OPTIONS = ' "$SENTRY_CONFIG_PY"; then
+            echo "Not attempting automatic TSDB migration due to presence of SENTRY_TSDB_OPTIONS"
```
BYK, line 85: I'd say we should limit this case to only when `switchover_timestamp` is defined. The mere existence of `SENTRY_TSDB_OPTIONS` is to large of a blanket, right?

Also we should probably point people to some useful page explaining what is happening. The link can be this PR itself if it has a meaningful body.

install.sh
```diff
@@ -71,6 +71,45 @@ if (($SUPPORTS_SSE42 == 0)); then
...
+            tsdb_settings="${tsdb_settings//$'\n'/\\n}"
+            cp "$SENTRY_CONFIG_PY" "$SENTRY_CONFIG_PY.bak"
+            sed -i -e "s/^SENTRY_TSDB = .*$/${tsdb_settings}/g" "$SENTRY_CONFIG_PY" || true
+
+            if grep -xq 'SENTRY_TSDB = "sentry.tsdb.redissnuba.RedisSnubaTSDB"' "$SENTRY_CONFIG_PY"; then
+                echo "Migrated TSDB to Snuba. Old configuration file backed up to $SENTRY_CONFIG_PY.bak"
+                return
+            fi
+
+            echo "Failed to automatically migrate TSDB. Reverting."
```
BYK, line 98:
```suggestion
            echo "Failed to automatically migrate TSDB. Reverting..."
```

install.sh
```diff
@@ -71,6 +71,45 @@ if (($SUPPORTS_SSE42 == 0)); then
...
+            mv "$SENTRY_CONFIG_PY.bak" "$SENTRY_CONFIG_PY"
```
BYK, line 99:
```suggestion
            mv "$SENTRY_CONFIG_PY.bak" "$SENTRY_CONFIG_PY"
            echo "$SENTRY_CONFIG_PY restored from backup."
```

install.sh
```diff
@@ -71,6 +71,45 @@ if (($SUPPORTS_SSE42 == 0)); then
...
+        fi
+
+        echo "FAIL: Your Sentry configuration uses a legacy data store for time-series data. Remove the options SENTRY_TSDB and SENTRY_TSDB_OPTIONS from $SENTRY_CONFIG_PY and add:"
```
BYK, line 102: Still not sure if this script is the right place to fail. I'd say once we remove this TSDB, we can error out in Sentry like other things?

install.sh
```diff
@@ -71,6 +71,45 @@ if (($SUPPORTS_SSE42 == 0)); then
...
+        echo ""
+        echo "$tsdb_settings"
+        echo ""
+        echo "Then rerun this script."
+        exit 1
+    fi
+}
+
+replace_tsdb
```
BYK, line 111: I'd move this way down, at least after we turn off the potentially running services (see lines 115 and 117).

Ideal location seems to be around line 144, after we ensured a config file and a `system.secret-key`.

install.sh
untitaker: The error message in Sentry will probably only appear in docker logs, right? This does not seem like a userfriendly way to inform them their config is broken.

One of the goals of this PR Is to unblock removal of Redis TSDB

BYK: >The error message in Sentry will probably only appear in docker logs, right? This does not seem like a userfriendly way to inform them their config is broken.

Yes and I agree that it is not the best way to surface. That being said, for this to be consistent, we'd need a full config audit in the install script. I'd still go with a warning and not failing the script completely.

untitaker: I actually don't know based on which condition I would let Sentry fail. This error is only triggered if the automatic migration failed which I deem unlikely. I would like to encourage users to migrate asap as any delay (incurred by the user ignoring warnings) extends the timeframe for which we have to keep the redis tsdb around. We do have to keep the redis tsdb for at least 90 days (prob much more) because migrating redis data over to snuba takes that long for users.

So to me the outcome of not erroring in the install script would seem to be:

* users who need to manually migrate will overlook the warning
* when we start erroring in sentry eventually, the error will be invisible because it will occur in a docker container

BYK: >I actually don't know based on which condition I would let Sentry fail. This error is only triggered if the automatic migration failed which I deem unlikely.

I concur. It seems unlikely for someone to hit this case.

>I would like to encourage users to migrate asap

Don't we already by automatically migrating them when possible?

> as any delay (incurred by the user ignoring warnings) extends the timeframe for which we have to keep the redis tsdb around. We do have to keep the redis tsdb for at least 90 days (prob much more) because migrating redis data over to snuba takes that long for users.

Not necessarily. Soon we'll have named releases which we can point people to, if they need to migrate or stay on Redis. Even without named releases, we can point them to a specific revision where we still have the RedisTSDB code.

>users who need to manually migrate will overlook the warning

I'd argue that if you have modified the TSDB setting and need a manual migration, you would be a person unlikely to look past a warning or even use this script :)

>when we start erroring in sentry eventually, the error will be invisible because it will occur in a docker container

I thought the container would simply refuse to start due to a missing class, forcing people to look into Docker logs, am I mistaken?

---

In any case, if you feel strongly about keeping failing in the install script, I'm fine by reserving my right to remove this in the future if it becomes a problem. I also trust you to monitor the forums for complaints around this.

untitaker: I don't really want to monitor the forum so I just removed the `exit 1`. I still think this will bite us in the ass once we want to remove redistsdb

install.sh
untitaker: done

.gitignore
BYK: Still stands as a nit.

install.sh
BYK: Ping?

untitaker: Sorry, forgot to respond. I would argue that in the case of `SENTRY_TSDB_OPTIONS` we cannot safely rewrite the config because it *might* be multiline.

.gitignore
untitaker: Yes, that was sloppy of me

install.sh
BYK: I think we can do something like this: https://github.com/getsentry/onpremise/pull/470/files#diff-3fbb47e318cd8802bd325e7da9aaabe8R198

That said I'll merge this as it is now and see if we need it later.

PR #434, opened 2020-04-16
build: Update Snuba commands

lynnagara:
Update the consumer and replacer commands. These now take a storage
instead of a dataset name as per https://github.com/getsentry/snuba/pull/875
and https://github.com/getsentry/snuba/pull/861.


lynnagara:
They aren't, the storage argument was only introduced in the PRs linked above so these new commands would fail if running a prior version of Snuba.

BYK:
This would require `getsentry/snuba:ab2e49cc1f475e59a037d882eb1cecddd23596b9` or more recent to work.

PR #462, opened 2020-05-01
fix(relay): Use Docker Hub as relay image registry

### LINKED ISSUE #445: cannot pull images us.gcr.io/sentryio/relay:latest
```
  relay:
    image: "us.gcr.io/sentryio/relay:latest"
    command: 'run --config /etc/relay'
    volumes:
      - type: bind
        source: ./relay
        target: /etc/relay
    depends_on:
      - kafka
      - redis
```

```
Error response from daemon: Get https://us.gcr.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

BYK:
We regularly prune old Google Cloud Build images and also GCB registry is not accessible to everyone all the time (firewall settings, being in China, etc.)

Fixes #445.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -177,7 +177,7 @@ services:
       - relay
   relay:
     << : *restart_policy
-    image: "us.gcr.io/sentryio/relay:latest"
+    image: "getsentry/relay:latest"
     volumes:
       - type: bind
         read_only: true


BYK:
I just merged this as the CI passed. LMK if you have any concerns @jan-auer.

jan-auer:
Thanks! Using Dockerhub is definitely more accessible. 

If we use Dockerhub as the canonical registry now, we should update [the docs](https://github.com/getsentry/relay/blob/6256c831e08ca3ad2d81d4631a60e8e880687758/docs/index.md) to point there instead of GCR.

PR #463, opened 2020-05-01
fix(nginx): Fix proxy settings for Sentry

### LINKED ISSUE #447: CSRF failures after Nginx introduction
Recently updated my install, and I've had CSRF failures.

I believe the problem is the nginx relay in https://github.com/getsentry/onpremise/blob/master/nginx/nginx.conf#L44 (introduced in #421) is not passing the original Host header (at least that fixed it locally for me).

Log line from web:

> 16:46:05 [WARNING] django.security.csrf: Forbidden (Referer checking failed - https://sentry.example.com/auth/login/org/ does not match any trusted origins.): /auth/login/org/ (status_code=403 request=<WSGIRequest: POST u'/auth/login/org/'>)

BYK:
Fixes #447.

This patch should also fix issuer regarding large uploads such as minidumps by turning off any buffering.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -19,8 +19,6 @@ http {
 
     sendfile        on;
     keepalive_timeout  65;
-    proxy_buffering off;
-    proxy_request_buffering off;
 
     upstream relay {
         server relay:3000;
@@ -35,6 +33,10 @@ http {
         # use the docker DNS server to resolve ips for relay and sentry containers
         resolver 127.0.0.11 ipv6=off;
         client_max_body_size 100M;
+
+        proxy_redirect off;
+        proxy_set_header Host       $host;
+
         location /api/store/ {
             proxy_pass http://relay;
         }
@@ -43,8 +45,6 @@ http {
         }
         location / {
            proxy_pass http://sentry;
-           proxy_redirect off;
-           proxy_set_header Host       $host;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


mattrobenolt:
Why would this fix it? We don’t do this in production.

nginx/nginx.conf
```diff
@@ -42,6 +43,8 @@ http {
         }
         location / {
            proxy_pass http://sentry;
+           proxy_redirect off;
+           proxy_set_header Host       $host;
```
mattrobenolt, line 47: Move this outside of the location block.

nginx/nginx.conf
```diff
@@ -18,8 +18,9 @@ http {
                       '"$http_user_agent" "$http_x_forwarded_for"';
 
     sendfile        on;
-
     keepalive_timeout  65;
+    proxy_buffering off;
```
mattrobenolt, line 22: These are not smart.

nginx/nginx.conf
jan-auer: See https://github.com/getsentry/sentry/blob/e8b7c215d9bb8fa789584ffc827611d060f9de23/config/reverse_proxy/nginx.conf#L49. I'll try to figure out how we missed this in the first place.

BYK: Sure, didn't know `relay` cared about this.

PR #464, opened 2020-05-01
upgrade(clickhouse): Use 19.17 as it seems to be fixing some issues

### LINKED ISSUE #433: upgrade clickhouse to fix high iowait issue
Hi, we have deployed sentry 10 with clickhouse 19.11 in a single node and find that the iowait would be very high from time to time. 
After an investigation we find out it is problem of clickhouse.
According to this https://github.com/ClickHouse/ClickHouse/issues/6372 that upgrade clickhouse may have great help.
Do you know which version of clickhouse should we use now?

BYK:
Fixes #433.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -92,7 +92,7 @@ services:
       - 'sentry-secrets:/etc/kafka/secrets'
   clickhouse:
     << : *restart_policy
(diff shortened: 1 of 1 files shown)


BYK:
image

PR #465, opened 2020-05-01
upgrade(confluent): Upgrade cp-stack to 5.5.0

BYK:
Suggested [on the forum](https://forum.sentry.io/t/connection-to-kafka-failed-when-installing/9162/10?u=byk) and the [upgrade docs](https://kafka.apache.org/25/documentation.html#upgrade) suggest upgrading from `5.1.x` without a rolling upgrade should be fine by just upgrading the code.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -63,7 +63,7 @@ services:
       - 'sentry-postgres:/var/lib/postgresql/data'
   zookeeper:
     << : *restart_policy
-    image: 'confluentinc/cp-zookeeper:5.1.2'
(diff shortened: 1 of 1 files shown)


mattrobenolt:
I wanna say whatever issues aren’t because of the version. We use 5.1.0 in single tenant obviously without this.

I also don’t really think there’d be an issue though upgrading. But it sees coincidental that changing the version magically fixes something like that.

BYK:
@mattrobenolt I mean there are a few patch releases in between so something may have gotten fixed affecting a subset of users. If there's no harm in upgrading, I'd simply upgrade.

PR #470, opened 2020-05-04
ref(relay): More robust Relay credentials setting

BYK:
This patch does two things:

1. Separate creating of Relay credentials from syncing them to Sentry config
2. Has a more flexible Relay credentials check and sync: look for the quoted public key in the config file, if it is there, assume this must be correctly set as it is very unlikely to have that random key in a different context with quotes around. The second one is to allow having other whitelisted relay keys by using an append method when adding the new key.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -175,38 +175,29 @@ fi
 
 
 if [ ! -f "$RELAY_CREDENTIALS_JSON" ]; then
-    echo ""
-    echo "Generating Relay credentials..."
+  echo ""
+  echo "Generating Relay credentials..."
 
-    # We need the ugly hack below as `relay generate credentials` tries to read the config and the credentials
-    # even with the `--stdout` and `--overwrite` flags and then errors out when the credentials file exists but
-    # not valid JSON. We hit this case as we redirect output to the same config folder, creating an empty
-    # credentials file before relay runs.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -175,38 +175,29 @@ fi
...
+
+RELAY_CREDENTIALS=$(sed -n 's/^.*"public_key"[[:space:]]*:[[:space:]]*"\([a-zA-Z0-9_-]\{1,\}\)".*$/\1/p' "$RELAY_CREDENTIALS_JSON")
+if [ -z "$RELAY_CREDENTIALS" ]; then
+  >&2 echo "FAIL: Cannot read credentials back from $RELAY_CREDENTIALS_JSON."
+  >&2 echo "      Please ensure this file is readable and contains valid credentials."
+  >&2 echo ""
+  exit 1
+fi
+
+if ! grep -q "\"$RELAY_CREDENTIALS\"" "$SENTRY_CONFIG_PY"; then
```
jan-auer, line 197: This would mean that if someone still has old keys, we add another line with `SENTRY_RELAY_WHITELIST_PK` but also leave the old one in. Is this intended?

BYK: Yes, exactly. We don't know if the "old" one is really old. Maybe they are running multiple Relays, one external etc.

jan-auer: Alright, that makes sense. It'll read a bit weird syntactically, but the behavior is definitely more desirable than before. Thanks for clarifying!

Issue #439, opened 2020-04-23
Feature request (and offer for pull request): Environment Variable Support


schrodingersket:
Hello!

I'm looking at integrating Sentry on-prem for a few of our business customers who have a hard requirement for on-site deployments (we use the Sentry SaaS for the bulk of our customer base), and I'm hoping to integrate that with our existing infrastructure; for all of our services, we do that via environment variable configuration. It looks like something similar to that was supported in 9.1 (https://github.com/getsentry/docker-sentry/blob/master/9.1/sentry.conf.py#L5-L39), but I did see that you intended to move away from environment configuration toward volume-mounted config files (https://github.com/getsentry/onpremise/pull/407 and https://github.com/getsentry/onpremise/issues/314).

Is there any chance you'd be open to a pull request that re-introduces environment-driven configuration which defaults back to the current configuration? I'd be happy to put in the work for the code/doc.

Thanks in advance!

schrodingersket:
After digging a little deeper into the Sentry source, it looks like there's already environment variable support for Docker containers in the main Sentry repo: https://github.com/getsentry/sentry/blob/master/docker/sentry.conf.py

I think a great way to add support for this while still respecting the move toward volume-mounted config files would be to create another `env-config.yml` (or some such... don't really have an opinion on the name, other than being something other than `docker-compose.override.yml`, as that would be automatically loaded) file which eliminates the `sentry` folder volume mount and configures via environment variables instead. This ensures that using environment variables is opt-in, and doesn't at all impact the current `install.sh` paradigm.

Happy to add support for this too (:

BYK:
This was a decision to simplify maintenance as any new config we add there needs to be mirrored to environment variables and we already have the `sentry.conf.py` and `config.yaml` files. I don't see the value of adding yet another mechanism with a continuous maintenance burden. Why do you need environment variables, or why does the current solution not work for you? :)

schrodingersket:
For sure! I definitely understand wanting to avoid extra maintenance. 

We use Docker Compose to run our on-site deployments as well but are highly discouraged from requiring the use of installer scripts (as such, `sed`-ing the daylights out of a conf file is discouraged), and are instead encouraged to add Docker Compose files driven by a single `.env` file. A big part of the motivation behind this this is to ensure our services and infrastructure are ready to move to something like Kubernetes or Docker Swarm down the road. It also helps ensure that our dataloaders are idempotent, facilitates cross-platform ease of use, etc. 

[...]

PR #481, opened 2020-05-08
#439: Added default support for environment-driven configuration.

schrodingersket:
This pull request addresses #439 and provides environment-driven support which respects existing file mounts and utilizes existing environment configuration from the official Sentry Docker image. This is achieved in two ways:

1) The primary configuration folder is moved from `/etc/sentry` -> `/etc/sentry/onpremise`. This allows the built-in config files to be used (which are environment-driven). 

2) Out-of-the-box configuration is provided by including `docker-compose.env.yml`, which contains defaults to run with exactly the same configuration as provided in the example config files.

### DIFF
--- a/docker-compose.env.yml
+++ b/docker-compose.env.yml
@@ -1,0 +1,122 @@
+version: '3.4'
+
+# All environment variables referenced from
+# https://github.com/getsentry/sentry/blob/master/docker/sentry.conf.py
+#
+x-restart-policy-run-once: &restart_policy_run_once
+  restart: on-failure
+x-sentry-defaults-override: &sentry_defaults_override
+  environment:
+    SENTRY_CONF: '/etc/sentry'
+    SNUBA: ${SENTRY_SNUBA_API:-http://snuba-api:1218}
+    SENTRY_POSTGRES_HOST: ${SENTRY_POSTGRES_HOST:-postgres}
+    SENTRY_POSTGRES_PORT: ${SENTRY_POSTGRES_PORT}
+    SENTRY_DB_NAME: ${SENTRY_DB_NAME:-postgres}
+    SENTRY_DB_USER: ${SENTRY_DB_USER:-postgres}
+    SENTRY_DB_PASSWORD: ${SENTRY_DB_PASSWORD}
+    SENTRY_RABBITMQ_HOST: ${SENTRY_RABBITMQ_HOST}
+    SENTRY_RABBITMQ_USERNAME: ${SENTRY_RABBITMQ_USERNAME:-guest}
+    SENTRY_RABBITMQ_PASSWORD: ${SENTRY_RABBITMQ_PASSWORD:-guest}
+    SENTRY_RABBITMQ_VHOST: ${SENTRY_RABBITMQ_VHOST:-/}
+    SENTRY_REDIS_HOST: ${SENTRY_REDIS_HOST:-redis}
+    SENTRY_REDIS_PORT: ${SENTRY_REDIS_PORT:-6379}
+    SENTRY_REDIS_PASSWORD: ${SENTRY_REDIS_PASSWORD}
+    SENTRY_REDIS_DB: ${SENTRY_REDIS_DB:-0}
+    SENTRY_MEMCACHED_HOST: ${SENTRY_MEMCACHED_HOST:-memcached}
+    SENTRY_MEMCACHED_PORT: ${SENTRY_MEMCACHED_PORT:-11211}
+
+    SENTRY_SINGLE_ORGANIZATION: ${SENTRY_SINGLE_ORGANIZATION:-True}
+    SENTRY_SECRET_KEY: ${SENTRY_SECRET_KEY:-super secret secret key}
+    SENTRY_USE_SSL: ${SENTRY_USE_SSL:-False}
+x-snuba-defaults-override: &snuba_defaults_override
+  environment:
+    SNUBA_SETTINGS: docker
+    CLICKHOUSE_HOST: ${SENTRY_CLICKHOUSE_HOST:-clickhouse}
+    DEFAULT_BROKERS: ${SENTRY_KAFKA_HOST:-kafka}:${SENTRY_KAFKA_PORT:-9092}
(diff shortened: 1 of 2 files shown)


docker-compose.override.yml
```diff
@@ -0,0 +1,5 @@
+version: '3.4'
```
BYK, line 1: I'm guessing you didn't intend to include this file in your commit? :)

docker-compose.yml
```diff
@@ -20,11 +20,13 @@ x-sentry-defaults: &sentry_defaults
     - symbolicator
     - kafka
   environment:
-    SENTRY_CONF: '/etc/sentry'
+    SENTRY_CONF: '/etc/sentry/onpremise'
     SNUBA: 'http://snuba-api:1218'
   volumes:
     - 'sentry-data:/data'
-    - './sentry:/etc/sentry'
+    - './sentry/config.yml:/etc/sentry/onpremise/config.yml'
```
BYK, line 27: Shall we switch to https://docs.docker.com/compose/compose-file/#configs as we are changing this part?

docker-compose.yml
```diff
@@ -164,8 +166,6 @@ services:
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
   nginx:
     << : *restart_policy
-    ports:
```
BYK, line 167: Revert this change please :)

docker-compose.env.yml
```diff
@@ -0,0 +1,131 @@
+version: '3.4'
+
+# All environment variables referenced from
```
BYK, line 3: I have 2 issues with this file:

1. It should not replicate _everything_ we have in the original `docker-compose.yml` file as maintaining parity between them would be a nightmare.
2. The list of environment variables getting out of sync with the original config file. Any idea how to simplify this?

schrodingersket: > 1. It should not replicate everything we have in the original docker-compose.yml file as maintaining parity between them would be a nightmare.

I think I'd mentioned in #439 only adding support for the connection information (host/port/db name) for the infra services... would that be acceptable? I think at the very least, avoiding configuration for anything that can be configured in-app is probably wise, yeah?

> 2. The list of environment variables getting out of sync with the original config file. Any idea how to simplify this?

I think in addressing point 1, that's probably mitigated since only core infrastructural config that's unlikely to change is exposed. What do you think?

docker-compose.override.yml
schrodingersket: Initially, I'd meant to; IIRC, when I'd created this branch initially, the port was exposed on the `web` service, which implied that the port exposure was a setting not meant for production (e.g., in production, that port would be closed off in favor of running behind a proxy, and the port mapping would be in developer overrides in `docker-compose.override.yml`). 

The main reason I'd added this in the first place is that an override file can't remove an existing port mapping, and wanted the public port exposure to be optional for the fairly common use-case of running behind a proxy without exposing a redundant port.

What are your thoughts on that? If you'd like to keep that mapping in the main `docker-compose.yml` file, would you be willing to make the publicly-exposed port configurable?

docker-compose.yml
schrodingersket: Sure! I assume we'd point to the config files generated after running `install.sh`, yeah?

schrodingersket:
What's the best way to add tests for this configuration? I admit that I'm not terribly familiar with Travis CI, but the test file here seems pretty straightforward. Would running the same test script as for the config file install be acceptable?

Thanks as always for your time on this!!

docker-compose.yml
schrodingersket: @BYK - looks like the `configs` feature is only supported in Docker Swarm: https://github.com/docker/compose/issues/5110

Shall we leave this as a folder mount?

docker-compose.env.yml
BYK: >I think I'd mentioned in #439 only adding support for the connection information (host/port/db name) for the infra services... would that be acceptable? I think at the very least, avoiding configuration for anything that can be configured in-app is probably wise, yeah?

Yup, that's easier. I'm wondering if we can somehow automate this. Worth thinking hard on it as that would make maintenance incredibly easier. Maybe establishing a pattern like `SENTRY_CONF_*` or something. Still though, I'm finding this harder to reason about the more implementation details I see:

 - If we need to list these in a file, then why not just list them in an `.env` file, with their values?
 - And then if we are listing them in a _file_, why can we not make that file the config file we already have?

I guess with tests + minimal connection options, this can be okay.

BYK:
>What's the best way to add tests for this configuration? I admit that I'm not terribly familiar with Travis CI, but the test file here seems pretty straightforward. Would running the same test script as for the config file install be acceptable?

I think at this point we'd need multiple jobs as defined here: https://docs.travis-ci.com/user/build-matrix/#explicitly-included-jobs-with-only-one-element-in-the-build-matrix

In the second job definition, we'd forcefully use the ENV-driven config and set some values and see if they actually work. That said, this probably means we should not rely on docker-compose to set other services up and then set them up manually. At this point, I again start to question whether this approach makes sense or not.

>Thanks as always for your time on this!!

Thank _you_ for driving this and pushing forward. I know I may sound negative/pessimistic at times but I really want to understand the use case and it's longer-term fit into the project while making this work for you and people similar to your use case. Your contributions are very valuable and much appreciated :)

BYK:
@schrodingersket it's been quite a while and I was going through some old issues and PRs that got stale. Is there any way to salvage the work you've already put here or should we close this PR?

Either way, thanks a lot for all your time and effort on this PR and the related issue!

schrodingersket:
@BYK - should be good to close this guy out. I think the complexity of the Docker Compose set up warrants simply installing this guy as a standalone deployment and integrating that way, rather than integrating it into other projects via composition of Docker Compose files. (: Thanks for all the awesome feedback though - definitely helped me put my own configuration and setup through its paces!

BYK:
@schrodingersket thanks! Happy to work on another PR that we can merge in the future :)

PR #496, opened 2020-05-18
fix(ipv6): Disable IPv6 across the board

BYK:
This should fix some rare network connectivity and DNS AAAA record issues such as https://forum.sentry.io/t/relay-errors-in-fresh-new-on-premise-install/9804?u=byk


BYK:
>I don't really agree with this and seems odd, but I don' t have any other suggestions. It seems like a heavy handed approach to me. 
>I imagine this would affect making outbound connections to ipv6 addresses.

Yeah, I agree with these concerns hence will close the PR w/o merging for now until this issue becomes more of a pain. Then I'll do some more investigation first before resorting back to this solution, if ever.

PR #499, opened 2020-05-19
fix(uwsgi): Make sure uWSGI talks proper HTTP/1.1

### LINKED ISSUE #486: Relay cannot authenticate

```
```

BYK:
This patch brings back the HTTP/1.1 related settings for uWSGI to fix #486 as apparently Relay tries to talk to Sentry Web with keep alives where uWSGI terminates the connection unexpectedly. It also ports some configs for uWSGI and nginx from single-tenant.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -1,28 +1,28 @@
-user  nginx;
-worker_processes  1;
+user nginx;
+worker_processes 1;
 
-error_log  /var/log/nginx/error.log warn;
-pid        /var/run/nginx.pid;
+error_log /var/log/nginx/error.log warn;
+pid /var/run/nginx.pid;
 
 
 events {
-    worker_connections  1024;
+	worker_connections 1024;
 }
 
 
 http {
-    default_type  application/octet-stream;
+	default_type application/octet-stream;
 
-    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
-                      '$status $body_bytes_sent "$http_referer" '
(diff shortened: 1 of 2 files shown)


nginx/nginx.conf
```diff
@@ -1,50 +1,73 @@
...
+
+	keepalive_timeout 75s;
+
+	gzip off;
+	server_tokens off;
+
+	server_names_hash_bucket_size 64;
+	types_hash_max_size 2048;
+	types_hash_bucket_size 64;
+	client_max_body_size 5m;
```
macnibblet, line 33: We went from 100m to 5m in this PR and caused a few issue on our send when uploading android/ios build maps. Is there anyway we can customise this without having to edit the nginx.conf every time?

Maybe add a include statement which includes custom settings that are not part of git?

mattrobenolt: I agree this is restrictive. We need different sizes per endpoint. I have this sorta addressed in a better way in an internal project that I hope to make public to replace this. My goals are to better keep these in sync with how we handle limits in production with values that are known to be safe, so ideally you wouldn't need to customize, but the limits should just match how we run SaaS.

But yes, @BYK we definitely need much larger than 5m max for upload endpoints.

BYK: @macnibblet - oops, sorry yeah this was an oversight. Updated (fixed) these in https://github.com/getsentry/onpremise/commit/85f267bec78a43f15f0fdc7da0e4bfe646eb8e16

PR #510, opened 2020-05-23
meta(versioning): Reword versioning sect. now that we have version tags

BYK:


README.md
```diff
@@ -29,21 +29,15 @@ If you have any issues or questions, our [Community Forum](https://forum.sentry.
...
 
 ```shell
 SENTRY_IMAGE=getsentry/sentry:83b1380 ./install.sh
 ```
 
-If you want to use different or specific images for other services, you may create a `docker-compose.overrides.yaml` file in the repo and override the `image` field for the corresponding services.
-
-We strongly recommend keeping the `latest` tags for all, if you are using this repository directly. We also recommend using specific commit tags if you are consuming any of our Docker images in an environment that needs consistent deploys such as [a Helm chart](https://github.com/helm/charts/tree/master/stable/sentry).
+Note that this may not work for all commit SHAs as this repository evolves with Sentry and its sattelite projects. It is highly recommended checking out a version of this repository that is close to the timestamp of the Sentry commit you are installing.
```
joshuarli, line 40: typo: `satellite`

joshuarli, line 40: `highly recommended checking out` -> `highly recommended to check out`

PR #512, opened 2020-05-23
feat(env): Add SENTRY_BIND var

### LINKED ISSUE #279: Allow configuring web port as env var

BYK:
Closes #279, supercedes #306. This is much simpler and safer now that we have `nginx` in front of everything.

Thanks a lot @larsnystrom!


mohammedi-haroune:
What would happen if someone changes the port using the added environment variable while the port number `9000` is hardcoded is other files:
* https://github.com/getsentry/onpremise/blob/master/nginx/nginx.conf#L57
* https://github.com/getsentry/onpremise/blob/master/relay/config.yml#L3
* https://github.com/getsentry/onpremise/blob/master/sentry/config.example.yml#L68
* https://github.com/getsentry/onpremise/blob/master/sentry/sentry.conf.example.py#L156

Replacing them manually is not convenient, especially for Nginx and Relay because they are source controlled and it'd make `git pull` tricky.

BYK:
@mohammedi-haroune the hard-coded port number is only used inside the Docker bridge network so it doesn't need to be changed anywhere.

docker-compose.yml
```diff
@@ -168,7 +168,7 @@ services:
   nginx:
     << : *restart_policy
     ports:
-      - '9000:80/tcp'
+      - '$SENTRY_ONPREMISE_PORT:80/tcp'
```
caugner, line 171: As a nice side effect, this would also allow restricting the port forwarding to `127.0.0.1`:

```
SENTRY_ONPREMISE_PORT=127.0.0.1:9000
```

mohammedi-haroune:
@BYK it means sentry will still accessible in docker network via `web:9000` even if the Nginx port is changed in `docker-compose.yml`?

docker-compose.yml
BYK: TIL you can use IP/Host here, nice! (ref to the docs where this is mentioned: https://docs.docker.com/compose/compose-file/#ports)

BYK:
@mohammedi-haroune 
>@BYK it means sentry will still accessible in docker network via web:9000 even if the Nginx port is changed in docker-compose.yml?

Yes.

mattrobenolt (approved):

PR #514, opened 2020-05-24
fix(install): Skip pull when SENTRY_IMAGE is set

BYK:
This also uses SENTRY_VERSION instead of hard-coding `:latest` as the tag when pulling (follow up to #509).


marbon87:
Hi @BYK,
is use SENTRY_IMAGE to set an image-name, that has our docker-proxy-registry as a prefix, like
```
SENTRY_IMAGE=docker-proxy.example.com/getsentry/sentry
```
Now the latest image isn't pulled anymore, because of this change.
What's the problem with always calling docker pull? If the image is already up to date, the call should take only a few seconds.

BYK:
@marbon87 - it was throwing an error when using a local image. I pushed 
83160e8 to master to alleviate the problem you mentioned.

PR #534, opened 2020-06-08
frontend: hello

mattrobenolt:
Replace home rolled nginx with the all encompassing frontend app

See: https://github.com/getsentry/single-tenant/pull/148


mattrobenolt:
I've tested this to the best of my knowledge and things appear fine.

docker-compose.yml
```diff
@@ -170,16 +170,17 @@ services:
...
       - '9000:80/tcp'
-    image: "nginx:1.16"
-    volumes:
-      - type: bind
-        read_only: true
-        source: ./nginx
-        target: /etc/nginx
+    image: 'us.gcr.io/sentryio/frontend:latest'
+    environment:
+      UPSTREAM_WEB: web:9000
```
mattrobenolt, line 179: Technically this is a change in behavior. I'm not sure if you cared before, but if `web` got a new IP behind it, it'd be broken forever before.

We now force DNS resolution at runtime with `FORCE_DNS_RESOLVE`. If you want to remove that, feel free and just set it to `false`, and it'll mimic current behavior. At that point, could also drop `DNS_RESOLVER`.

docker-compose.yml
```diff
@@ -170,16 +170,17 @@ services:
...
     << : *restart_policy
     ports:
       - '9000:80/tcp'
-    image: "nginx:1.16"
-    volumes:
-      - type: bind
-        read_only: true
-        source: ./nginx
-        target: /etc/nginx
+    image: 'us.gcr.io/sentryio/frontend:latest'
```
mattrobenolt, line 177: I am assuming you'd rather not pin a strict version here, but if you would, those will come soon and we can wait.

docker-compose.yml
```diff
@@ -170,16 +170,17 @@ services:
...
+    environment:
```
mattrobenolt, line 178: There's also an additional variable which might be nice to add:

```
LOG_FORMAT: json
```

And this will dump out structured access logs. It's not doing this today, so didn't set it.

docker-compose.yml
BYK: On-premise users based in China cannot reach this registry so we need to put these images on Docker Hub if possible (somehow they are reachable inside).

docker-compose.yml
BYK: I think this is a good bugfix as we would never want to end up in a permanently broken state?

docker-compose.yml
BYK: Also is this repo public yet?

mattrobenolt: Sure, we can push it to docker hub at some point, and no, it's not public yet.

PR #535, opened 2020-06-09
ref(setup): Rename config to SENTRY_TRUSTED_RELAY_PKS

jan-auer:
Renames the setting for trusted Relays to more sensitive language.

Requires https://github.com/getsentry/sentry/pull/19240


jan-auer:
@BYK please let me know if you'd like to have both as a transition phase. Since there was no release since introducing Relay, however, I'd say we can just change this.

BYK:
This is now obsolete with #572 so closing.

PR #543, opened 2020-06-16
Fixing Relay authorization issue

kotso:
credentials are correct `SENTRY_RELAY_WHITELIST_PK` contains correct key but auth still fails. 

was able to fix issue only after forced relay to communicate via Nginx, not directly.  (did not had more time to troubleshoot network communication and do some tcpdump capture analyses)

### DIFF
--- a/relay/config.yml
+++ b/relay/config.yml
@@ -1,6 +1,6 @@
 ---
 relay:
-  upstream: "http://web:9000/"
+  upstream: "http://nginx/"
   host: 0.0.0.0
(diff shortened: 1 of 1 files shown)


BYK:
Thanks a lot for this! We're looking into ways for avoiding going through Nginx and making this work with uWSGI somehow but I'll keep the PR open in case we cannot achieve this.

BYK:
Closing in favor of #576. Thanks again @kotso!

PR #552, opened 2020-06-24
fix(zookeeper): cp: cannot create regular file

### LINKED ISSUE #547: Sentry migrate from 10 to 20.6.0 - Zookeeper cp: cannot create regular file
On migrating to v20.6.0 from Sentry 10, I encountered an error during the install process which had the following description:

```
cp: cannot create regular file ‘/var/lib/zookeeper/data/version-2/snapshot.0’: No such file or directory
```

After some debugging and investigation I noticed that the issue lies in the line of code at:
https://github.com/getsentry/onpremise/blob/master/install.sh#L178

What I've found is that the copy target folder doesn't exist within the container, more specifically, this copy target location:
`/var/lib/zookeeper/data/version-2`

I've rectified the issue by updating the erroring line of code to include creating the copy target location:

```
$dcr -v $(pwd)/zookeeper:/temp zookeeper bash -c 'mkdir -pv /var/lib/zookeeper/data/version-2 && cp /temp/snapshot.0 /var/lib/zookeeper/data/version-2/snapshot.0'
```

I haven't been able to find any articles relating to this issue so I thought I'd post it here with the solution, however, I can create a PR with that change if it is to your satisfaction. Would this be an ideal fix for the issue?

Docker version: `Docker version 17.05.0-ce, build 89658be`
Compose version: `docker-compose version 1.24.1, build 4667896b`

strange-developer:
Hi,

I've been through quite a few different ways of implementing this fix and settled on creating a variable to store the output of checking whether the zookeeper copy target folder exists and copying the snapshot file based on the copy target folder existing. I've ran quite a few manual tests for each option as well. Currently the PR sits on Option 3 from the below options.

**Option 1**
Judging from the [Jira issue](https://issues.apache.org/jira/browse/ZOOKEEPER-3056), it seems like the work around for zookeeper upgrades could be omitted entirely since the issue relates to upgrades from v3.4.10 to v3.5.4. I've tested removing the zookeeper workaround entirely and that install ran smoothly on a clean install of Sentry (no existing data) as well as an install of Sentry that currently has very minimal amount of log entries (roughly 100 log entries). Could we possibly remove the workaround entirely? 

**Option 2**
The second option was to simply add a check to the currently [existing line](https://github.com/getsentry/onpremise/blob/master/install.sh#L178) of whether the copy target folder exists and perform the snapshot file copy only if the copy target folder exists. This is the least amount of code and possibly the simpler fix while also setting the `ZOOKEEPER_SNAPSHOT_TRUST_EMPTY` env var to `true`, however, some unnecessary calculations will be done to determine the `ZOOKEEPER_LOG_FILE_COUNT` and `ZOOKEEPER_SNAPSHOT_FILE_COUNT`.

**Option 3**
I've created a variable to store whether the copy target folder exists and proceed with the zookeeper upgrade workaround only if the copy target folder exists. This means that if the copy target folder does not exist, the env var `ZOOKEEPER_SNAPSHOT_TRUST_EMPTY` will not be set.

Fixes #547.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -170,15 +170,16 @@ $dc build --force-rm --parallel
 echo ""
 echo "Docker images built."
 
-
-ZOOKEEPER_LOG_FILE_COUNT=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/log/version-2/* | wc -l | tr -d '[:space:]'')
-ZOOKEEPER_SNAPSHOT_FILE_COUNT=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/data/version-2/* | wc -l | tr -d '[:space:]'')
-# This is a workaround for a ZK upgrade bug: https://issues.apache.org/jira/browse/ZOOKEEPER-3056
-if [ "$ZOOKEEPER_LOG_FILE_COUNT" -gt "0" ] && [ "$ZOOKEEPER_SNAPSHOT_FILE_COUNT" -eq "0" ]; then
-  $dcr -v $(pwd)/zookeeper:/temp zookeeper bash -c 'cp /temp/snapshot.0 /var/lib/zookeeper/data/version-2/snapshot.0'
-  $dc run -d -e ZOOKEEPER_SNAPSHOT_TRUST_EMPTY=true zookeeper
+ZOOKEEPER_SNAPSHOT_EXISTS=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/data/version-2 | wc -l | tr -d '[:space:]'')
+if [ "$ZOOKEEPER_SNAPSHOT_EXISTS" -eq "1" ]; then  
+  ZOOKEEPER_LOG_FILE_COUNT=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/log/version-2/* | wc -l | tr -d '[:space:]'')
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -170,16 +170,17 @@ $dc build --force-rm --parallel
...
 echo "Docker images built."
 
-
-ZOOKEEPER_LOG_FILE_COUNT=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/log/version-2/* | wc -l | tr -d '[:space:]'')
-ZOOKEEPER_SNAPSHOT_FILE_COUNT=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/data/version-2/* | wc -l | tr -d '[:space:]'')
-# This is a workaround for a ZK upgrade bug: https://issues.apache.org/jira/browse/ZOOKEEPER-3056
-if [ "$ZOOKEEPER_LOG_FILE_COUNT" -gt "0" ] && [ "$ZOOKEEPER_SNAPSHOT_FILE_COUNT" -eq "0" ]; then
-  $dcr -v $(pwd)/zookeeper:/temp zookeeper bash -c 'cp /temp/snapshot.0 /var/lib/zookeeper/data/version-2/snapshot.0'
-  $dc run -d -e ZOOKEEPER_SNAPSHOT_TRUST_EMPTY=true zookeeper
+ZOOKEEPER_SNAPSHOT_FOLDER_EXISTS=$($dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/data/version-2 | wc -l | tr -d '[:space:]'')
```
BYK, line 173: Maybe we can make a function that returns `$dcr zookeeper bash -c 'ls 2>/dev/null -Ubad1 -- /var/lib/zookeeper/$SOME_PATH | wc -l | tr -d '[:space:]''` to reduce the repetition here.

Issue #550, opened 2020-06-23
Nginx with ssl gives CRSF error - improvement


paladium:
Hi everyone, I tried installing sentry on premise without SSL and didn`t have any issues. After I have added the SSL configuration to Nginx:
```nginx
server{
listen 80 ssl;
ssl_certificate ./cert.crt;
ssl_certificate_key ./cert.key;
...
}
```
And in sentry.conf.py uncommented these lines:
```py
SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
SESSION_COOKIE_SECURE = True
CSRF_COOKIE_SECURE = True
```
I was getting: ***CSRF Failed: CSRF token missing or incorrect*** in UI. After digging in the forum, I found that default Nginx configuration needs to be altered slightly. Namely, the line:
```nginx
proxy_set_header Host $host;
```
Needs to be changed to:
```nginx
proxy_set_header Host $http_host;
```
And everything seems to work. I suggest adding it to documentation of SSL for sentry setup or changing the default Nginx configuration.
Thank you.

BYK:
Thanks a lot for taking the effort for sharing this @paladium. Took note for the future "How to run w/ SSL docs".

BYK:
This should be addressed via #534.

BYK:
Closing as we now recommend terminating SSL/TLS at load balancer level, instead of modifying the built-in nginx config: https://develop.sentry.dev/self-hosted/#productionalizing

PR #558, opened 2020-07-01
SSL doc and stubs support

### LINKED ISSUE #553: Cant enable ssl in sentry.config.conf?
Hey i am running the latest nightly Sentry 20.7.0.dev 02dfdc2d

first off bravo, this was a breeze to get going to serve a basic onpremise, legit took longer to spin up a box in AWS :D.

Sorry for posting in two places, but i mostly solved my first issue ( https://forum.sentry.io/t/sentry-nightly-expose-port-80/10260/3 ) and was wondering if i have stumbled across a possible bug in the nightly? 

basically when i enable the ssl settings in sentry.config.conf, the server just starts throwing 502s (see attached link for error detail from nginx docker container) o.0. 

So i could just ignore it but, im also getting told that the page is "insecure ssl". not a big deal but bugs the perfectionist in me :D 

any idea what i have done wrong ? 

Thanks in advance!

phyzical:
fixes #553
* added commented out placeholders to allow for an easier ssl "on" to docker-compose and nginx.conf and ssl folder stubs
* updated readme to help explain the commented stubs


phyzical:
@BYK 👋 

solves the confusion behind setting up a https/ssl install by adding a few simple stubs for enabling and adding a https enforcement. 

let me know if anything isnt up to scratch and i shall update it further

README.md
```diff
@@ -49,6 +49,48 @@ If you'd like to protect your Sentry install with SSL/TLS, there are
 fantastic SSL/TLS proxies like [HAProxy](http://www.haproxy.org/)
```
BYK, line 49: Heh, we should definitely remove this line as it predates us adding Nginx to the mix.

README.md
```diff
@@ -49,6 +49,48 @@ If you'd like to protect your Sentry install with SSL/TLS, there are
...
 
+OR do the following things to enable auto redirect to https behind a reverse proxy
+
+Uncomment the following in `docker-compose.yml` nginx directive
+```
+ #expose:
+ #  - '443'
+ ports:
+ #  - '443:443'
+   - '$SENTRY_ONPREMISE_PORT:80/tcp'
```
BYK, line 60: For this, we need #512 merged first, right?

README.md
```diff
@@ -49,6 +49,48 @@ If you'd like to protect your Sentry install with SSL/TLS, there are
...
+```
+
+And uncomment the following in your sentry.config.py
```
BYK, line 63: I think instead of repeating the same config here in the README, we can mark the start and end of this section clearly with `SSL CONFIG START`, `SSL_CONFIG_END` and just say "Uncomment all the config inside the `SSL CONFIG` section.

Even better, we may wrap that section in an
```py
if env('SENTRY_USE_SSL'):
   ...
```

block, and just tell people to enable this in their `.env` file.

README.md
```diff
@@ -49,6 +49,48 @@ If you'd like to protect your Sentry install with SSL/TLS, there are
...
+```
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+SESSION_COOKIE_SECURE = True
+CSRF_COOKIE_SECURE = True
+SOCIAL_AUTH_REDIRECT_IS_HTTPS = True
+SENTRY_USE_SSL = True
+```
+
+And uncomment the following `#`s in your nginx config, and simply replace the contents of the `cert.crt` and `cert.key` in the ssl folder found inside the nginx folder
```
BYK, line 72: Same concern again, let's not repeat config in the readme, just use markers instead.

docker-compose.yml
```diff
@@ -172,7 +172,10 @@ services:
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
   nginx:
     << : *restart_policy
+    #expose:
+    #  - '443'
```
BYK, line 176: This instruction is obsolete so let's just remove it

```suggestion
```

nginx/nginx.conf
```diff
@@ -41,7 +41,9 @@ http {
 	# Remove the Connection header if the client sends it,
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';
+	## comment me and uncomment below me to support ssl
 	proxy_set_header Host $host;
+	#proxy_set_header Host $http_host;
```
BYK, line 46: Why this?

nginx/ssl/cert.crt
```diff
@@ -0,0 +1 @@
+TOTALGARBAGEPLACEHOLDERSHIZ
```
BYK, line 1: I wouldn't check these files in and instead, add them to `.gitignore`. Nginx would fail when it cannot read these files anyway?

README.md
phyzical: correct this is currently branched off of #512 and i've targeted back into 512 at this time for merging

docker-compose.yml
phyzical: it is? opps i will test without and confirm it as working.

nginx/nginx.conf
phyzical: this was based on #550 but i will try rolling it back incase it was another change i didn't need

nginx/ssl/cert.crt
phyzical: this was more of a bootstrapping for the average user for guidance on placing the ssl files to line up with the nginx conf changes, it would fail yes but only if they enable it, and if they enable they should atleast understand that they need to replace the ssl files.

i can remove these files if you prefer, but then i think git will not checkout the ssl folder as it will be empty

README.md
phyzical: i will confirm this as working and change in default sentry.conf, and update readme

README.md
phyzical: done

README.md
phyzical: confirmed and updated

docker-compose.yml
phyzical: confirmed as not needed

phyzical:
updated as per your changes, just await your feedback @BYK  on a couple of comments

README.md
phyzical: resolved

nginx/nginx.conf
phyzical: i infact did not need it, but what is #550 all about then? when you just want ssl via HTTP ?

phyzical:
ah actually.... ive spoken too soon events are running into 502s, i will let you know when i have resolved this

phyzical:
lol... so i reboot the docker-compose and it starts working how weird...

nginx/ssl/cert.crt
phyzical: (or am i missing something and there is actually selfsigned ssl stuff as part of this image?)

nginx/nginx.conf
BYK: @phyzical I think they were using an older version of our Nginx config, which was lacking the Host setting.

nginx/ssl/cert.crt
BYK: >this was more of a bootstrapping for the average user for guidance on placing the ssl files to line up with the nginx conf changes, it would fail yes but only if they enable it, and if they enable they should atleast understand that they need to replace the ssl files.

I think we can find better ways to signal this. They still need to know where to look for these files and then actually open them (people rarely open cert files in text editors I think) to see your message :)

>i can remove these files if you prefer, but then i think git will not checkout the ssl folder as it will be empty

If you think we should have the `ssl` folder, we can have an empty `.gitignore` file in it to keep it in Git.

README.md
```diff
@@ -45,9 +45,16 @@ Sentry comes with a cleanup cron job that prunes events older than `90 days` by
...
-fantastic SSL/TLS proxies like [HAProxy](http://www.haproxy.org/)
-and [Nginx](http://nginx.org/). You'll likely want to add this service to your `docker-compose.yml` file.
+If you'd like to protect your Sentry install with SSL/TLS and enforce https simply apply the changes found below
+
+* Generate a ssl crt and key for example you can use letsencrypt (see [Letsencrypt Guide](https://www.linode.com/docs/security/ssl/install-lets-encrypt-to-create-ssl-certificates/))
+Once generated simply replace the contents of the `cert.crt` and `cert.key` in the ssl folder found inside the nginx folder
+
+* Uncomment the lines in `docker-compose.yml` from NGINX_SSL_START to NGINX_SSL_END
+* Set `SENTRY_USE_SSL=True` in the `.env` to apply the sentry.config.py ssl settings
+* Uncomment the lines in `nginx\nginx.conf` from NGINX_SSL_START to NGINX_SSL_END, and comment out the lines from NGINX_NON_SSL_START to NGINX_NON_SSL_END
```
BYK, line 55:
```suggestion
* Uncomment the lines in `nginx/nginx.conf` from NGINX_SSL_START to NGINX_SSL_END, and comment out the lines from NGINX_NON_SSL_START to NGINX_NON_SSL_END
```

README.md
```diff
@@ -45,9 +45,16 @@ Sentry comes with a cleanup cron job that prunes events older than `90 days` by
...
+
+NOTE: if you noticed some things not using the appropriate URL i.e favicons in <HEAD> then your url-prefix is probably wrong, please double check this at sentry.DOMAIN.com/manage/settings/
```
BYK, line 57:
```suggestion
NOTE: If you notice things like not using the correct URLs in `<HEAD>` (such as favicons etc.), then your `system.url-prefix` setting is probably wrong. Please double check this at `https://your-sentry-domain.com/manage/settings/` or set it in your `config.yml` file.
```

phyzical: updated

README.md
phyzical: updated

nginx/ssl/cert.crt
phyzical: noted makes sense :) 

and resolved

uhlhosting:
Why isnt this approved in so many weeks now?

BYK:
@uhlhosting because we are also working on #534 which is conflicting with certain things here.

uhlhosting:
hope this will be solved. if need help to test would be more than happy to.

skelgaard:
As there is a $SENTRY_ONPREMISE_PORT shouldn't there be a $SENTRY_ONPREMISE_SSL_PORT also ? and able to set these in enviroment variables would be a good addition to it

PR #564, opened 2020-07-03
fix(nginx): expose port 80

zer0tonin:
There's nothing on nginx port 9000.


BYK:
Closing as this is not something we can merge, ever.

PR #566, opened 2020-07-04
Add volumed-out config directory for symbolicator.

mikepurvis:
I would like to be able to customize the configuration for my Sentry 10 symbolicator instance, which this change allows me to easily do.

See related: https://github.com/getsentry/symbolicator/issues/245


BYK:
Why not do this with a `docker-compose.override.yml` file instead of adding an empty config?

mikepurvis:
@BYK I don't have a strong opinion on that. I was just following the pattern of `nginx/nginx.conf`, `relay/config.yml`, `sentry/config.yml`.

install.sh
```diff
@@ -95,6 +96,7 @@ echo ""
 ensure_file_from_example $SENTRY_CONFIG_PY
 ensure_file_from_example $SENTRY_CONFIG_YML
 ensure_file_from_example $SENTRY_EXTRA_REQUIREMENTS
+ensure_file_from_example $RELAY_CONFIG_YML
```
mikepurvis: `s/RELAY/SYMBOLICATOR/g`?

BYK: Yup, fixed that already ;)

symbolicator/config.example.yml
```diff
@@ -0,0 +1,9 @@
+# See: https://getsentry.github.io/symbolicator/#configuration
+cache_dir: "/data"
+bind: "0.0.0.0:3021"
+logging:
+  level: "warn"
+  enable_backtraces: true
```
jan-auer, line 6: Setting the default level to `"warn"` makes sense, for performance one could also choose to disable backtraces. If anything, this is the one I'd propose to remove and leave to symbolicator.

BYK:
@mikepurvis thank you very much for taking the time to share your solution with us and the broader community and pushing us for a better developer experience. Very much appreciated! 🎉

mikepurvis:
Thanks for the quick merge!

PR #571, opened 2020-07-09
fix(slack): Point to newer docs, better defaults

phyzical:
* added new default going forward uncommented
* added link to future guide
* added support for new signing-secret
* slight rejig to config

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -94,13 +94,12 @@ transaction-events.force-disable-internal-project: true
 # Slack Integration #
 #####################
 
-# Refer to https://forum.sentry.io/t/how-to-configure-slack-in-your-on-prem-sentry/3463 for setup instructions.
-# Or https://develop.sentry.dev/integrations/slack/
+# Refer to https://develop.sentry.dev/integrations/slack/ for setup instructions.
 
 # slack.client-id: <'client id'>
 # slack.client-secret: <client secret>
 # slack.signing-secret: <signing secret>
-## if you made your slack bot before july 2020 set legacy-app to True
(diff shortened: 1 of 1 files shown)


sentry/config.example.yml
```diff
@@ -95,9 +95,12 @@ transaction-events.force-disable-internal-project: true
 #####################
 
 # Refer to https://forum.sentry.io/t/how-to-configure-slack-in-your-on-prem-sentry/3463 for setup instructions.
+# Or https://develop.sentry.dev/integrations/slack/
```
BYK, line 98: Let's make this the default link :)

sentry/config.example.yml
```diff
@@ -95,9 +95,12 @@ transaction-events.force-disable-internal-project: true
...
 # slack.client-id: <'client id'>
 # slack.client-secret: <client secret>
+# slack.signing-secret: <signing secret>
+## if you made your slack bot before july 2020 set legacy-app to True
+## and use verfication-token instead of signing-secret
 # slack.verification-token: <verification token>
-## only uncomment legacy-app if you made your slack bot after july 2020
-# slack.legacy-app: False
\ No newline at end of file
+slack.legacy-app: False
```
BYK, line 106: I'd move this above the verification token or reword the comment so the ordering matches (first token, then legacy-app)

sentry/config.example.yml
phyzical: done :)

sentry/config.example.yml
phyzical: how about that?

BYK: Nice :)

PR #572, opened 2020-07-09
ref(relay): Remove PK and rely on INTERNAL_IPS

BYK:
This patch adds `INTERNAL_IPS` definition to `sentry.conf.py` by sniffing the network from eth0 and relies on this for trusted Relays instead of the ALLOWLISTED PKs. This removes the necessity of syncing Relay PKs to `sentry.conf.py`.

This PR needs getsentry/sentry#19798 to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,4 +79,5 @@ sentry/config.yml
 sentry/*.bak
 sentry/requirements.txt
 relay/credentials.json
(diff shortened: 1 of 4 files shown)


BYK:
>in that case we might as well hardcode credentials.json as it serves no purpose

Good idea. Will consider this.

PR #574, opened 2020-07-10
Make cron entrypoint.sh executable

NullIsNot0:
Shell scripts lost their executable permissions while downloading project from GitHub, modifying it to my needs and later copying it to server. I had to manually set executable parameter for install script, but forgot about other scripts. When images were built, cron containers could not start. There were errors:
```
ERROR: for sentry_onpremise_symbolicator-cleanup_1  Cannot start service symbolicator-cleanup: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/entrypoint.sh\": permission denied": unknown

ERROR: for symbolicator-cleanup  Cannot start service symbolicator-cleanup: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/entrypoint.sh\": permission denied": unknown

ERROR: for snuba-cleanup  Cannot start service snuba-cleanup: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/entrypoint.sh\": permission denied": unknown

ERROR: for sentry-cleanup  Cannot start service sentry-cleanup: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/entrypoint.sh\": permission denied": unknown
ERROR: Encountered errors while bringing up the project.
```
I think it's a good idea to explicitly set cron image's executable permissions in Dockerfile, so image builds correclty in all environments.

### DIFF
--- a/cron/Dockerfile
+++ b/cron/Dockerfile
@@ -3,4 +3,5 @@ FROM ${BASE_IMAGE}
 RUN apt-get update && apt-get install -y --no-install-recommends cron && \
     rm -r /var/lib/apt/lists/*
 COPY entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]


BYK:
Hi @NullIsNot0, thanks a lot for taking the time to submit this PR. Since this step is redundant, creates yet another layer in the built Docker image, and happens only when you manually copy the repo somewhere without the `-a` flag, I don't think it makes sense to make this the default.

Thanks again for your time and effort.

PR #576, opened 2020-07-10
fix(relay): Fix relay cannot authenticate w/ Sentry

### LINKED ISSUE #486: Relay cannot authenticate

BYK:
Fix #486.

I finally figured out what the issue was: a missing `post-buffering` option to let Relay finish it's POST request for auth. This PR supersedes #543.

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -173,7 +173,7 @@ SENTRY_WEB_OPTIONS = {
     'thunder-lock': True,
     'log-x-forwarded-for': False,
     'buffer-size': 32768,
-    # Realy cannot authenticate without the following
+    # Relay cannot authenticate without the following
     'post-buffering': 32768,
     'limit-post': 209715200,
     'disable-logging': True,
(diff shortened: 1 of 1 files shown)


sentry/sentry.conf.example.py
```diff
@@ -173,6 +173,8 @@
     'thunder-lock': True,
     'log-x-forwarded-for': False,
     'buffer-size': 32768,
+    # Realy cannot authenticate without the following
```
mattrobenolt, line 176: Relay

BYK: It's really Relay :P

jan-auer:
@BYK Could you explain why this is needed?

BYK:
>@BYK Could you explain why this is needed?

I thought there was an issue draining the POST body as this was the common cause for so many people. Turns out my initial suspicion was correct: uWSGI doesn't do proper HTTP/1.1 unless you use the `http` option (not the `http-socket` we are using) or the newly added `http11-socket`, even if you set `http-keepalive` to `True`.

Moreover (even worse?), `http-keepalive` is not a `boolean` value as it seems, but is actually the keep-alive timeout value in seconds (so we were setting a 1 second timeout, which is too short).

Now I fixed all these in [`068bc86` (#578)](https://github.com/getsentry/onpremise/pull/578/commits/068bc86b8094aa9b360ab4b464a1ed2266b412d1).

/cc @mattrobenolt

PR #578, opened 2020-07-12
fix(ingest): Fix Relay auth issues and add e2e event ingestion test

### LINKED ISSUE #486: Relay cannot authenticate

BYK:
This is a long-needed test that tests the whole pipeline from Nginx, Relay, to Kafka, and Snuba. The final missing piece is testing the Symbolicator integration.

This PR is also a follow up to #576 as it didn't solve the Relay issues fully (the earlier fix was a coincidence or is not as reliable as it seemed).

Fixes #486 (finally?).

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -190,7 +190,8 @@ SENTRY_WEB_OPTIONS = {
     # This is need to prevent https://git.io/fj7Lw
     "uwsgi-socket": None,
     "so-keepalive": True,
-    "http-keepalive": 5,
+    # Keep this between 15s-75s as that's what Relay supports
+    "http-keepalive": 15,
     "http-chunked-input": True,
     # the number of web workers
     "workers": 3,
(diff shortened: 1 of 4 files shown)


sentry/sentry.conf.example.py
```diff
@@ -185,32 +185,32 @@ def get_internal_network():
...
 SENTRY_WEB_PORT = 9000
 SENTRY_WEB_OPTIONS = {
-    "http-keepalive": True,
+    "http": "%s:%s" % (SENTRY_WEB_HOST, SENTRY_WEB_PORT),
+    "protocol": "uwsgi",
+    # This is need to prevent https://git.io/fj7Lw
+    "uwsgi-socket": None,
     "so-keepalive": True,
-    "http-auto-chunked": True,
+    "http-keepalive": 5,
```
jan-auer, line 193: Relay keeps connections for 15 (min) to 75 (max) seconds. Might want to align this?

BYK: Done. I read Apache 2.2+ reduced this to 5, hence the 5 second limit here but I'll align it with Relay.

PR #585, opened 2020-07-13
fix(test): Don't use sentry-cli for test event

BYK:
Running Docker on the host network is not supported on GCB, which is breaking our builds over at getsentry/sentry. This patch removes the use of sentry-cli and does a manual `curl` request to create the test event.


untitaker:
lgtm

PR #594, opened 2020-07-15
install(clickhouse): Ensure we have the newest transactions table

### LINKED ISSUE #587: Clickhouse: We should ensure the newest version of transactions table
We have rebuilt the transactions table on Clickhouse on April 2020 for it to be more performant on larger data sets. Since we lack a formal migrations framework on Snuba for now, we need to update this table manually in the install script. We can only do this if the table is empty. Here are the possible cases:

1. The table is created After April: No action needed
2. The table was created before April and it is empty: Drop the table
3. The table was created before April and for some reason there is some data in it:
  a. All good as long as the instance does not start running queries with more than 10000 values in their tags, which is some non trivial volume.
  b. The instance has a massive amount of data (10k+ values in tags) and this happens before we build data migrations: Turn off the `discover2.tags_facet_enable_sampling` feature flag, introduced in getsentry/sentry#198882

BYK:
See getsentry/sentry#19882 and getsentry/sentry/#19883.

Fixes #587.

This PR replaces #589 as somehow it got broken.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -21,10 +21,10 @@ SENTRY_EXTRA_REQUIREMENTS='sentry/requirements.txt'
 
 # Courtesy of https://stackoverflow.com/a/2183063/90297
 trap_with_arg() {
-    func="$1" ; shift
-    for sig ; do
-        trap "$func $sig" "$sig"
-    done
+  func="$1" ; shift
+  for sig ; do
+      trap "$func $sig "'$LINENO' "$sig"
+  done
 }
 
 DID_CLEAN_UP=0
@@ -36,7 +36,7 @@ cleanup () {
   DID_CLEAN_UP=1
 
   if [ "$1" != "EXIT" ]; then
-    echo "An error occurred, caught SIG$1";
+    echo "An error occurred, caught SIG$1 on line $2";
     echo "Cleaning up..."
   fi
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -198,6 +198,33 @@ if [ "$ZOOKEEPER_SNAPSHOT_FOLDER_EXISTS" -eq "1" ]; then
...
+until clickhouse_query 'SELECT 1' > /dev/null; do
+  ((CLICKHOUSE_CLIENT_MAX_RETRY--))
+  [[ CLICKHOUSE_CLIENT_MAX_RETRY -eq 0 ]] && echo "Clickhouse server failed to come up in 5 tries." && exit 1;
+   echo "Trying again. Remaining tries #$CLICKHOUSE_CLIENT_MAX_RETRY"
+  sleep 0.5;
+  SNUBA_TRANSACTIONS_NEED_UPGRADE=$(clickhouse_query "SHOW CREATE TABLE transactions_local" | grep -v 'SAMPLE BY cityHash64(span_id)')
+done
+set -e
+
+if [ ! -z "$SNUBA_TRANSACTIONS_NEED_UPGRADE" ]; then
```
fpacifici, line 216: not very familiar with bash scripting, but is this the simplest for for this condition? it took me a while to figure out that if you did not to update the variable would be empty, that `-z` returns true if it is empty and that we negate it.

install.sh
```diff
@@ -198,6 +198,33 @@ if [ "$ZOOKEEPER_SNAPSHOT_FOLDER_EXISTS" -eq "1" ]; then
...
+set +e
+CLICKHOUSE_CLIENT_MAX_RETRY=5
+# Wait until clickhouse server is up
+until clickhouse_query 'SELECT 1' > /dev/null; do
+  ((CLICKHOUSE_CLIENT_MAX_RETRY--))
+  [[ CLICKHOUSE_CLIENT_MAX_RETRY -eq 0 ]] && echo "Clickhouse server failed to come up in 5 tries." && exit 1;
+   echo "Trying again. Remaining tries #$CLICKHOUSE_CLIENT_MAX_RETRY"
+  sleep 0.5;
+  SNUBA_TRANSACTIONS_NEED_UPGRADE=$(clickhouse_query "SHOW CREATE TABLE transactions_local" | grep -v 'SAMPLE BY cityHash64(span_id)')
+done
```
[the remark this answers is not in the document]
BYK: Fixed the failure when table didn't exist.

I'll try the table not empty case later on as it is very unlikely to happen without the transactions consumer.

install.sh
BYK: Okay, I think this new version is clearer and also is the correct logic.

install.sh
```diff
@@ -198,6 +198,35 @@ if [ "$ZOOKEEPER_SNAPSHOT_FOLDER_EXISTS" -eq "1" ]; then
...
+   echo "Trying again. Remaining tries #$CLICKHOUSE_CLIENT_MAX_RETRY"
+  sleep 0.5;
+done
+set -e
+
+SNUBA_HAS_TRANSACTIONS_TABLE=$(clickhouse_query 'EXISTS TABLE transactions_local' | tr -d '\n\r')
+SNUBA_TRANSACTIONS_NEEDS_UPDATE=$([ "$SNUBA_HAS_TRANSACTIONS_TABLE" == "1" ] && clickhouse_query 'SHOW CREATE TABLE transactions_local' | grep -v 'SAMPLE BY' || echo '')
+
+if [ "$SNUBA_TRANSACTIONS_NEEDS_UPDATE" ]; then
+  SNUBA_TRANSACTIONS_TABLE_CONTENTS=$(clickhouse_query "SELECT * FROM transactions_local LIMIT 1")
```
fpacifici, line 219: I would just ensure these queries cannot individually transitorily fail silently. 
Like if this query fails temporarily, it is important that the failure is not treated in the same way as a "table empty" response since it may trigger a drop of the table by mistake.
Is a failure here going to stop the script ? (desirable)

BYK: The whole script is in -e mode which makes it fail on non-zero exits. I only temporarily disable that during the clickhouse server up check. So no, if they error out, the script will exit.

PR #595, opened 2020-07-15
feat(snuba): Add transactions consumer

### LINKED ISSUE #588: Snuba: we need a transactions consumer
We have enabled the Performance view on on-premise with #586 but we don't have the necessary consumer to push this data into Snuba to make the feature work.

We need to add this consumer.

BYK:
Needed for Performance to work (see #586).

Depends on #593. Fixes #588.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,7 @@ x-sentry-defaults: &sentry_defaults
     - snuba-consumer
     - snuba-outcomes-consumer
     - snuba-sessions-consumer
+    - snuba-transactions-consumer
     - snuba-replacer
     - symbolicator
     - kafka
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -120,6 +121,10 @@ services:
   snuba-sessions-consumer:
     << : *snuba_defaults
     command: consumer --storage sessions_raw --auto-offset-reset=latest --max-batch-time-ms 750
+  # Kafka consumer responsible for feeding transactions data into Clickhouse
+  snuba-transactions-consumer:
+    << : *snuba_defaults
+    command: consumer --storage transactions --auto-offset-reset=latest --max-batch-time-ms 750
```
fpacifici, line 127: I am not sure what the reasoning was to use latest as auto offset reset, but since it is used for all the other consumers, i guess this should be connnsistent

Issue #607, opened 2020-07-22
Reducing in-place upgrade downtime


ktosiek:
Currently `install.sh` stops the whole installation before pulling and building containers. Would it make sense to move that step later in the script, or even replacing it with restarting each container in turn?

ktosiek:
This is especially useful for people with more involved builds, for example when using sentry-ldap-auth: https://github.com/getsentry/onpremise/issues/253. Moving the `$dc down` step to just after ` echo "Docker images built."` reduced downtime on upgrade from ~3 minutes to ~1:30.

untitaker:
What is the most important component/endpoint that you want to keep available? Historically we did not want to commit to writing code that works for any rolling restart outside of our own production systems (e.g. can deal with data coming from old processes).

ktosiek:
I'm mostly worried about events ingestion, I think that's `relay` + `nginx`? `web` would be nice to have too.

BYK:
@ktosiek I think this makes sense as getting new images should not touch any of the already-running containers until a full down/up cycle.

Would you be willing to submit a PR for your idea & improvement?

ktosiek:
I've made a PR for the late stop of the whole cluster. To get additional seconds of availability I'd need a way to stop "everything but nginx and relay", which doesn't seem to be provided by `docker-compose`. (I'm assuming relay can queue events for a minute without Kafka and Redis - but is that true?).

BYK:
>To get additional seconds of availability I'd need a way to stop "everything but nginx and relay", which doesn't seem to be provided by docker-compose. (I'm assuming relay can queue events for a minute without Kafka and Redis - but is that true?).

I think you can use `docker-compose rm -s` as documented here: https://docs.docker.com/compose/reference/rm/
Combined with `docker-compose config --services` and a clever `grep`, it should get you the list of all services but `nginx` and `relay`. And, yes, relay can buffer things for a while so this would work as long as you don't forget to upgrade `nginx` and `relay` themselves afterwards.

This may become tricky if there are any cross-dependencies between `relay` and `sentry` though. Any thoughts @mattrobenolt, @jan-auer, @untitaker?

mattrobenolt:
> This may become tricky if there are any cross-dependencies between relay and sentry though.

This is rarely the case if ever. You should be safe 99% of the time with the assumption that they can operate independently.

untitaker:
I think for relay to stay operational while sentry is down, nginx, redis and kafka need to be up.

PR #615, opened 2020-08-04
install: Keep relay available while upgrading

ktosiek:
This continues on the ideas from #607. By "downtime" here I mean "not accepting events" - web, smtp and background processes are out of scope.

This PR adds a `--minimize-downtime` option to install.sh. This options changes the behaviour of the script by:
1. keeping nginx and relay running until the very end,
2. disabling cleanup on exit and failure,
3. explicitly reloading nginx configuration,
4. and starting the whole cluster at the end.

This is a work in progress, and only tested on a local, non-production instance. But the results are promising: no downtime if relay version doesn't change, and only a second when it does. So far this was only tested with a curl loop, so I'm still not sure if Relay flushes the events to Sentry before getting recreated by `$dc up`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -329,9 +329,17 @@ if [ ! -f "$RELAY_CREDENTIALS_JSON" ]; then
 fi
 
 if [[ "$MINIMIZE_DOWNTIME" ]]; then
+  # Start the whole setup, except nginx and relay.
+  $dc up -d --remove-orphans $($dc config --services |grep -v -e '^nginx$' -e '^relay$')
   $dc exec nginx service nginx reload
-  # Start the whole setup, this might restart nginx and relay.
-  $dc up -d --remove-orphans
+
+  echo "Waiting for Sentry to start..."
+  while [[ "$(curl -m 1 http://localhost:9000/_health/ --silent)" != "ok" ]]; do
+    sleep 0.5
+  done
+
+  # Make sure everything is up. This should only touch relay and nginx
+  $dc up -d
 else
   echo ""
   echo "----------------"
--- a/install.sh
+++ b/install.sh
@@ -21,6 +21,30 @@ RELAY_CONFIG_YML='relay/config.yml'
 RELAY_CREDENTIALS_JSON='relay/credentials.json'
 SENTRY_EXTRA_REQUIREMENTS='sentry/requirements.txt'
 
+load_options() {
+  while [[ -n "$@" ]]; do
+    case "$1" in
+	    -h | --help) show_help; exit;;
+	    --) ;;
+	    *) echo "Unexpected argument: $1. Use --help for usage information."; exit 1;;
+    esac
+    shift
+  done
+}
+
+show_help() {
+  cat <<EOF
+Usage: $0 [-h | --help]
+
+Install Sentry with docker-compose.
+
+Options:
+ -h, --help             Show this message and exit.
+EOF
+}
+
+load_options $(getopt -n "$0" -o 'h' -l 'help' -- "$@")
+
 # Courtesy of https://stackoverflow.com/a/2183063/90297
 trap_with_arg() {
   func="$1" ; shift
--- a/install.sh
+++ b/install.sh
@@ -20,11 +20,13 @@ SYMBOLICATOR_CONFIG_YML='symbolicator/config.yml'
 RELAY_CONFIG_YML='relay/config.yml'
 RELAY_CREDENTIALS_JSON='relay/credentials.json'
 SENTRY_EXTRA_REQUIREMENTS='sentry/requirements.txt'
+MINIMIZE_DOWNTIME=
 
 load_options() {
   while [[ -n "$@" ]]; do
     case "$1" in
 	    -h | --help) show_help; exit;;
+	    --minimize-downtime) MINIMIZE_DOWNTIME=1;;
 	    --) ;;
 	    *) echo "Unexpected argument: $1. Use --help for usage information."; exit 1;;
     esac
@@ -34,16 +36,19 @@ load_options() {
 
 show_help() {
   cat <<EOF
-Usage: $0 [-h | --help]
+Usage: $0 [-h | --help] [--minimize-downtime]
 
 Install Sentry with docker-compose.
 
 Options:
  -h, --help             Show this message and exit.
+ --minimize-downtime    EXPERIMENTAL: try to keep accepting events for as long as possible while upgrading.
+                        This will disable cleanup on error, and might leave your installation in partially upgraded state.
+                        This option might not reload all configuration, and is only meant for in-place upgrades.
 EOF
 }
 
-load_options $(getopt -n "$0" -o 'h' -l 'help' -- "$@")
+load_options $(getopt -n "$0" -o 'h' -l 'help,minimize-downtime' -- "$@")
 
 # Courtesy of https://stackoverflow.com/a/2183063/90297
 trap_with_arg() {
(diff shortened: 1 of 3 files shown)


BYK:
@ktosiek - okay, I'm excited about this. What keeps this PR from being ready to review?

ktosiek:
Testing - I still need to see how it will perform when relay is getting upgraded too.

With the current implementation web, Kafka and Redis would go down, and after a minute everything (including Relay) would get restarted. If Relay has no on-disk queue then this PR will only make things worse (clients would get a 200, but events would be dropped).

BYK:
I am fairly sure Relay has at least an in-memory queue. I remember having heard some storage in Redis but we need to wait for @jan-auer for confirmation. He'll be back next week.

jan-auer:
> If Relay has no on-disk queue then this PR will only make things worse

I'm afraid this is correct. If either Redis or Kafka are down, Relay will drop events. Likewise, while Relay is starting up and not authenticated, it will also drop events. Both of these are known issues. For the startup behavior, we are planning to implement a fix in the short term.

For Kafka or Redis being down we do not have fixes on the immediate roadmap. This is definitely on our radar, though, as we're looking into the queueing behavior at the moment.

BYK:
Well, then we can at least restart these sequentially at the end? Like:

1. Restart all things Sentry
2. Restart Redis
3. Restart all things Kafka
4. Restart Snuba
5. Restart Relay
6. Restart Nginx

ktosiek:
I think this is ready now - I've tested actually sending events, and every event that was accepted with 200 was visible in Sentry after upgrade (well, at least the count of events matched). Events were not being ingested for ~10 seconds - seems great for a single server setup :-)

The script I've used to send events is not included in the PR: https://gist.github.com/ktosiek/f35a51cd71b151aaf6ff7f1e981ea01f

install.sh
```diff
@@ -287,9 +328,23 @@ if [ ! -f "$RELAY_CREDENTIALS_JSON" ]; then
...
-echo ""
-echo "  docker-compose up -d"
-echo ""
+if [[ "$MINIMIZE_DOWNTIME" ]]; then
+  # Start the whole setup, except nginx and relay.
+  $dc up -d --remove-orphans $($dc config --services |grep -v -e '^nginx$' -e '^relay$')
+  $dc exec nginx service nginx reload
+
+  echo "Waiting for Sentry to start..."
+  while [[ "$(curl -m 1 http://localhost:9000/_health/ --silent)" != "ok" ]]; do
```
kohenkatz, line 337: Can there be a better way to do this for those of us not exposing port 9000?

I'm using Traefik as an edge router, so I don't expose any ports from Sentry at all.

My suggestion would be to keep `http://localhost:9000` as the default but allow overriding it, either with an environment variable (preferred) or an argument to the script. If using an argument, maybe it could be `--minimize-downtime` for no change to the URL and `--minimize-downtime=https://sentry.example.com` to override it.

kohenkatz: If you want, I can wait until this is merged and then open a new PR to add the behavior I need.

kohenkatz: Related: #512

ktosiek: Good catch! I don't really care for the proxy here, so it should probably run in a container and wait for `web:9000`.

BYK: I think we can merge #512 before to make things easier for this PR, that said we can also do something like https://unix.stackexchange.com/a/45943/108960 replacing the `tail -f` part with `$dc logs -f web` and look for some uWSGI/Django success message and avoid curling and ports altogether.

install.sh
```diff
@@ -329,9 +329,17 @@ if [ ! -f "$RELAY_CREDENTIALS_JSON" ]; then
 fi
 
 if [[ "$MINIMIZE_DOWNTIME" ]]; then
+  # Start the whole setup, except nginx and relay.
+  $dc up -d --remove-orphans $($dc config --services |grep -v -e '^nginx$' -e '^relay$')
```
BYK, line 333:
```suggestion
  $dc up -d --remove-orphans $($dc config --services | grep -v -e '^(nginx|relay)$')
```

install.sh
ktosiek: I've changed it to waiting for `http://web:9000/_health/` from an alpine container - this way I don't have to think about the port, as I'm running inside the compose project's network. Plus this sheds the `curl` dependency on host :-)

install.sh
ktosiek: Fixed in both places.

BYK:
@ktosiek - what do you say, are we ready to go?

install.sh
```diff
@@ -20,6 +20,35 @@ SYMBOLICATOR_CONFIG_YML='symbolicator/config.yml'
...
 SENTRY_EXTRA_REQUIREMENTS='sentry/requirements.txt'
+MINIMIZE_DOWNTIME=
+
+load_options() {
+  while [[ -n "$@" ]]; do
+    case "$1" in
+	    -h | --help) show_help; exit;;
+	    --minimize-downtime) MINIMIZE_DOWNTIME=1;;
+	    --) ;;
+	    *) echo "Unexpected argument: $1. Use --help for usage information."; exit 1;;
```
BYK, line 31:
```suggestion
        -h | --help) show_help; exit;;
        --minimize-downtime) MINIMIZE_DOWNTIME=1;;
        --) ;;
        *) echo "Unexpected argument: $1. Use --help for usage information."; exit 1;;
```

BYK: Let's just use spaces for consistency :)

ktosiek: I see the code mixes 2- and 4-space indents - do you think an .editorconfig (in a separate PR, of course) would help?

BYK: >I see the code mixes 2- and 4-space indents - do you think an .editorconfig (in a separate PR, of course) would help?

Possibly, for vim users I think 😀 I think ultimately we'd need a formatter and a lint rule for this file.

>And if so, what's the preferred style?

I'd go with 2-spaces to be consistent with the `docker-compose.yml` file (although my Python brain wants 4-spaces)

install.sh
```diff
@@ -287,9 +328,22 @@ if [ ! -f "$RELAY_CREDENTIALS_JSON" ]; then
...
-echo ""
-echo "----------------"
-echo "You're all done! Run the following command to get Sentry running:"
-echo ""
-echo "  docker-compose up -d"
-echo ""
+if [[ "$MINIMIZE_DOWNTIME" ]]; then
+  # Start the whole setup, except nginx and relay.
+  $dc up -d --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay)$')
+  $dc exec nginx service nginx reload
```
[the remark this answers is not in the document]
ktosiek: This line is here only for one case: when the nginx configuration changes, but the container configuration doesn't. In that case docker-compose won't restart nginx at the final `$dc up -d`.

ktosiek:
The test I've added only checks if a no-op upgrade works, but I don't have a good idea for how to do it differently.
I think this is ready to go, that's how I did my last production upgrade already ^_^

install.sh
kohenkatz: @BYK `.editorconfig` is supported in every editor I know of, including VSCode, all IntelliJ-based editors, Atom, Eclipse, Notepad++, Netbeans, SublimeText, and more. You can also set different formats for different file types (or even different file names).

There are also already linters that check all files against the editorconfig configuration, including:

- https://github.com/editorconfig-checker/editorconfig-checker
- https://github.com/jedmao/eclint
- https://gitlab.com/greut/eclint

BYK (approved):

PR #626, opened 2020-08-17
fix(env): Read and set .env in install.sh, portably

### LINKED ISSUE #622: ./install.sh does not work on CentOS 7
Starting from version 20.7.1 `install.sh` does not work on CentOS 7. Install script can't process line:
```
source <(grep -v '^#' .env | sed -E 's|^(.+)=(.*)$|: ${\1=\2}; export \1|g')
```
the error:
```
[root@sentry sentry]# source <(grep -v '^#' .env | sed -E 's|^(.+)=(.*)$|: ${\1=\2}; export \1|g')
-bash: /dev/fd/63: line 9: unexpected EOF while looking for matching `"'
-bash: /dev/fd/63: line 10: syntax error: unexpected end of file
```

chadwhitacre:
h/t https://unix.stackexchange.com/a/79077

Fixes #622, cf. #600 

cc: @NullIsNot0

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
-source <(grep -v '^#' .env | sed -E 's|^(.+)=(.*)$|: ${\1=\2}; export \1|g')
+# With a tip o' the hat to https://unix.stackexchange.com/a/79077
+set -a && . ./.env && set +a
 
 dc="docker-compose --no-ansi"
 dcr="$dc run --rm"


chadwhitacre:
From https://unix.stackexchange.com/a/79077:

> In `bash`, beware that it also causes all **functions** declared while `allexport` is on to be exported to the environment

Afaict that's not a concern here, since `.env` is a trusted file (we trust the version we're shipping; the user should trust the version they're writing).

Issue #627, opened 2020-08-17
Exercise install.sh in multiple environments in CI


chadwhitacre:
I noticed on #626 that CI only runs `install.sh` in one environment, the Travis [default](https://docs.travis-ci.com/user/reference/overview/) (Xenial at the time of writing). Seems like it'd be good to run it in multiple environments in CI, since it needs to run "everywhere" (e.g., [CentOS 7](https://github.com/getsentry/onpremise/issues/622)).

### TODO

- [x] duplicate CI in GitHub actions under Xenial: #634, #639
- [x] compare run time to Travis and consider caching
- [x] run docker in docker under GitHub actions
- [x] add CentOS 7 as a second CI environment
- [ ] add macOS as a CI environment: https://github.com/getsentry/self-hosted/pull/1587
- [x] remove Travis once we trust GitHub Actions

chadwhitacre:
Not really possible at Travis: #628.

I think the complication is that, while afaict there are CI services that allow for running in arbitrary Docker containers, the thing-under-test is itself a Docker build ... so would we need to run Docker inside Docker? 🤔

BYK:
Oh, I think this is a neat idea! Running docker "inside" Docker is a lot easier than you may think. Most of the time mounting `/var/run/docker.sock` into the container under the same path "just works". Combined with a switch to GitHub Actions we can support both MacOS and CentOS builds.

I'll admit that running this repo on non-Linux environments is not really a goal but it may be useful to detect incompatibilities.

chadwhitacre:
I've updated the description with a TODO, and kicked out a first PR: #634.

chadwhitacre:
Travis: [5m 44s](https://travis-ci.com/github/getsentry/onpremise/builds/180124681)
GitHub: [5m 2s](https://github.com/chadwhitacre/onpremise/runs/998127947)

💃

chadwhitacre:
[...]

PR #628, opened 2020-08-17
Run CI on more than one OS

chadwhitacre:
Addresses #627.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: bash
 services: docker
 
+os:
+  - linux
+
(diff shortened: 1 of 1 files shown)


chadwhitacre:
Uh-oh ...

> We do not currently support use of Docker on macOS.

https://docs.travis-ci.com/user/docker/

Indeed, the documentation is sound. 😌 

> Services are not supported on osx

https://travis-ci.com/github/getsentry/onpremise/jobs/372588710#L52

Okay! Not so easy! 😖

PR #630, opened 2020-08-17
upgrade(clickhouse): Use the Clickhouse version we use in prod

BYK:


joshuarli:
I don't believe we have consistent clickhouse versions but the minimum we're running right now is 20.3.9.70. @JTCunning might have a better answer as to what a good version to upgrade to might be.

BYK:
I got this from our ops repo 🤷

JTCunning:
Hi.
`20.3.9.70` is best, for now. The value is overridden on the machines that deal with accepting writes from snuba.

It's to avoid this bug (https://github.com/ClickHouse/ClickHouse/issues/12583) that won't land until 20.7.

docker-compose.yml
```diff
@@ -96,7 +96,7 @@ services:
       - 'sentry-secrets:/etc/kafka/secrets'
   clickhouse:
     << : *restart_policy
-    image: 'yandex/clickhouse-server:19.17'
+    image: 'yandex/clickhouse-server:20.4.3.16'
```
joshuarli, line 99:
```suggestion
    image: 'yandex/clickhouse-server:20.3.9.70'
```

joshuarli (approved):
Approving ahead of time, assuming the suggestion's merged.

PR #634, opened 2020-08-18
ci(github): Reimplement CI in GitHub Actions

chadwhitacre:
This is a fairly wooden port of the Travis CI config to GitHub actions. Note that this doesn't remove Travis yet. I figure we should run both in parallel for a bit to make sure GitHub Actions is trustworthy. E.g., there might be some caching we want to do to speed up the build, I didn't look into that yet.

Presumably there's some button clicking that needs to happen before this'll run in the main repo. Here's a good run in my fork:

https://github.com/chadwhitacre/onpremise/runs/998127947

And here's a sample bad run in my fork, showing that test failures indeed break the build:

https://github.com/chadwhitacre/onpremise/runs/998137076

Part of #627.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,33 @@
+name: test
+on: push
+env:
+  DOCKER_COMPOSE_VERSION: 1.24.1
+jobs:
+  test:
+    runs-on: ubuntu-16.04
+    name: "test"
+    steps:
+
+      - name: Pin docker-compose
+        run: |
+          sudo rm /usr/local/bin/docker-compose
+          curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+          chmod +x docker-compose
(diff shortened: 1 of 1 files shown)


chadwhitacre:
Ready for review, @BYK et al.

BYK (approved):
This is just brilliant 🎇 

Thank you so much for doing this!

BYK:
GHA has been quite stable for some other things so I think it should be safe to remove Travis and see if we can make this even better. Maybe separate install and actual test steps? I don't know if we can leverage any caching easily as you indicated but why not try?

.github/workflows/test.yml
```diff
@@ -0,0 +1,33 @@
+name: test
+on: push
```
BYK, line 2: Oh wait, this will only run on branch pushes. We should also add `pull_request` for this to be a proper commit check, right @chadwhitacre?

chadwhitacre: Yup, good call. Maybe https://github.com/getsentry/onpremise/pull/639 does what we want?

PR #638, opened 2020-08-21
Test whether GHA runs on push event for new PR

chadwhitacre:
Knowledge-gathering pursuant to https://github.com/getsentry/onpremise/pull/634#discussion_r474287108. 🕵️


chadwhitacre:
Action did not run, I _think_ because no push action was fired?

PR #642, opened 2020-08-23
Use the custom Nginx access log format

### LINKED ISSUE #635: nginx custom log format isn't actually used anywhere
The nginx configuration has the following directive to set a custom log format:

```
log_format main '$remote_addr - $remote_user [$time_local] "$request" '
'$status $body_bytes_sent "$http_referer" '
'"$http_user_agent" "$http_x_forwarded_for"';
```

However, this is not actually used anywhere. To make it work, the following line needs to be added:

```
access_log /var/log/nginx/access.log main;
```

Without that line, the access log simply uses the default `combined` log format.

In the interest of simplifying the user experience for [users trying to make changes to the log format](https://forum.sentry.io/t/changes-to-nginx-configuration-not-being-applied/10789), either the `access_log` directive should be added or the `log_format` directive should be removed.

kohenkatz:
If we are going to define a custom log format, we should probably be using that format.

Fixes #635

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -17,6 +17,8 @@ http {
 	'$status $body_bytes_sent "$http_referer" '
 	'"$http_user_agent" "$http_x_forwarded_for"';
 
+	access_log /var/log/nginx/access.log main;
+
 	sendfile on;
 	tcp_nopush on;
 	tcp_nodelay on;


nginx/nginx.conf
```diff
@@ -17,6 +17,8 @@ http {
 	'$status $body_bytes_sent "$http_referer" '
 	'"$http_user_agent" "$http_x_forwarded_for"';
 
+	access_log /var/log/nginx/access.log main;
```
BYK, line 20: We may also use `/dev/stdout` to be more explicit as this is already what `nginx` does:

https://github.com/nginxinc/docker-nginx/blob/aa41ddeef871b7f0ea64a44f26d3f4aa0e6d5e7b/mainline/buster/Dockerfile#L95-L97

kohenkatz: Using the symlinks that the Nginx Docker image already provides means that anyone who already uses the Docker Nginx container will be familiar with what is going on.

This way, if someone wants to redirect the nginx logs to a file on disk, all they need to do is add a `volume` entry to the `docker-compose.yml` file for the `/var/log/nginx` directory (or for the `access.log` and/or `error.log` files). However, if we explicitly default to logging to `/dev/stdout`, then someone who wants to make that change will have to change this file too and might not expect to need to do that.

I think that sticking with the Nginx default log paths is cleaner and easier to use.

BYK: Great reasoning, thanks a lot for taking the time for this detailed response!

BYK (approved):

PR #646, opened 2020-08-26
install: Add flag to skip create user prompt during install

### LINKED ISSUE #611: Pass Flag to install script for User creation prompt
Why isn't it possible to simply pass `-n` or another flag to the install script to _not_ create a user? I'm installing through Puppet and it keeps breaking at this step (because of the interactivity I'm assuming). I see that the script does check for an interactive prompt, but it would be nice to have this alternative approach for more flexibility in the install process. If I missed something please point me in the right direction!

bonzobavaria:
Fixes #611.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docke
 
 To get started with all the defaults, simply clone the repo and run `./install.sh` in your local check-out.
 
+During the install, a prompt will ask if you want to create a user account. If you require that the install not be blocked by the prompt, run `./install.sh --no-user-prompt`, or `./install.sh -n`.
+
 There may need to be modifications to the included example config files (`sentry/config.example.yml` and `sentry/sentry.conf.example.py`) to accommodate your needs or your environment (such as adding GitHub credentials). If you want to perform these, do them before you run the install script and copy them without the `.example` extensions in the name (such as `sentry/sentry.conf.py`) before running the `install.sh` script.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,6 +15,8 @@ Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docke
 
 To get started with all the defaults, simply clone the repo and run `./install.sh` in your local check-out.
 
+During the install, a prompt will ask if you want to create a user account. If you require that the install not be blocked by the prompt, run `./install.sh --no-user-prompt`, or `./install.sh -n`.
```
BYK, line 18: As much as I like short flags, `-n` is usually reserved for "dry run" so let's not use that for now.

```suggestion
During the install, a prompt will ask if you want to create a user account. If you require that the install not be blocked by the prompt, run `./install.sh --no-user-prompt`.
```

install.sh
```diff
@@ -1,4 +1,17 @@
 #!/usr/bin/env bash
+
+skip_user_prompt=0
+
+while [[ -n $1 ]]; do
+  case $1 in
+    -n | --no-user-prompt)
```
BYK, line 7: Removing `-n` per my other comment.
```suggestion
    --no-user-prompt)
```

Also, this part is likely to clash with the upcoming PR: #615

bonzobavaria:
This can wait for #615 to be merged. I'll rebase and include the needed flag in the arg parsing introduced there.

BYK (approved):

Issue #616, opened 2020-08-05
Memory usage / documentation


nigeisel:
We have a rather small setup with ~1 event / minute and usually not more than one/two user using the web interface at the same time. However we recorded a combined memory usage well above 4G with both Kafka and Clickhouse using more than 1G each.

Since the docs mention 2400MB memory as a minimum I was wondering whether this is actually realistic or if something goes wrong on our side? Are there are ways to configure the setup to reduce memory usage? Otherwise maybe the docs should be updated to show a more realistic minimum requirement (about 5g?).

Memory usage: Top graphs are kafka and clickhouse, followed by the worker and the web container.
<img width="811" alt="Screen Shot 2020-08-05 at 17 45 50" src="https://user-images.githubusercontent.com/9933308/89435073-a4272c00-d744-11ea-8677-5d7ad34753e9.png">

Memory usage combined:
<img width="876" alt="Screen Shot 2020-08-05 at 17 45 19" src="https://user-images.githubusercontent.com/9933308/89435050-9f627800-d744-11ea-97a9-c73865b7d060.png">

McSneaky:
You can reduce system load quite a bit by reducing Kafka size 
Here's an issue about it that also has some workarounds in comments: https://github.com/getsentry/onpremise/issues/502

renchap:
I am having a look at the memory usage, as you need over 4GB of memory to run this without crashes after a few hours/days due to out of memory errors and the OOM Killer killing either clickhouse or Kafka.

For Kafka, limiting the number of shards using `KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS=2` as said in #502 really helps.

For Clickhouse, I looked at the settings and found this option: https://clickhouse.tech/docs/en/operations/server-configuration-parameters/settings/#max_server_memory_usage_to_ram_ratio

It allows you to specify a percentage of the max total memory available to Clickhouse. Without it, Clickhouse does not limit the memory used by queries and defaults to a max of 10G, which is far overkill for most on-premise sentry installations.

I would suggest to configure `max_server_memory_usage_to_ram_ratio` to something like 0.3, or `max_server_memory_usage` to 1G.

The easiest way to do this seems to build a custom clickhouse container, with one additional file in `/etc/clickhouse-server/config.d/sentry.xml` containing the custom settings, something like:

```xml
<yandex>
  <max_server_memory_usage_to_ram_ratio>0.3</max_server_memory_usage_to_ram_ratio>
</yandex>
```

[...]

PR #651, opened 2020-08-29
Allow to limit clickhouse memory usage

renchap:
Currently Clickhouse memory is not limited, and it requires at least 10G of memory by default.

In this PR, I introduce a way to configure Clickhouse max memory using an environment variable. I chose to set it relative to the host max memory rather than a fixed amount so it will adapt to the size of the host. This is especially useful for people deploying Sentry on hosts with limited memory.

See related discussion in #616

### DIFF
--- a/clickhouse/Dockerfile
+++ b/clickhouse/Dockerfile
@@ -1,0 +1,4 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+COPY ./sentry.xml /etc/clickhouse-server/config.d/sentry.xml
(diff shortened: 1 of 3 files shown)


BYK (changes requested):
Hi @renchap, thanks for taking the time to build this up! The Clickhouse repo recommends using a volume mount for this, which I also prefer: https://github.com/ClickHouse/ClickHouse/tree/v20.3.9.70-lts/docker/server#start-server-instance-with-custom-configuration

Would you mind making this so we just mount the config file to `/etc/clickhouse-server/config.xml`?

We do something very similar with `nginx` so I think you can just reuse that logic.

BYK:
@JTCunning adding you as the SME as I have no idea if the 30% ratio would cause any issues. Based on your comments on #616, I think this is safe for smaller scales which the on-premise repo is intended for.

renchap:
Thanks for patching my PR, I have been very busy in the last week and did not had time to make the change!

BYK:
Closing as #662 was merged instead.

PR #655, opened 2020-09-03
feat(sentry): Disallow internal system IPs

BYK:
Set `SENTRY_DISALLOWED_IPS` to `INTERNAL_SYSTEM_IPS` to prevent source and sourcemap requests to internal services in the Docker network.

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -31,6 +31,7 @@ def get_internal_network():
 
 
 INTERNAL_SYSTEM_IPS = get_internal_network()
+SENTRY_DISALLOWED_IPS = INTERNAL_SYSTEM_IPS
 
 
 DATABASES = {


mattrobenolt:
So I've personally been on the fence about this setting in us setting a "safe" default or not.

imo it's pretty common that an on-premise user would want access to internal network resources. For example their own internal GitHub Enterprise, GitLab, Jira, whatever.

Pairing that with user's on-premise installations aren't usually exposed to the public internet (I'm making a broad assumption with no data here), so the risk is lower.

I'd rather roll this up into a generic setting for like, "Allow accessing internal networks" as a configuration rather than the granular `DISALLOWED_IPS` setting we have. I can't think of many or any situations where you'd need/want to fine grain set IPs here. It's more of an all or nothing situation imo.

And I think this is a pretty common thing that should be easy to toggle from the admin UI or something. And the use cases are going to vary per user.

With that, I'd be fine with a secure by default, if it's easy to turn off and opt into accessing internal networks, which this PR doesn't give.

BYK:
@mattrobenolt I share the same concerns, that said if you think about it, the "internal network" here is not a company intranet or something, it is just the private internal network that Docker Compose creates. So it felt pretty safe (in terms of not breaking functionality) and rational to add this restriction.

What do you say?

mattrobenolt:
I guess this network is private from the normal internal network as an intranet. I guess this is just a bag of worms. I'm not sure the best options then since I don't have any real experiences with this. We wouldn't run this in an isolated network if we needed to poke our actual internal services. But I don't think this repo would even allow you to do that easily.

BYK:
>I guess this network is private from the normal internal network as an intranet.

Yes 🙂 

>I guess this is just a bag of worms.

Why do you think it is a bag of worms?

>I'm not sure the best options then since I don't have any real experiences with this. We wouldn't run this in an isolated network if we needed to poke our actual internal services.

Docker and Docker Compose allows attaching to this internal network. It's just not shared by default. Classic docker stuff, if you think about it.

>But I don't think this repo would even allow you to do that easily.

Exactly. The repo defaults to this internal private network created by docker-compose and relies on the private DNS resolution for service configuration anyway. This just builds on top of that. I think it makes sense given all this.

mattrobenolt:
> Why do you think it is a bag of worms?

Because if you allow running as like, `--net=host` to avoid the docker networking, then everything will fall apart here because of how you rely on service discovery. So to even support this would become tricky. I'm not sure if there's a way to effectively NAT to "external" network? I dunno. I've never done this.

> Docker and Docker Compose allows attaching to this internal network. It's just not shared by default. Classic docker stuff, if you think about it.

Yeah, I know, but if I run this in my infrastructure, sentry is it's own isolated network. And say I have Jira running next to it on a different internal address, but not within the sentry network, it can't access it. And I'm not sure how to accomplish that if someone wanted to.

mattrobenolt (approved):
Let's go with this for now until it causes problems.

BYK:
>Yeah, I know, but if I run this in my infrastructure, sentry is it's own isolated network. And say I have Jira running next to it on a different internal address, but not within the sentry network, it can't access it. And I'm not sure how to accomplish that if someone wanted to.

They need to expose their ports and then they can go through the special `host.docker.internal` host name. This now works in Linux too. They may also be able to "join" these networks but I don't know enough about that.

mattrobenolt:
Is `host.docker.internal` a fixed IP that we can rely on? If so, it'd be nice to exclude that. I'm not sure how this works.

BYK:
You know what, I'm going to punt on this as the risk it mitigates is quite low.

BYK:
>Is `host.docker.internal` a fixed IP that we can rely on? If so, it'd be nice to exclude that. I'm not sure how this works.

Practically it is, techincally, it is not guaranteed to be.

PR #662, opened 2020-09-08
feat(clickhouse): Reduce max memory usage to 30% of RAM

### LINKED ISSUE #616: Memory usage / documentation

BYK:
Closes #616, supercedes #651

Adds an option to reduce max memory usage of Clickhouse server. Sets it to 30% of all available RAM as the default.

Co-authored-by: @renchap

### DIFF
--- a/clickhouse/config.xml
+++ b/clickhouse/config.xml
@@ -1,0 +1,3 @@
+<yandex>
+    <max_server_memory_usage_to_ram_ratio from_env="MAX_MEMORY_USAGE_RATIO" />
+</yandex>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,6 +97,11 @@ services:
   clickhouse:
     << : *restart_policy
     image: 'yandex/clickhouse-server:20.3.9.70'
+    environment:
+      # This limits Clickhouse's memory to 30% of the host memory
+      # If you have high volume and your search return incomplete results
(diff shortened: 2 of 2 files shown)


mattrobenolt (approved):

amitseth7:
@BYK , does the PR merge resolve the issue of high memory usage by clickhouse . Do you suggest a re install with the newest images yet?

BYK:
@amitseth7 this should resolve the high memory usage of Clickhouse, yes. To be able to use this fix, you don't need a re-install with new images.

You can just apply this PR manually and do `docker-compose restart clickhouse`. Make sure to check the log output to see `sentry.xml` getting applied in the init phase of `clickhouse` and you should be good.

PR #663, opened 2020-09-08
feat(snuba): Run new migration system

lynnagara:
This is for the onpremise release on Sept 15th.

The new migration system has a migration to handle recreating the
transaction table if the old one is present, we no longer need to do
this in install.sh.


BYK (approved):
Wow, so much code removal!

wow

PR #664, opened 2020-09-08
ci(gha): Fix intermittent test failures

BYK:
GitHub team pointed us to docker/compose#3586 as the likely root cause and some digging around got us to https://git.io/JUn7p as a potential fix, which we are trying here.


BYK:
At least it passed on the first try here so will merge and see what happens.

PR #673, opened 2020-09-14
fix(install): Use proper bash testing

RichardBronosky:
Fixes https://github.com/getsentry/onpremise/issues/672

I split this PR up into 4 commits. The first one is the bare minimum for the issue. The rest are just consistency corrections that we neckbeards at irc://chat.freenode.net/%23bash would always make.


BYK (approved):
I tried to avoid bash-izms in the past but we already passed that point. Thanks for bringing order to this chaos! :)

PR #674, opened 2020-09-14
fix(kafka): Reduce Kafka resource usage

### LINKED ISSUE #502: Kafka starts with 50 partitions bogging down system resources
For some reason Kafka starts with 50 partitions, which takes unnecessary system resources

`kafka-topics --describe events --zookeeper zookeeper:2181`
```
Topic:__consumer_offsets    PartitionCount:50    ReplicationFactor:1    Configs:segment.bytes=104857600,cleanup.policy=compact,compression.type=producer
    Topic: __consumer_offsets    Partition: 0    Leader: 1001    Replicas: 1001    Isr: 1001
    Topic: __consumer_offsets    Partition: 1    Leader: 1001    Replicas: 1001    Isr: 1001
...
```
50 partitions is extremely huge. Some systems are allocating 1 partition for 1k req/sec as their default. Which is relatively good. I doupt anyone will have 50k error requests incoming every second and runs Sentry off with docker-compoe

Since with docker-compose all partitions run on the same host it's just a waste of system resources to run so many partitions on single host

It would be nice to scale down default number of partitions to some. Or is there some exposed setting to start less of them (didn't find it from doc anywhere)

More reading: https://www.confluent.io/blog/how-choose-number-topics-partitions-kafka-cluster/

_Running it on double the minimum requirements EC2 instance, recieving < 7k errors per day (evenly spread), which makes it about 0.08 req/sec, but 100% of RAM and CPU is constantly allocated by it, which makes server unresponsive and crashing constantly_

BYK:
Fixes #502 and applies the suggestions from there:

- Number of partitons=1 (from 40)
- Log retention to 1 day (from 7 days)

These settings should be more suited towards the scale this repo is intended for.

NOTE: The partition count change will only affect new installs unless `sentry-kafka` and related volumes are cleaned.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.4'
+mpoomposeversion: '3.4'
 x-restart-policy: &restart_policy
   restart: unless-stopped
 x-sentry-defaults: &sentry_defaults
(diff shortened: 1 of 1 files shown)


JTCunning (approved):
🍪 

>Number of partitons=1 (from 40)
>Log retention to 1 day (from 7 days)

Ouch. These are some pretty big defaults. Good call.

BYK:
>Good call.

Wish I was the one who noticed and found better defaults. Hail the community hail @mattrobenolt! 🙂

mattrobenolt (commented):
It'd probably be a good idea to mirror these over into devservices: https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L1498-L1508

Issue #629, opened 2020-08-17
Worker / celery stops working


buffcode:
After upgrading to `20.8.0.dev 069e8ccd` events stop showing up in the frontend sporadically.

The UI shows `Background workers haven't checked in recently. It seems that you have a backlog of 71 tasks. Either your workers aren't running or you need more capacity.`

The worker log shows:
```
11:42:28 [INFO] sentry.deletions.async: object.delete.executed (object_id=26394L model=u'Group' transaction_id=u'2109ad23955d4c59                               91283be5016442b7')
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/celery/worker/consumer/consumer.py", line 316, in start
    blueprint.start(self)
  File "/usr/local/lib/python2.7/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/usr/local/lib/python2.7/site-packages/celery/worker/consumer/consumer.py", line 592, in start
    c.loop(*c.loop_args())
  File "/usr/local/lib/python2.7/site-packages/celery/worker/loops.py", line 91, in asynloop
    next(loop)
  File "/usr/local/lib/python2.7/site-packages/kombu/asynchronous/hub.py", line 354, in create_loop
    cb(*cbargs)
  File "/usr/local/lib/python2.7/site-packages/kombu/transport/redis.py", line 1047, in on_readable
    self.cycle.on_readable(fileno)
  File "/usr/local/lib/python2.7/site-packages/kombu/transport/redis.py", line 344, in on_readable
    chan.handlers[type]()
  File "/usr/local/lib/python2.7/site-packages/kombu/transport/redis.py", line 721, in _brpop_read
    **options)
  File "/usr/local/lib/python2.7/site-packages/redis/client.py", line 680, in parse_response
    response = connection.read_response()
  File "/usr/local/lib/python2.7/site-packages/redis/connection.py", line 624, in read_response
    response = self._parser.read_response()
  File "/usr/local/lib/python2.7/site-packages/redis/connection.py", line 403, in read_response
    (e.args,))
ConnectionError: Error while reading from socket: ('Connection closed by server.',)
11:43:29 [WARNING] celery.worker.consumer.consumer: consumer: Connection to broker lost. Trying to re-establish the connection...
Restoring 13 unacknowledged message(s)
```

Restarting the worker reprocesses the events and works for a short time. The connection won't be re-established as printed.

[...]

PR #681, opened 2020-09-22
fix(redis): Increase file descriptors to 10032

BYK:
@xbenjii reported the following error on #629:
>You requested maxclients of 10000 requiring at least 10032 max file descriptors.

Increasing this limit by default makes sense to make Redis more available to heaveier loads.


JTCunning (approved):
🍪 
I have never been (and will never be) in a situation in which limiting file descriptors has helped anyone.

mattrobenolt (approved):

PR #684, opened 2020-09-22
fix(kafka): Enable auto topic creation

### LINKED ISSUE #683: ingest-consumer lacks related kafka topic
## Version Information

Version: 20.9.0

## Description

https://github.com/getsentry/onpremise/issues/629#issuecomment-696627325

We are getting the following error from `ingest-consumer`:
>ingest-consumer_1              | Exception: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: ingest-attachments: Broker: Unknown topic or partition"}

## Steps to Reproduce

Run the onpremise instances, wait a bit.

## Logs

Please share any applicable logs:

>ingest-consumer_1              | Exception: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: ingest-attachments: Broker: Unknown topic or partition"}

BYK:
We don't explicitly create topics outside of Snuba which causes issues with Sentry<->Relay-only topics like `ingest-attachments`. Enabling this will save us from any future troubles around forgotten kafka topics and maintaining a separate list here on onpremise.

Fixes #683.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -88,7 +88,7 @@ services:
       KAFKA_LOG_RETENTION_HOURS: '24'
       KAFKA_MESSAGE_MAX_BYTES: '50000000' # 50MB or bust
       KAFKA_MAX_REQUEST_SIZE: '50000000' # 50MB on requests apparently too
-      KAFKA_AUTO_CREATE_TOPICS_ENABLE: 'true' # We need this for topic used outside of Snuba (attachments etc.)
+      KAFKA_PRODUCER_AUTO_CREATE_TOPICS: 'client-side' # We need this for topic used outside of Snuba (attachments etc.)
       CONFLUENT_SUPPORT_METRICS_ENABLE: 'false'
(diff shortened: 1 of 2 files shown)


mattrobenolt:
On one hand I'm ok with this, on the other hand, I'm not sure if this falls under plans that @getsentry/sns were doing to better handle migrations? I did a while back, stuff in Snuba to explicitly define and manage topics.

There's also this tool that @b1naryth1ef was going to look into for Single Tenant, but we haven't investigated yet. https://github.com/segmentio/topicctl

I'm just not sure if auto topic creation is just a good long term answer, if it was, I feel like we'd just do this in SaaS and ST and ignore trying to manage anything manually.

BYK:
@mattrobenolt this was for non-Snuba topics (`ingest-attachments` etc.) and I agree that auto creation isn't ideal but it should have been fine for self-hosted. Anyway, since it is now super hard to enable, I give up 😦

mattrobenolt:
Yeah, I think this is just all fall out of upgrading the kafka image beyond what we use, since apparently we did rely on this implicit behavior. We don't maintain a canonical list of topics needed outside of Snuba yet for this apparently.

PR #685, opened 2020-09-22
fix(ingest-consumer): Create the missing Kafka topics

### LINKED ISSUE #683: ingest-consumer lacks related kafka topic

BYK:
Fixes #683, supercedes #684.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -245,6 +245,16 @@ $dcr snuba-api bootstrap --no-migrate --force
 $dcr snuba-api migrations migrate --force
 echo ""
 
+# NOTE: This step relies on `kafka` being available from the previous `snuba-api bootstrap` step
+# XXX(BYK): We cannot use auto.create.topics as Confluence and Apache hates it now (and makes it very hard to enable)
+# TODO(BYK): Make this compatible with a list of topics & sync that from a central source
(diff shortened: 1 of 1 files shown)


mattrobenolt (approved):
glhf, maybe look into https://github.com/segmentio/topicctl for longer term since it's what we were looking to adopt.

jan-auer (commented):
Since we're only talking about four topics, you could consider just listing them out. Even though we're going with identical configuration initially, each of the topics has very different load patterns, so it may make sense to set separate operations in the future.

PR #686, opened 2020-09-25
fix(install): Check that all ingest topics exist

tonyo:
Probably this is not needed anymore after https://github.com/getsentry/sentry/pull/20984 is merged, but just in case


BYK (approved):
lol, thanks! :D

PR #690, opened 2020-10-01
State that config files need to be updated as well

pataar:
Our relay connection wasn't behaving correctly because of some missing config in our `sentry.conf.py`. Also, we didn't have the performance module enabled because of a missing feature flag.

That's why an entry in the documentation regarding this would be nice.


BYK (approved):
Thanks a lot for this @pataar! We actually state this over at https://develop.sentry.dev/self-hosted/#upgrading but I guess it needs to stand out a bit more.

pataar:
Thanks @BYK!

PR #697, opened 2020-10-06
ref(config): Move GitHub config over to config.yml

BYK:
Follow up to getsentry/sentry#21041

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -76,6 +76,7 @@ transaction-events.force-disable-internal-project: true
 # GitHub Integration #
 ######################
 
+# github-app.extended-permissions: ['repo']
 # github-app.id: GITHUB_APP_ID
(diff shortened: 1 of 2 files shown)


sentry/config.example.yml
```diff
@@ -76,6 +76,7 @@ transaction-events.force-disable-internal-project: true
 # GitHub Integration #
 ######################
 
+# github-app.extended-permissions: ['repo']
```
dashed, line 79: I think this is supposed to be `github-login.extended-permissions`. Or are they the same?

BYK: They are the same on self-hosted (when in `SENTRY_SINGLE_ORGANIZATION` mode). See the hack here: 

https://github.com/getsentry/sentry/pull/21041/files#diff-208a1aca957f5fa6a22fb24ad1141cecR145-R151

dashed (approved):

PR #702, opened 2020-10-14
feat(py3): Add Python 3 support via SENTRY_PYTHON3 env variable

BYK:


joshuarli (commented):
Can this be SENTRY_PYTHON3 to avoid confusion, since that's what's used throughout sentry (mostly CI but it also appears in setup.py)?

sentry/Dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG SENTRY_IMAGE
-FROM ${SENTRY_IMAGE}
+ARG SENTRY_PYTHON3
+FROM ${SENTRY_IMAGE}${SENTRY_PYTHON3:+-py3}
```
joshuarli, line 3: Huh, TIL docker has limited support for shell-style envvar expansion. https://docs.docker.com/engine/reference/builder/#environment-replacement

joshuarli (approved):

PR #731, opened 2020-10-29
fix(install): Increase stop timeout to 60 seconds

BYK:
This is to ensure clean shutdown of Celery, with fully drained queues. This is needed as versions may change the event format and not be backwards compatible. FWIW this is a hacky workaround without a strong guarantee that the queues will be empty. Ideally we'd shutdown everything first, spin up the workers and check for queues being drained every second or so.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -15,6 +15,11 @@ MIN_DOCKER_VERSION='19.03.6'
 MIN_COMPOSE_VERSION='1.24.1'
 MIN_RAM=2400 # MB
 
+# Increase the default 10 second SIGTERM timeout
+# to ensure celery queues are properly drained 
+# between upgrades as event formats may change across
+# versions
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -15,6 +15,11 @@ MIN_DOCKER_VERSION='19.03.6'
 MIN_COMPOSE_VERSION='1.24.1'
 MIN_RAM=2400 # MB
 
+# Increase the default 10 second SIGTERM timeout
+# to ensure celery queues are properly drained 
+# between upgrades as event formats may change across
```
markstory, line 20:
```suggestion
# between upgrades as task signatures may change across
```

markstory (commented):
Looks good to me.

markstory (approved):

PR #734, opened 2020-11-09
feat(search): Enable advanced search for self-hosted

BYK:
Addresses https://forum.sentry.io/t/advanced-search-access-in-onpremise-install/11736?u=byk


markstory (commented):
Looks good to me. This only have a handful of checks and they should all be fine in self-hosted scenarios.

markstory (approved):

PR #735, opened 2020-11-11
feat: Allow metric alerts to be used in on-prem

wedamija:
This enables metric alerts for all on-prem users. There wasn't too much to do, mostly just making 
sure the consumers are running, and then enabling the feature.


wedamija:
Looking at the logs, initially I saw some `cimpl.KafkaException: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: events-subscription-results: Broker: Unknown topic or partition"}`. This should resolve itself as events come in, but if we want to avoid it we could set `KAFKA_CONSUMER_AUTO_CREATE_TOPICS = True` in the config. Then once https://github.com/getsentry/sentry/pull/21801 is merged we should be good.

BYK:
>This should resolve itself as events come in, but if we want to avoid it we could set KAFKA_CONSUMER_AUTO_CREATE_TOPICS = True in the config.

Can we set this in https://github.com/getsentry/sentry/blob/master/docker/sentry.conf.py or in our default `server.py` config?

BYK (commented):
LGTM except for the comment I left. Also I'm assuming you tested this?

wedamija:
> > This should resolve itself as events come in, but if we want to avoid it we could set KAFKA_CONSUMER_AUTO_CREATE_TOPICS = True in the config.
> 
> Can we set this in https://github.com/getsentry/sentry/blob/master/docker/sentry.conf.py or in our default `server.py` config?

Actually it is already set to True, I realised I just saw the error due to https://github.com/getsentry/sentry/pull/21801 not being merged.

> LGTM except for the comment I left. Also I'm assuming you tested this?

Yep, mentioned in a comment above.

BYK (approved):

PR #739, opened 2020-11-17
fix(metric_alerts): Fix transaction alerts

wedamija:
I only tested error alerts while testing, turns out this was broken in both dev and on-prem. This
fixes the issue.


wedamija:
Note: I've tested this out in my on-prem install and it's all working fine now

BYK (approved):

PR #741, opened 2020-11-17
fix(install): Disable the script on git-bash

BYK:
We kept getting issue reports that we traced down to `git-bash` which doesn't seem to play nice with Docker for Windows with bind mounts. This PR uses the existence of `$MSYSTEM` to detect `git-bash` and exit early with a relevant message.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 set -e
 
+if [ -x ${MSYSTEM+x} ]; then
+  echo "Seems like you ar using Git-Bash which is not supported. Please use WSL instead.";
+  exit 1
+fi
+
 # Read .env for default values with a tip o' the hat to https://stackoverflow.com/a/59831605/90297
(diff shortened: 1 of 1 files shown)


BYK:
Related: https://github.com/getsentry/onpremise/issues/705#issuecomment-728903946

/cc @sdohle

install.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 set -e
 
+if [[ -n "$MSYSTEM" ]]; then
+  echo "Seems like you ar using Git-Bash which is not supported. Please use WSL instead.";
```
chadwhitacre, line 5:
```suggestion
  echo "Seems like you are using an MSYS2-based system (such as Git Bash) which is not supported. Please use WSL instead.";
```

chadwhitacre (changes requested):

chadwhitacre (approved):

PR #754, opened 2020-11-27
fix(sentry): Pass SENTRY_EVENT_RETENTION_DAYS to sentry services

BYK:
We are already referencing this env var here:

https://github.com/getsentry/onpremise/blob/19f4561a9e2abe32dc5eb5a03a332b50f2265b4b/sentry/sentry.conf.example.py#L62-L64

@lynnagara should we be passing this to Snuba and any other services too?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ x-sentry-defaults: &sentry_defaults
   environment:
     SENTRY_CONF: '/etc/sentry'
     SNUBA: 'http://snuba-api:1218'
+    SENTRY_EVENT_RETENTION_DAYS: 
   volumes:
     - 'sentry-data:/data'
     - './sentry:/etc/sentry'


docker-compose.yml
```diff
@@ -25,6 +25,7 @@ x-sentry-defaults: &sentry_defaults
   environment:
     SENTRY_CONF: '/etc/sentry'
     SNUBA: 'http://snuba-api:1218'
+    SENTRY_EVENT_RETENTION_DAYS:
```
mattrobenolt, line 28: Don't you uhh, need a value here?

BYK: Nope. This is how you pass the value defined on the host to the container/service. The default value of this var is defined in `.env` and we want to pass that value.

chadwhitacre:
@BYK Is there some documentation we should update with this?

BYK:
@chadwhitacre not really as we already added this a long time ago (when switching to v10). This is just an oversight where we forgot to pass its value to Sentry containers so they operated on the 90-day default anyway.

lynnagara (approved):
Seems good

lynnagara:
I'm surprised no one raised this issue earlier. We should also find a way to update Snuba's default retention to match this value. https://github.com/getsentry/snuba/blob/master/snuba/settings.py#L91.

BYK:
@lynnagara: https://github.com/getsentry/snuba/pull/1526

After that we need another patch here to pass that variable to Snuba instances.

docker-compose.yml
lynnagara: Is it confusing because of the trailing colon?

BYK: It definitely is but I don't really know a better way :(

chadwhitacre: I forget ... can we put comments in YAML? It's JSON where we can't really comment, right?

Worth a comment here?

(I mean, this is already merged, but ... )

mattrobenolt: FOO: null

Would be equivalent and look at least intentional.

BYK: Added the comments in #759 which is the follow up.

I don't want `FOO: null` as it also looks confusing: did you want to unset this? What? Why?

PR #761, opened 2020-12-03
ci(release): Use action-prepare-release and rework dry-run

BYK:
Switches to using getsentry/action-prepare-release for common steps and removes the dry run option from the prepare stage as it prevents us from testing the publish step. Prepare step should be safe anyways and can be skipped with the `skip-prepare` input.

Related: https://app.asana.com/0/1198192131329257/1198192131329301, https://app.asana.com/0/1198192131329257/1198192573081717

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,43 +27,20 @@ jobs:
     runs-on: ubuntu-latest
     name: 'Release a new version'
     steps:
-      - id: killswitch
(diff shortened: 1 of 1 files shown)


chadwhitacre (approved):
Passes the prepare step (but not the _craft_ prepare step):

https://github.com/getsentry/onpremise/runs/1493705249?check_suite_focus=true

Issue #698, opened 2020-10-07
Install script crash when using a custom network


soullivaneuh:
In order to setup a Traefik reverse proxy, I put the following extra configuration on `docker-compose.override.yml`:

```yaml
version: '3.4'

services:
  router:
    # Private image based on Traefik
    image: registry.gitlab.com/company/traffic:1.1.5
    restart: always
    volumes:
      - router:/data
      - /var/run/docker.sock:/var/run/docker.sock:ro
    labels:
      - traefik.http.routers.api.rule=Host(`router.sentry.company.net`)
    ports:
      - "80:80"
      - "443:443"

  nginx:
    labels:
      - traefik.enable=true
      - traefik.docker.network=company
      - traefik.http.routers.sentry.rule=Host(`sentry.company.net`)
      - traefik.http.routers.sentry.tls.certresolver=letsencrypt

networks:
  default:
    name: company

volumes:
  router: ~ 
```

The first install works like a charm, but the script crash on update while removing the network:

```
Docker images built.
Removing network company
error while removing network: network company id c9ac9f1a4ed7c1776415025704a5a2647dbe830c4c117c9d0b59ee902ee42c87 has active endpoints
An error occurred, caught SIGERR on line 227
Cleaning up...
```

Relaunching the script "solve" the issue.

BYK:
Maybe we can add a retry loop here that tries this a couple of times?

soullivaneuh:
> Maybe we can add a retry loop here that tries this a couple of times?

Looks a bit dirty, don't you think?

BYK:
>Looks a bit dirty, don't you think?

It is, yes but at least it alleviates the immediate issue without too much refactoring. After this we can invest in a more precise solution.

chadwhitacre:
Not seeing a lot of demand for this after a year, closing.

Issue #694, opened 2020-10-06
Let port binding management to docker-compose.override.yml instead of variable usage


soullivaneuh:
## Summary

Port binding of the nginx server is currently defined with a env variable: https://github.com/getsentry/onpremise/blob/ae9378862caead70b620bb347bf5510833889478/docker-compose.yml#L199-L200

Instead, we should remove the port binding completely and let the user set it up on `docker-compose.override.yml` like this:

```yaml
version: '3.4'

services:
  nginx:
    ports:
      - 80 # Do whatever you want here.
```

## Motivation

Because of a [lack of functionnality of docker-compose](https://github.com/docker/compose/issues/3729) we are currently not able to **remove** the port binding.

As you [suggested on your documentation](https://github.com/getsentry/onpremise#securing-sentry-with-ssltls) I would like to setup a reverse proxy by adding a Traefik router service on the docker stack.

In that case, I do not want to expose the nginx port directly for security reason and because it will become totally useless.

I technically can not do that expect if I modify your `docker-compose.yml` directly, but this is not the proper way and I will always have a git diff to stash at each update.

## Additional Context

Any other context or screenshots or API request payload/responses that
pertain to the feature.

soullivaneuh:
Here is my working `docker-compose.override.yml` with Traefik setup:

```yaml
version: '3.4'

services:
  router:
    # Private image based on Traefik
    image: registry.gitlab.com/company/traffic:1.1.5
    restart: always
    volumes:
      - router:/data
      - /var/run/docker.sock:/var/run/docker.sock:ro
    labels:
      - traefik.http.routers.api.rule=Host(`router.sentry.company.net`)
    ports:
      - "80:80"
      - "443:443"

  nginx:
    labels:
      - traefik.enable=true
      - traefik.docker.network=company
      - traefik.http.routers.sentry.rule=Host(`sentry.company.net`)
      - traefik.http.routers.sentry.tls.certresolver=letsencrypt

networks:
  default:
    name: company

volumes:
  router: ~ 
```

With this config, I can access Sentry on https://sentry.company.net but also on http://sentry.company.net:9000 that I don't want to.

BYK:
Hmm, this is actually a good solution. I think the best non-breaking way forward would be to handle this in the install script: detect if there is an already existing `docker-compose.override.yml` file and if not, create one with the current port binding.

[...]

PR #766, opened 2020-12-09
Integrate with MaxMind out of the box

### LINKED ISSUE #310: GEOIP_PATH_MMDB support
How should one configure GEOIP_PATH_MMDB for the docker image?

chadwhitacre:
Closes #310.

### Todo

- [x] figure out which mmdb we want
- [x] test with relay
- [x] think through configuration - envvar burden okay? expose frequency?
- [x] think through startup/install scenario
- [x] docs: https://github.com/getsentry/develop/pull/219
  - [x] https://develop.sentry.dev/self-hosted/geolocation/
  - [x] https://develop.sentry.dev/self-hosted/#configuration
- [x] use bind mount instead of volume mount
- [x] tighten up nginx config
- [x] trick relay in empty case
- [x] sort out reload story
- [x] legal?
- [x] ~~CI?~~ — #768


chadwhitacre:
https://github.com/maxmind/geoipupdate/blob/master/docker/entry.sh for reference

docker-compose.yml
```diff
@@ -127,6 +128,20 @@ services:
       # If you have high volume and your search return incomplete results
       # You might want to change this to a higher value (and ensure your host has enough memory)
       MAX_MEMORY_USAGE_RATIO: 0.3
+  geoipupdate:
+    << : *restart_policy
+    image: maxmindinc/geoipupdate
+    environment:
+      GEOIPUPDATE_FREQUENCY: 168  # one week, in hours
+      # Set the following (e.g., via docker-compose.override.yml) to enable IP geolocation.
```
BYK, line 136: With the current definition, they can pass these as environment variables too. `docker-compose` will inherit the value set in the environment already.

docker-compose.yml
```diff
@@ -252,6 +267,8 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
+  sentry-geoip:
+    external: true
```
BYK, line 271: ~~Should this really be external? It is data we can easily recover so I'd make this a "local" volume.~~

We shouldn't even need this. I expect the downloader to put the database in a local folder on the host which we can set. So I'd add `geoip` directory to `.gitignore`, download the DB there and mount that directory to the containers.

Makes sense?

nginx/nginx.conf
```diff
@@ -44,8 +44,8 @@ http {
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';
 	proxy_set_header Host $host;
-	proxy_set_header X-Real-IP $remote_addr;
-	proxy_set_header X-Forwarded-For $remote_addr;
+	proxy_set_header X-Real-IP $proxy_add_x_forwarded_for;
+	proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
BYK, line 48: This needs @mattrobenolt check.

nginx/nginx.conf
```diff
@@ -44,8 +44,8 @@ http {
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';
 	proxy_set_header Host $host;
-	proxy_set_header X-Real-IP $remote_addr;
-	proxy_set_header X-Forwarded-For $remote_addr;
+	proxy_set_header X-Real-IP $proxy_add_x_forwarded_for;
```
mattrobenolt, line 47: X-Real-IP would still be `$remote_addr`.

But in theory, we don't have enough information really to assume anything here. But you don't have a list of IPs in `X-Real-IP`.

In practice, you'd wanna use this: https://nginx.org/en/docs/http/ngx_http_realip_module.html

And pass along the correct IPs.

docker-compose.yml
chadwhitacre: Six of one, half-dozen of the other. I can do this, sure.

docker-compose.yml
chadwhitacre: Yup. Hence the "e.g." ... override.yml is one way to do it. I think it's the way we should recommend because otherwise people have to come up with some other means of managing envvars, which is complex and highly idiosyncratic.

docker-compose.yml
BYK: Not sure what you mean by six of one. The reason I prefer a local folder mount is because it is easier to understand, easier to manage, and is more portable. What do you think?

docker-compose.yml
BYK: I think for these kind of "secret"s, it is better to pass these in via env variables? Also we have not advertised `docker-compose.override.yml` yet. I know you are going to add that to the docs, which is great, but I think we need some homework to do before that.

See #694, or #698 for instance.

docker-compose.yml
chadwhitacre: Sorry ... I mean I don't see a material difference between a bind mount and a volume mount here. Sure, it's a little more complicated to access the mmdb file from the host if it's in a volume, but it'll very rarely be needed from the host. It's not obvious to me how a bind mound improves manageability or portability.

Done in 7d2c36be2c0117f1959c494cfc0482195082aeff.

P.S. Why do we use the long syntax for read-only bind mounts, but the short syntax for read/write bind mounts and named volumes?

docker-compose.yml
chadwhitacre: Remove reference to `docker-compose.override.yml` in f47e60c0b7c26ae92360c567944041d1eaea2695.

nginx/nginx.conf
chadwhitacre: Ftr I started with [this](https://github.com/getsentry/onpremise/issues/554#issuecomment-660522070). I don't `ag real-ip` anywhere in `sentry`. Rather, what I'm seeing is that `sentry` [logs](https://github.com/getsentry/sentry/blob/20.11.1/src/sentry/middleware/auth.py#L35) `request.META["REMOTE_ADDR"]`—which is [overriden](https://github.com/getsentry/sentry/blob/20.11.1/src/sentry/middleware/proxy.py#L97-L107) with `X-Forwarded-For` [by](https://github.com/getsentry/sentry/blob/20.11.1/src/sentry/middleware/proxy.py#L84) [default](https://github.com/getsentry/sentry/blob/20.11.1/src/sentry/conf/server.py#L1625)—and that's [where geo is knit in](https://github.com/getsentry/sentry/blob/20.11.1/src/sentry/models/userip.py#L37-L44). (`relay`, of course, geolocates the IP address provided in the request [body](https://docs.sentry.io/platforms/python/enriching-events/identify-user/), not the headers.)

Tweaked in a1716ebb2b8fd21c0ece58f3a0ac0989be398612.

docker-compose.yml
BYK: >Sure, it's a little more complicated to access the mmdb file from the host if it's in a volume, but it'll very rarely be needed from the host. It's not obvious to me how a bind mound improves manageability or portability.

It is tricky to backup, move etc a named volume as Docker hides them away. This is especially an issue on non-Linux systems where things happen through a VM. I can easily imagine people copying their mmdb files around once they have a license once. External volumes also need managing from the install script (compose project scoped volumes don't have this issue but they are still opaque)

>P.S. Why do we use the long syntax for read-only bind mounts, but the short syntax for read/write bind mounts and named volumes?

Because we cannot specify the `readonly` flag in the short syntax? (or if we can, I don't know how to)

chadwhitacre: Okay, okay. Bind mount is easier to work with here, I agree. 😁 

Re: syntax ... as I suspected. 😌 The short syntax does [allow](https://docs.docker.com/compose/compose-file/#short-syntax-3) for an optional mode at the end, like so:

`./foo:/foo:ro`

BYK: >Okay, okay. Bind mount is easier to work with here, I agree. 😁

Hahaha, if you don't, we can still try. It is hard to be very objective about these things. 😁 

>Re: syntax ... as I suspected. 😌 The short syntax does allow for an optional mode at the end, like so:

Then I'm all in for making these short and consistent in a separate patch 🥳

BYK:
Looks pretty good so far, great job!

chadwhitacre:
Greetings! 👋 My name is ChadBot 🤖, and I will be performing your manual continuous integration services for this PR. 🐭

### Commit

c53d8a9f7b323f0ebf55860cc9011e8ee18f5095

### Test Matrix

| conf | mmdb | expected |
|:----:|:-------:|:----------:|
| missing | missing | error |
| missing | present | error |
| bad | present | error |
| good | present | done |
| bad | missing | error |
| good | missing | done |

### Test Run

```
$ ./install/geoip.sh
Setting up IP address geolocation ...
Installing (empty) IP address geolocation database ... done.
IP address geolocation is not configured for updates.
See https://develop.sentry.dev/self-hosted/geolocation/ for instructions.
Error setting up IP address geolocation.
$
$
$ ./install/geoip.sh
Setting up IP address geolocation ...
IP address geolocation database already exists.
IP address geolocation is not configured for updates.
See https://develop.sentry.dev/self-hosted/geolocation/ for instructions.
Error setting up IP address geolocation.
$
$
$ touch geoip/GeoIP.conf
$
$ ./install/geoip.sh
Setting up IP address geolocation ...
IP address geolocation database already exists.
IP address geolocation is configured for updates.
Updating IP address geolocation database ...
Creating sentry_onpremise_geoipupdate_run ...
Creating sentry_onpremise_geoipupdate_run ... done
error loading configuration file /sentry/GeoIP.conf: the `EditionIDs` option is required
Error updating IP address geolocation database.
Error setting up IP address geolocation.
$
$
$ cp ~/Desktop/GeoIP.conf geoip 
$
$ ./install/geoip.sh           
Setting up IP address geolocation ...
IP address geolocation database already exists.
IP address geolocation is configured for updates.
Updating IP address geolocation database ... 
Creating sentry_onpremise_geoipupdate_run ... 
Creating sentry_onpremise_geoipupdate_run ... done
Done updating IP address geolocation database.
Done setting up IP address geolocation.
$
$
$ rm geoip/GeoIP.conf
$ rm geoip/GeoLite2-City.mmdb
$ touch geoip/GeoIP.conf
$
$ ./install/geoip.sh
Setting up IP address geolocation ...
Installing (empty) IP address geolocation database ... done.
IP address geolocation is configured for updates.
Updating IP address geolocation database ...
Creating sentry_onpremise_geoipupdate_run ...
Creating sentry_onpremise_geoipupdate_run ... done
error loading configuration file /sentry/GeoIP.conf: the `EditionIDs` option is required
Error updating IP address geolocation database.
Error setting up IP address geolocation.
$
$
$ rm geoip/GeoIP.conf 
$ rm geoip/GeoLite2-City.mmdb
$ cp ~/Desktop/GeoIP.conf geoip 
$
$ ./install/geoip.sh 
Setting up IP address geolocation ...
Installing (empty) IP address geolocation database ... done.
IP address geolocation is configured for updates.
Updating IP address geolocation database ... 
Creating sentry_onpremise_geoipupdate_run ... 
Creating sentry_onpremise_geoipupdate_run ... done
Done updating IP address geolocation database.
Done setting up IP address geolocation.
$
$
$ ls -l geoip 
total 65864
-rw-r--r--  1 chadwhitacre  staff        70 Dec 11 10:02 GeoIP.conf
-rw-r--r--  1 chadwhitacre  staff  66456088 Dec 11 10:02 GeoLite2-City.mmdb
-rw-r--r--  1 chadwhitacre  staff      1055 Dec 10 11:47 GeoLite2-City.mmdb.empty
$
```

### Result

💃

README.md
```diff
@@ -72,3 +72,19 @@ For more information regarding updating your Sentry installation, please visit [
...
+| [`memcached`](https://hub.docker.com/_/memcached) | [Danga Interactive, Inc.](https://www.danga.com/) | [3-clause BSD](https://github.com/memcached/memcached/blob/master/LICENSE) |
+| [`nginx`](https://hub.docker.com/_/nginx) | [Nginx, Inc.](https://www.nginx.com/) | [2-clause BSD](http://nginx.org/LICENSE) |
+| [`postgres`](https://hub.docker.com/_/postgres) | [The PostgreSQL Global Development Group](https://www.postgresql.org/) | [PostgreSQL](https://www.postgresql.org/about/licence/) |
+| [`redis`](https://hub.docker.com/_/redis) | [Salvatore Sanfilippo and contributors](https://github.com/redis/redis/graphs/contributors) | [3-clause BSD](https://redis.io/topics/license) |
+| [`tianon/exim4`](https://hub.docker.com/r/tianon/exim4) | [Tianon Gravi](https://github.com/tianon) | [MIT](https://github.com/tianon/dockerfiles/blob/master/LICENSE) |
+| [`yandex/clickhouse-server`](https://hub.docker.com/r/yandex/clickhouse-server) | [Yandex LLC](https://yandex.com/company/) | [Apache 2](https://github.com/ClickHouse/ClickHouse/blob/master/LICENSE) |
```
BYK, line 90: I don't think we need this and it is a hassle to maintain. I'd rather have some automatic checks for licenses if we need to.

That said as long as we are not deriving new images from these, any license should be fine.

geoip/.gitignore
```diff
@@ -0,0 +1,3 @@
+GeoIP.conf
```
BYK, line 1: Since we now have an empty file in `geoip` directory, can we move these definitions to the `.gitignore` at the root level?

install.sh
```diff
@@ -325,6 +325,10 @@ if [[ ! -f "$RELAY_CREDENTIALS_JSON" ]]; then
   echo "Relay credentials written to $RELAY_CREDENTIALS_JSON"
 fi
 
+
+./install/geoip.sh
```
BYK, line 329: I know install.sh is getting a bit unwieldy but does splitting geoip part only help much?

install/geoip.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+if [ ! -f 'install.sh' ]; then echo 'Where are you?'; exit 1; fi
+
+dc="docker-compose --no-ansi"
+dcr="$dc run --rm"
```
BYK, line 6: These all suggest me that this file should not be a separate module. Or this file needs to be `source`d. I still prefer things being in a single place until we have a very good reason tho.

relay/config.example.yml
```diff
@@ -10,3 +10,4 @@ processing:
     - {name: "bootstrap.servers", value: "kafka:9092"}
     - {name: "message.max.bytes", value: 50000000} #50MB or bust
   redis: redis://redis:6379
+  geoip_path: "/geoip/GeoLite2-City.mmdb"
```
BYK, line 13: If we point these to a specific file, why do we need the `.empty` file?

BYK (changes requested):

install/geoip.sh
chadwhitacre: "These all" ... there are only two. :P

I agree there is more refactoring to be done, I didn't want to load this PR up too much or block it on a bigger refactor.

install.sh
chadwhitacre: Yes. By splitting out the geoip part I am able to [test it in isolation](https://github.com/getsentry/onpremise/pull/766#issuecomment-743247145).

geoip/.gitignore
chadwhitacre: Done in 1364eaf79cab9150e64b91422434a9bf9a0e39be.

relay/config.example.yml
chadwhitacre: 1. It's simpler for our users (operators) if we set a path in the default `*.example.*` config files, so that they don't have to set the path themselves in order to turn on ip geolocation. This is especially true since there are multiple similarly named files/dbs that MaxMind provides, which is a source of confusion/error.

1. Relay (and maybe Sentry?) hard-fail on startup if the file pointed to by `geoip_path` doesn't exist. So to satisfy (1) we need to have _some_ file there.

1. We want `geoip/GeoLite2-City.mmdb` itself to be gitignored, because we want to continually update it in production without cluttering git status.

1. By shipping an empty mmdb file and copying it into place if otherwise non-existent during install.sh, we keep git status clean (3) while allowing relay to start (2) with minimal confusion for users (1).

README.md
chadwhitacre: Legal confirmed that referring to third-party images in Dockerfiles (and presumably also docker-compose.ymls) does not introduce a compliance burden. I force-pushed off 6a41f11.

install/geoip.sh
BYK: Let's do something like the following:

- https://stackoverflow.com/a/42101141/90297
- https://stackoverflow.com/a/10822868/90297
- https://stackoverflow.com/a/36830054/90297

BYK (changes requested):

docker-compose.yml
```diff
@@ -126,6 +127,11 @@ services:
       # If you have high volume and your search return incomplete results
       # You might want to change this to a higher value (and ensure your host has enough memory)
       MAX_MEMORY_USAGE_RATIO: 0.3
+  geoipupdate:
+    image: 'maxmindinc/geoipupdate:latest'
+    entrypoint: ['/usr/bin/geoipupdate', '-d', '/sentry', '-f', '/sentry/GeoIP.conf']
```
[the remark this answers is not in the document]
chadwhitacre: We're overriding the entrypoint to avoid the [envvar shenanigans](https://github.com/maxmind/geoipupdate/blob/v4.5.0/docker/entry.sh#L9-L20), not to change the location of the mmdb file.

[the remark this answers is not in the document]
chadwhitacre: By default they look for the conf file and mmdb file in different directories, so we need to do _something_ to wire up between the host and the geoipupdate container. This seems a fine way imo.

install.sh
chadwhitacre: I don't either. :) I think we should refactor install.sh to make it more modular and therefore more testable ... and we should test it! 😁 I don't see this bloating the root director with many `.sh` files, you can see here I split the subscript into an `install/` directory. Yes we can and will share components between these scripts ... again, one step at a time.

BYK (approved):

install/geoip.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo -n "Installing (empty) IP address geolocation database ... "
+    cp "$mmdb.empty" "$mmdb"
+    echo "done."
+  else
+    echo "IP address geolocation database already exists."
+  fi
+
+  if [[ ! -f "$conf" ]]; then
+    echo "IP address geolocation is not configured for updates."
+    echo "See https://develop.sentry.dev/self-hosted/geolocation/ for instructions."
```
BYK, line 25: Just thought about this, maybe we can try something like https://unix.stackexchange.com/a/437581/108960?

chadwhitacre: Would we try to sniff the terminal in use? Terminal.app is noticeably absent from the list of supported terminals. :-/

BYK: Ehh, then forget it. Not worth the effort I'd say.

PR #775, opened 2020-12-15
ci(release): Move to getsentry/publish for releases

BYK:
A copy of getsentry/sentry#22657 with fixes included.

That said this one is a bit different. We used to use the extra option `no-merge` when publishing to keep the release branches and also keep master on nightly versions. If we want to keep this, we need to add per-project overrides to getsentry/publish which would increase complexity at this early stage for this fringe case. Instead, I opted to follow what getsentry/sentry does: merge but then immediately after that update the version.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,4 +1,4 @@
-minVersion: "0.10.0"
+minVersion: "0.14.0"
 github:
   owner: getsentry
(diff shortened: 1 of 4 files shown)


chadwhitacre (approved):
Skimmed, looks reasonable, have not tested

PR #776, opened 2020-12-15
Drop to daily locking now that we're caught up

chadwhitacre:


BYK (approved):
Why 3:11 and not something else? 😀

chadwhitacre:
No reason. :)

Figured random was better than not.

PR #778, opened 2020-12-17
meta: Pin lock-threads, crank frequency to test

chadwhitacre:
Per https://github.com/getsentry/sentry/pull/22754#discussion_r544690140.


BYK (approved):
Yezz

PR #779, opened 2020-12-17
Futz with action pin syntax

chadwhitacre:
[Failing](https://github.com/getsentry/onpremise/runs/1571980788?check_suite_focus=true#step:1:14) with:

> Download action repository 'dessant/lock-threads@63786a6c74ee3cfc4584f36de4360305c55e5127'
Error: An action could not be found at the URI 'https://api.github.com/repos/dessant/lock-threads/tarball/63786a6c74ee3cfc4584f36de4360305c55e5127'

Will this convince actions to parse as SHA?

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -6,7 +6,7 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@63786a6c74ee3cfc4584f36de4360305c55e5127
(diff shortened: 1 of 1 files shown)


BYK (approved):
YOLO

chadwhitacre:
Ftr I had a corrupt SHA in https://github.com/getsentry/onpremise/pull/778.

Correct: https://github.com/dessant/lock-threads/commit/63786a6c74ee3cfc4584f36de4360305c55e5126

PR #784, opened 2020-12-23
meta: Install stalebot

chadwhitacre:
https://app.asana.com/0/1199602781063150/1198192573081727

- Labels un-triaged issues with `Status: Stale` after three weeks of inactivity.
- Closes issues so labeled after one more week of inactivity.
- Use `Status: Accepted` to hold open indefinitely.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -14,20 +14,34 @@ jobs:
           days-before-close: 7
           only-labels: ""
           operations-per-run: 100
-          remove-stale-when-updated: false
+          remove-stale-when-updated: true
           debug-only: false
           ascending: false
 
-          exempt-issue-labels: "Triage: Accepted"
-          stale-issue-label: "Triage: Stale"
-          stale-issue-message: "This message is required, but you shouldn't see it."
-          skip-stale-issue-message: true
+          exempt-issue-labels: "Status: Accepted"
+          stale-issue-label: "Status: Stale"
+          stale-issue-message: |-
+            This issue has gone three weeks without activity. In another week, I will close it.
+
+            But! If you comment or otherwise update it, I will reset the clock, and if you label it `State: Accepted`, I will leave it alone entirely.
+
+            ----
+
+            "A weed is but an unloved flower." ― _Ella Wheeler Wilcox_ 🥀
+          skip-stale-issue-message: false
           close-issue-label: ""
           close-issue-message: ""
 
-          exempt-pr-labels: "Triage: Accepted"
-          stale-pr-label: "Triage: Stale"
-          stale-pr-message: "This message is required, but you shouldn't see it."
-          skip-stale-pr-message: true
+          exempt-pr-labels: "Status: Accepted"
+          stale-pr-label: "Status: Stale"
+          stale-pr-message: |-
+            This pull request has gone three weeks without activity. In another week, I will close it.
+
+            But! If you comment or otherwise update it, I will reset the clock, and if you label it `State: Accepted`, I will leave it alone entirely.
+
+            ----
+
+            "A weed is but an unloved flower." ― _Ella Wheeler Wilcox_ 🥀
+          skip-stale-pr-message: false
           close-pr-label:
           close-pr-message: ""
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -23,7 +23,7 @@ jobs:
           stale-issue-message: |-
             This issue has gone three weeks without activity. In another week, I will close it.
 
-            But! If you comment or otherwise update it, I will reset the clock, and if you label it `State: Accepted`, I will leave it alone entirely.
+            But! If you comment or otherwise update it, I will reset the clock, and if you label it `Status: Accepted`, I will leave it alone entirely.
 
             ----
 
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
+name: 'close stale issues/PRs'
+on:
+  schedule:
+    - cron: '* */6 * * *'
+  workflow_dispatch:
```
BYK, line 5: Do we need this trigger?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@87c2b794b9b47a9bec68ae03c01aeb572ffebdb1
+        with:
+          repo-token: ${{ github.token }}
+          days-before-stale: 21
+          days-before-close: 7
+          only-labels: ""
+          operations-per-run: 100
+          remove-stale-when-updated: false
```
BYK, line 17: Shall we set this to true?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+          debug-only: false
+          ascending: false
+
+          exempt-issue-labels: "Triage: Accepted"
+          stale-issue-label: "Triage: Stale"
+          stale-issue-message: "This message is required, but you shouldn't see it."
+          skip-stale-issue-message: true
+          close-issue-label: ""
+          close-issue-message: ""
```
BYK, line 26: Shouldn't we have some sort of a message to explain what happened?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+
+          exempt-pr-labels: "Triage: Accepted"
```
BYK, line 28: I think for PRs we may also use WIP?

BYK (changes requested):

.github/workflows/stale.yml
chadwhitacre: It's helpful for testing or manual kick-off. Resolving unless you think it's harmful to have (footgun?).

.github/workflows/stale.yml
chadwhitacre: I see `wip` in [`sentry-docs`](https://github.com/getsentry/sentry-docs/labels) but not in [`sentry`](https://github.com/getsentry/sentry/labels) or [`onpremise`](https://github.com/getsentry/onpremise/labels). On the one hand, `wip` labels are obsolete since GitHub implemented [Draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/). On the other, I would think that draft/WIP PRs are precisely what we _want_ to prune if they go stale.

chadwhitacre: On the third hand(?), `sentry` has a `Status: Accepted` label, so I'm going to switch to that.

.github/workflows/stale.yml
chadwhitacre: Good catch. I think I read this as `remove-stale-when-closed`. Confirmed that's [not the behavior](https://github.com/actions/stale/blob/87c2b794b9b47a9bec68ae03c01aeb572ffebdb1/src/IssueProcessor.ts#L235-L241).

.github/workflows/stale.yml
chadwhitacre: I was thinking the default message GitHub adds would be enough to start with, but I agree a friendlier comment is a better user experience. Here's what I came up with:

Screen Shot 2021-01-04 at 3 52 35 PM

.github/workflows/stale.yml
```diff
@@ -14,20 +14,34 @@ jobs:
...
-          exempt-issue-labels: "Triage: Accepted"
-          stale-issue-label: "Triage: Stale"
-          stale-issue-message: "This message is required, but you shouldn't see it."
-          skip-stale-issue-message: true
+          exempt-issue-labels: "Status: Accepted"
+          stale-issue-label: "Status: Stale"
+          stale-issue-message: |-
+            This issue has gone three weeks without activity. In another week, I will close it.
+
+            But! If you comment or otherwise update it, I will reset the clock, and if you label it `State: Accepted`, I will leave it alone entirely.
```
BYK, line 26:
```suggestion
            But! If you comment or otherwise update it, I will reset the clock, and if you label it `Status: Accepted`, I will leave it alone forever.
```

Let's be more dramatic 😛

.github/workflows/stale.yml
```diff
@@ -14,20 +14,34 @@ jobs:
...
-          exempt-pr-labels: "Triage: Accepted"
-          stale-pr-label: "Triage: Stale"
-          stale-pr-message: "This message is required, but you shouldn't see it."
-          skip-stale-pr-message: true
+          exempt-pr-labels: "Status: Accepted"
+          stale-pr-label: "Status: Stale"
+          stale-pr-message: |-
+            This pull request has gone three weeks without activity. In another week, I will close it.
+
+            But! If you comment or otherwise update it, I will reset the clock, and if you label it `State: Accepted`, I will leave it alone entirely.
```
BYK, line 40:
```suggestion
            But! If you comment or otherwise update it, I will reset the clock, and if you label it `Status: Accepted`, I will leave it alone forever.
```

.github/workflows/stale.yml
BYK: Ehh, not a fan but I don't have a great argument to remove it so YOLO

BYK (approved):
Don't forget to apply the `State` -> `Status` fixes in your awesome stale message. Otherwise looks pretty great :)

PR #802, opened 2021-01-07
Fix forked lockbot pinning one more time

chadwhitacre:


chadwhitacre:
Tested and working in https://github.com/getsentry/onpremise/runs/1664983458?check_suite_focus=true

BYK (approved):

Issue #787, opened 2021-01-04
Provide better guidelines on self-hosted limits


chadwhitacre:
Our intention with self-hosted is to provide a turnkey solution for small workloads, and to leave the provisioning of Sentry for larger workloads as an exercise for the reader. We should be clearer in our docs about what exactly constitutes a "small" workload.

Immediate impetus is @kanadaj's [comments](https://discord.com/channels/621778831602221064/621783515423440927/792443846721667072):

> Sentry has way too many separately moving parts to be easy to maintain on your own
> onpremise has way too many ways to break on a sufficiently large deployment
> if zookeeper breaks, you basically need a PhD in ZK to fix it
> and probably a review of the ZK config cause iirc CH has some recommended configs for ZK
> I don't use docker-compose or really any of that
> I'm transcribing /onpremise to a kubernetes yml

## +1

1. Am2020 - [forum](https://forum.sentry.io/t/enable-performance-monitoring/12205/5)

max-wittig:
Another solution would be to provide a kubernetes yaml instead, as docker-compose doesn't really work for productive use on it's own

kanadaj:
@max-wittig I've discussed it with Chad, if the docker-compose is rewritten to be using volume mounts instead of custom compiling Dockerfiles to embed the config, it'd be trivial to convert it to kubernetes yaml files, probably even by existing CLI tools assuming they can pick up and embed the required files into ConfigMaps/Secrets.

Even if they can't, it's not hard to automate that.

Another option would be an operator, which is way more overhead to work out, but could help with updates between Sentry versions in the future if a more complex upgrade path is required.

chadwhitacre:
Filing for later ...

> The 2400MB min RAM requirement in the readme for onpremise seems waaaay below what I'd realistically expect to use

https://discord.com/channels/621778831602221064/621783515423440927/795686074851524700

> Intel(R) Xeon(R) Silver 4108 CPU @ 1.80GHz with 4 cores
> 8 GB of ram

https://discord.com/channels/621778831602221064/621783515423440927/795673428287684639

kanadaj:
If it helps any with the numbers, with a freshly wiped ClickHouse, our own Sentry deployment uses 3.74GB RAM atm, highest I've seen was around the 5-6GB mark, ingesting ~50-100 errors and 5400 transactions an hour, though I expect this to go way up once we have integration for other languages.

chadwhitacre:
[...]

PR #803, opened 2021-01-08
Bump RAM requirement in install.sh

chadwhitacre:
In #787 we decided to up the minimum RAM requirement from 2.4 to 8 GB, handled the README in 2dadbda1cc44e5a9a1c1a84f6dd0306258f17a47, now for `install.sh` ... where it really matters. :)

h/t @AlexanderRydberg

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -18,7 +18,8 @@ exec &> >(tee -a "$log_file")
 
 MIN_DOCKER_VERSION='19.03.6'
 MIN_COMPOSE_VERSION='1.24.1'
-MIN_RAM=8000 # MB
+MIN_RAM_HARD=4000 # MB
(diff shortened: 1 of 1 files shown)


BYK:
Let's go with 4G as it should be enough for smaller instances. It says _minimum_ and we can always have _recommended_ settings. That can be a soft warning during install?

BYK (changes requested):
See my note earlier.

chadwhitacre:
Dialed back to just a comment for now to minimize disruption.

chadwhitacre:
Pool is closed but warning works.

BYK (approved):
Thought for future: should we send these error and warning messages to stderr instead of stout?

PR #822, opened 2021-01-19
Remove injection vector

chadwhitacre:


.github/workflows/validate-new-issue.yml
```diff
@@ -22,7 +23,7 @@ jobs:
         fi
 
         # Look for a template where the headings match this issue's
-        echo "${{ github.event.issue.body }}" > issue-body
+        echo "$BODY" > issue-body
```
BYK, line 26: Are we sure this is safe enough? I just discovered that github does something like this:
https://github.com/actions/toolkit/blob/85f6235ca99674b76ff1dbde0683a5e9df71e1d2/packages/github/src/context.ts#L26-L31

Which means the event payload is saved on disk so that we can extract it safely by using `jq`?

BYK (changes requested):
I haz question and a proposal

.github/workflows/validate-new-issue.yml
chadwhitacre: This was GHLS's suggestion, and I did do some testing around it:

```bash
$ BODY='"; echo "bar"; echo "blah'
$ echo "$BODY" 
"; echo "bar"; echo "blah
$


```

chadwhitacre: Here's a screenshot showing that the workflow script is not tampered with in this case:

Screen Shot 2021-01-19 at 11 49 45 AM

Since we're dumping to disk I believe we're in the same class of fix as you point to.

BYK: >Since we're dumping to disk I believe we're in the same class of fix as you point to.

Not exactly as we dump to disk using `echo` and using the shell. Anyway I'll accept to unblock and see if I can use the one on the disk directly in a follow-up.

BYK (approved):

.github/workflows/validate-new-issue.yml
chadwhitacre: > the event payload is saved on disk so that we can extract it safely by using `jq`

Good find!

Screen Shot 2021-01-19 at 12 11 45 PM

I agree this is a better approach, I _think_ we're safe with `echo "$BODY"` but extracting from the file already on disk feels much safer. :)

chadwhitacre: Done in 4dca8e4fa2494342e01a95ad0b253f6bc1d88ae9.

.github/workflows/validate-new-issue.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         fi
 
         # Look for a template where the headings match this issue's
-        echo "${{ github.event.issue.body }}" > issue-body
+        cat "$GITHUB_EVENT_PATH" | jq -r .issue.body > issue-body
```
BYK, line 25:
```suggestion
        jq -r .issue.body "$GITHUB_EVENT_PATH" > issue-body
```

BYK (approved):

PR #824, opened 2021-01-19
Explore pool closures

chadwhitacre:


BYK:
This shouldn't be about memory. It is a bug in the `docker-py` library. I'll submit a fix but until then, simply increasing the pool size until we don't have issues should be the fix.

PR #825, opened 2021-01-19
ci(test): Even larger pools

### LINKED ISSUE #823: ci(test): CI test regularly dies with "Pool is closed"
Attempted to fix in #768, etc. Still not resolved.

BYK:
Attempt to fix #823

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
 
       - name: Install and test
         env:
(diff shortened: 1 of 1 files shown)


chadwhitacre (approved):
This feels like whack-a-mole to me. We should get to the root cause at some point.

BYK:
@chadwhitacre I think I documented the root cause in #819. Even if we fix that, we'll need to wait till `docker-py` gets released and that release makes into `docker-compose`

PR #827, opened 2021-01-19
Add some structure to logging with ::group::

chadwhitacre:
GHA provides nice UI using [this syntax](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines), which isn't _too_ ugly in raw logs.

(Baby step towards https://github.com/getsentry/onpremise/issues/795, btw.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,9 +35,10 @@ jobs:
         env:
           SENTRY_PYTHON2: ${{ matrix.py2 == '1' || '' }}
         run: |
+          echo "Testing clobber-y upgrade"
           ./install.sh
           ./test.sh
-          printf "Testing in-place upgrade"
+          echo "Testing in-place upgrade"
           ./install.sh --minimize-downtime
           ./test.sh
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
         env:
           SENTRY_PYTHON2: ${{ matrix.py2 == '1' || '' }}
         run: |
-          echo "Testing clobber-y upgrade"
+          echo "Testing initial install"
           ./install.sh
           ./test.sh
           echo "Testing in-place upgrade"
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -35,9 +35,10 @@ jobs:
         env:
           SENTRY_PYTHON2: ${{ matrix.py2 == '1' || '' }}
         run: |
+          echo "Testing clobber-y upgrade"
```
BYK, line 38: This is the initial install, not an upgrade 😀

install.sh
```diff
@@ -6,15 +6,16 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+# Thanks to https://unix.stackexchange.com/a/145654/108960
+log_file="sentry_install_log-`date +'%Y-%m-%d_%H-%M-%S'`.txt"
+exec &> >(tee -a "$log_file")
+
+echo "::group::Defining variables and helpers ..."
```
BYK, line 13: This is nice in GitHub Actions but it will clobber our logs and other people's terminals during install. Maybe detect GHA and limit this there?

BYK (changes requested):
As much as I love the tidy output, this tailors the `install.sh` script for our CI rather than the users it is intended for.

.github/workflows/test.yml
chadwhitacre: Addressed in 8f39262d443266d043ae0e61f6b77aa26e2b3f20.

install.sh
chadwhitacre: Done in de6bcf018ab6e3fab0ea5ff837c74d3d77c804fe for `install.sh`. Leaving `test.sh` groups as-is since that is CI-centric.

chadwhitacre:
@BYK Sorry for the confusion, thought we were on the same page. Varied between CI/local in the latest. Here's what local looks like:

Screen Shot 2021-01-19 at 5 20 18 PM

chadwhitacre:
[FTR](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables):

> `GITHUB_ACTIONS` | Always set to `true` when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.
-- | --

BYK (approved):
Hah, quite clever! Also my apologies as it was my fault for not reading your message properly 🙁 

Thanks for the quick update!

PR #835, opened 2021-01-24
fix(nginx): Remove X-Real-IP header entry

Mixaill:
Removes the obsolete and confusing `X-Real-IP` header setting from the Nginx config.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -44,7 +44,6 @@ http {
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';
 	proxy_set_header Host $host;
-	proxy_set_header X-Real-IP "";
 	proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 	proxy_set_header X-Forwarded-Proto $scheme;
 	proxy_set_header X-Request-Id $request_id;
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -44,7 +44,7 @@ http {
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';
 	proxy_set_header Host $host;
-	proxy_set_header X-Real-IP $remote_addr;
+	proxy_set_header X-Real-IP "";
 	proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 	proxy_set_header X-Forwarded-Proto $scheme;
 	proxy_set_header X-Request-Id $request_id;


BYK (commented):
This makes sense that said I wonder whether we should drop `X-Real-IP` as `X-Forwarded-For` should have all previous hops, including the original client IP.

@mattrobenolt, thoughts?

mattrobenolt:
This is incorrect behavior and we shouldn't allow this.

With that said, we should stop setting X-Real-IP since we don't know if this is the outer most proxy. That's what XFF is for. But this also needs to be configured correctly and safely. We are very naive about this in Sentry, so in theory it can easily be spoofed: https://github.com/getsentry/sentry/blob/master/src/sentry/middleware/proxy.py#L105 Sentry expects this header to already be sanitized by something like this: https://nginx.org/en/docs/http/ngx_http_realip_module.html

I'd highly suggest doing this, but it's going to need to be configured per-user, etc. Since it's unique to your setup and what networks you trust. We might be able to have some sensible defaults by populating with all private network ranges or something, I dunno.

BYK:
@mattrobenolt we don't even seem to be using `X-Real-IP` in Sentry: https://github.com/getsentry/sentry/commit/e8d3126a4dfc4f968edfeb4f5362ebbbace52065

I'm a bit confuse. Is it Django that picks this up automatically?

BYK (changes requested):
@Mixaill thanks a lot for your PR that said looks like not setting `X-Real-IP` seems like the better path forward. Can you check if it works for you too?

mattrobenolt:
Nothing picks it up. We don’t use it. It effectively does nothing here. It’s just “wrong”.

Mixaill:
Well, looks like a messed a bit with my setup and tested two things at once (`option forwardfor` on HAProxy side and `X-Real-IP` on NGinx side)

Then, I think, we can delete it to elimite confusion, it
* does not used by sentry
* has invalid behavior

nginx/nginx.conf
```diff
@@ -44,7 +44,7 @@ http {
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';
 	proxy_set_header Host $host;
-	proxy_set_header X-Real-IP $remote_addr;
+	proxy_set_header X-Real-IP "";
```
BYK, line 47:
```suggestion
```

Shouldn't we just remove this line?

Mixaill: Removing this line leaves header untouched if it already exists. 
In this case there is probability that header will be accidentally used by backend in future.

BYK (approved):

BYK:
Thanks a lot @Mixaill!

mattrobenolt:
We shouldn't go out of our way to remove it just because we may potentially use it in the future, which, we won't. Otherwise, that's suggesting we delete every header we don't currently use. We just should ignore it.

PR #840, opened 2021-01-29
Change MIN_RAM_HARD from 4000 to 3800

### LINKED ISSUE #839: Maybe lower MIN_RAM_HARD from 4000 to 3800?
How much memory does sentry really need?
With a hard limit of 4000 I can't even install it on a machine with 4gb memory because the machine in my case shows just 3888. The next step at most providers would be 8gb, but I think that is not what you intended with 4000?

https://github.com/getsentry/onpremise/blob/54a20969a62675cb3e5b8a976c573f86806a081b/install.sh#L28

arusa:
On machines with 4gb the available memory is often a little bit lower than 4000

closes #839

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -25,7 +25,7 @@ source ./install/docker-aliases.sh
 
 MIN_DOCKER_VERSION='19.03.6'
 MIN_COMPOSE_VERSION='1.24.1'
-MIN_RAM_HARD=4000 # MB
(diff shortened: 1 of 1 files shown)


chadwhitacre (approved):

chadwhitacre:
Thanks for the PR, @arusa! 🙏

PR #841, opened 2021-01-29
fix(config): extended-permissions is on github-login, not app

BYK:
Fixes the issue where we set an invalid option, `github-app.extended-permissions`, instead of the correct one, `github-login.extended-permissions`. Some people mentioned this warning earlier but never clearly enough to point that it was coming from our default settings suggestions.


chadwhitacre (approved):
Taking your word for it, not sure where this coming from or going, but your story seems plausible. :)

PR #844, opened 2021-01-29
ref(requirements): Add min CPU requirement, relax soft RAM

BYK:
Adds minimum of 4 CPU cores requirement as anything below will perform quite poorly even on lower loads. Relaxes the soft RAM requirement from 8000 MB to 7800 MB as even when there is 8 GB RAM installed, the system reserves some of it to itself and under reports the amount.


chadwhitacre (approved):
lgtm

PR #861, opened 2021-02-15
fix(cron): Fix sentry-cleanup entrypoint issue

### LINKED ISSUE #860: sentry-cleanup does not start
### Version Information

Version: *20.12.0*

### Steps to Reproduce

1. Upgrade to the latest version

### Expected Result

I expect all docker services to start.

### Actual Result

sentry-cleanup-onpremise-local         "/etc/sentry/entrypo…"   4 minutes ago       Restarting (127) 10 seconds ago                                                                                  sentry_onpremise_sentry-cleanup_1

### Logs

```
$ docker logs sentry_onpremise_sentry-cleanup_1 -f
/docker-entrypoint.sh: line 27: exec: 0 0 * * * gosu sentry sentry cleanup --days 90: not found
```

I believe that issue was introduced by
https://github.com/getsentry/onpremise/pull/834#discussion_r575919003

```
  sentry-cleanup:
    <<: *sentry_defaults
    image: sentry-cleanup-onpremise-local
    build:
      context: ./cron
      args:
        BASE_IMAGE: "$SENTRY_IMAGE"
    entrypoint: "/entrypoint.sh" <---- Added line
```

I was able to start failing service

BYK:
Fixes #860 and adds a test case to ensure all cleanup jobs are working.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -219,6 +219,7 @@ services:
       context: ./cron
       args:
         BASE_IMAGE: "$SENTRY_IMAGE"
+    entrypoint: "/entrypoint.sh"
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
   nginx:
     <<: *restart_policy
--- a/test.sh
+++ b/test.sh
@@ -113,3 +113,7 @@ do
   echo "Pass."
 done
 echo "::endgroup::"
+
+echo "::group::Ensure cleanup crons are working ..."
+docker-compose ps | grep -q -- "-cleanup_.\+[[:space:]]\+Up[[:space:]]\+"
+echo "::endgroup::"


test.sh
```diff
@@ -113,3 +113,7 @@ do
   echo "Pass."
 done
 echo "::endgroup::"
+
+echo "::group::Ensure cleanup crons are working ..."
+docker-compose ps | grep -q -- "-cleanup_.\+[[:space:]]\+Up[[:space:]]\+"
+echo "::endgroup::"
```
chadwhitacre, line 119: Yay for test case! 😍

docker-compose.yml
```diff
@@ -219,6 +219,7 @@ services:
       context: ./cron
       args:
         BASE_IMAGE: "$SENTRY_IMAGE"
+    entrypoint: "/entrypoint.sh"
```
chadwhitacre, line 222: Don't we need this for all cron-ish images?

BYK: Nope because we don't override any other `entrypoint`s yet. Once that happens, we'll have a `<<: cron_common` or something. We can add that as a precaution but then we'd need to add a note to the `Dockerfile` that we are overriding the `ENTRYPOINT` command anyway.

chadwhitacre: I.e., also `snuba-cleanup` and `symbolicator-cleanup`.

chadwhitacre (changes requested):
What about other `./cron` images?

BYK:
@chadwhitacre answered here: https://github.com/getsentry/onpremise/pull/861#discussion_r576266036

docker-compose.yml
chadwhitacre: > we don't override any other `entrypoint`s yet

Meaning in `SNUBA_IMAGE` and `SYMBOLICATOR_IMAGE`, yes?

test.sh
chadwhitacre: Appears to pass, even. :)

https://github.com/getsentry/onpremise/runs/1904387743?check_suite_focus=true#step:4:1946

docker-compose.yml
chadwhitacre: I see now, `snuba-cleanup` inherits from `snuba_defaults`/`x-snuba-defaults`, which inherits from `restart_policy`/`x-restart-policy`, which doesn't have the `entrypoint` override that `x-sentry-defaults` has. Similarly with `symbolicator-cleanup`, which inherits from `restart_policy`/`x-restart-policy` directly. 👍

chadwhitacre (approved):

PR #870, opened 2021-02-24
fix(relay): Remove http._client usage

untitaker:
As per https://github.com/getsentry/relay/pull/938 this option no longer
exists. Existing values will be ignored, however.


chadwhitacre (approved):
I'll take your word for it, seems reasonable.

PR #886, opened 2021-03-15
fix(config): Remove deprecated `slack.legacy-app` config from defaults

BYK:


chadwhitacre (approved):
lgtm ... @MeredithAnya?

Issue #795, opened 2021-01-06
Refactor install.sh into install/*.sh


chadwhitacre:
The `install.sh` script has grown unwieldy. In #766 we introduced a new pattern with a subscript at `install/geoip.sh` that is referred to in the main `install.sh` script. This provides better modularity and comprehension, and makes it easier to test parts of the install script in isolation, whether manually or (someday) automatically.

### Acceptance Criteria

1. `install.sh` works as always.
2. The content of `install.sh` is (almost entirely) a series of calls to individual steps under `install/`.
    1. Find some way to share common things like the `dc`/`dcr` aliases.
    2. Consider going with a `00-foo.sh` prefix scheme, and loading subscripts dynamically in `install.sh`.

PR #900, opened 2021-03-25
Refactor relay config

chadwhitacre:
Step towards #795, introduces more fine-grained testing (not wired to CI yet tho).

### DIFF
--- a/install/_test_setup.sh
+++ b/install/_test_setup.sh
@@ -3,18 +3,6 @@ test ${DEBUG:-''} && set -x
 cd "$(dirname $0)"
 _SANDBOX="$(mktemp -d -t sentry-onpremise-test)"
 
-dump() {
-  echo
-  echo 'dump =============================================\'
-  echo
-  pwd
-  echo
-  ls -FGla *
-  echo
-  echo '\=========================================== dumped'
-  echo
-}
-
 teardown() {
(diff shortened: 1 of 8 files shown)


install/_test_setup.sh
```diff
@@ -0,0 +1,50 @@
+set -euo pipefail
```
BYK, line 1: Not following this file but the rest looks pretty good. Should we add this to our CI suite?

BYK (approved):

install/_test_setup.sh
chadwhitacre: This file is the test framework, basically. `setup` is the important part, it clones the repo into a temp directory, propagates any local changes (expected to be a no-op in CI, but quite useful in local dev: since we symlink changed files, you can edit files in your main working copy and re-run the test inside of the sandbox), and wires up cleanup on teardown (unless `DEBUG` envvar is set).

BYK: It is not clear how one would use this to test, that's what I was thinking.

Maybe if we can add this to CI, we can get cues from there.

chadwhitacre: Ah. For that, check out `relay-credentials-test.sh` on this PR, more tests on the next ... :)

PR #901, opened 2021-03-26
Factor out volume creation

chadwhitacre:
Part of #795, follow-on to #900.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,11 +40,10 @@ jobs:
           ./test.sh
 
       - name: Unit Tests
+        working-directory: install
         run: |
-          cd install
           ./create-docker-volumes-test.sh
           ./relay-credentials-test.sh
-          cd ..
 
       - name: Inspect failure
         if: failure()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,7 @@ jobs:
       - name: Unit Tests
         run: |
           cd install
+          ./create-docker-volumes-test.sh
           ./relay-credentials-test.sh
           cd ..
 
--- a/install/_lib.sh
+++ b/install/_lib.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
+cd "$(dirname $0)/.."
 if [[ ! -d 'install' ]]; then echo 'Where are you?'; exit 1; fi
 _ENV="$(realpath ./.env)"
 
--- a/install/_lib.sh
+++ b/install/_lib.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Work from the onpremise root, no matter which script is called from where.
 if [[ "$(basename $0)" = "install.sh" ]]; then
   cd "$(dirname $0)"
 else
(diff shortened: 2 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       - name: Unit Tests
         run: |
           cd install
+          ./create-docker-volumes-test.sh
           ./relay-credentials-test.sh
           cd ..
```
BYK, line 47:
```suggestion
        working-directory: install
        run: |
          ./create-docker-volumes-test.sh
          ./relay-credentials-test.sh
```

install/_lib.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
+cd "$(dirname $0)/.."
```
BYK, line 3: Maybe a comment explaining here why this is needed and what this is intending to do?

install/_test_setup.sh
```diff
@@ -21,7 +25,7 @@ setup() {
       D)
         rm "$_SANDBOX/$filepath"
         ;;
-      A | M | AM)
+      A | M | AM | ??)
```
BYK, line 28: No idea what this is, comments? :D

install/create-docker-volumes-test.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+source "$(dirname $0)/_test_setup.sh"
+
+expected=7
+count() {
+  docker volume ls --quiet | grep '^sentry-.*' | wc -l
```
BYK, line 6: This actually highlights the need to get this list out somewhere. I think we can do the following:

- Rename all internal volumes in `docker-compose.yml` to *NOT* have the `sentry-` prefix
- Run `docker-compose config --volumes` and filter it with `grep '^sentry-'`
- Use this list to create all global volumes
- Use this list to verify all volumes created

What do you think?

BYK (approved):
Loving this work!

install/create-docker-volumes-test.sh
chadwhitacre: A similar thought occurred to me but I want to avoid scope creep. If/when we land #795 entirely that will give us a solid foundation for doing more refactors like this one in the future.

.github/workflows/test.yml
chadwhitacre: Done in cde39996b19350443764541682b14102b3e84086.

install/_test_setup.sh
chadwhitacre: Done in cde39996b19350443764541682b14102b3e84086. :)

install/_lib.sh
chadwhitacre: Done in cde39996b19350443764541682b14102b3e84086.

install/create-docker-volumes-test.sh
BYK: Just realized renaming these volumes may cause issues during upgrades so we may need to find an alternative method anyway.

BYK (approved):

install/create-docker-volumes-test.sh
chadwhitacre: Right, didn't want to go down that 🐰 🕳️ here! 🙈

PR #910, opened 2021-03-30
ci(test): Separate unit tests so they don't rely on integration tests

BYK:


BYK:
>Do we care about pinning docker-compose?

Shouldn't be necessary for unit tests.

chadwhitacre:
> Shouldn't be necessary for unit tests.

Why not? It's necessary for the integration test, which is basically like running all the unit tests, right?

chadwhitacre (approved):

PR #921, opened 2021-04-12
fix: test.sh should use the variable $_group

BYK:


chadwhitacre (changes requested):
Yes! But we need to set the variables before using them, eh?

BYK:
>Yes! But we need to set the variables before using them, eh?

I mean, I thought about that but seems like they are set elsewhere _magically_ somehow?

chadwhitacre:
Here's where they are set under `install.sh`:

https://github.com/getsentry/onpremise/blob/dd5a7d430e6341315a98dd0ad8da8988d65bab12/install/_lib.sh#L20-L26

Does that cover us here tho? 🤔

install/_lib.sh
```diff
@@ -5,7 +5,7 @@ test "${DEBUG:-}" && set -x
 log_file="sentry_install_log-`date +'%Y-%m-%d_%H-%M-%S'`.txt"
 exec &> >(tee -a "$log_file")
 
-# Work from the onpremise root, no matter which script is called from where.
+# Work from /install/ for install.sh, project root otherwise
```
chadwhitacre, line 8:
```suggestion
# Work from /install/ for install.sh, undefined otherwise
```

chadwhitacre (changes requested):
Comment tweaks

chadwhitacre (approved):
Sike.

chadwhitacre:
Lol confusing.

PR #923, opened 2021-04-13
fix: Ensure snuba cleanup job runs on the correct storage

lynnagara:
Follow up to #920.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -172,7 +172,7 @@ services:
       context: ./cron
       args:
         BASE_IMAGE: "$SNUBA_IMAGE"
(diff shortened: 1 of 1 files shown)


BYK (approved):
Do we need a cleanup job for transactions storage?

lynnagara:
> Do we need a cleanup job for transactions storage?

Yeah, it would be good to add a job for that too

BYK:
@lynnagara can you guide me to add that?

lynnagara:
@BYK The command would be exactly the same as this line, but with `--storage transactions` instead of `--storage errors`

PR #926, opened 2021-04-17
ref(relay): Remove comment from sample config

jan-auer:
This comment has unprofessional tone and does not serve any purpose. Since this is copied to every onpremise installation and I keep copying this code to public issue comments, it's better removed.

### DIFF
--- a/relay/config.example.yml
+++ b/relay/config.example.yml
@@ -8,6 +8,6 @@ processing:
   enabled: true
   kafka_config:
     - {name: "bootstrap.servers", value: "kafka:9092"}
-    - {name: "message.max.bytes", value: 50000000}
+    - {name: "message.max.bytes", value: 50000000} # 50MB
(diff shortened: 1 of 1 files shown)


relay/config.example.yml
```diff
@@ -8,6 +8,6 @@ processing:
   enabled: true
   kafka_config:
     - {name: "bootstrap.servers", value: "kafka:9092"}
-    - {name: "message.max.bytes", value: 50000000} #50MB or bust
+    - {name: "message.max.bytes", value: 50000000}
```
jan-auer, line 11: Alternatively, we could also leave the "50MB" there.

```suggestion
    - {name: "message.max.bytes", value: 50000000} # 50MB
```

BYK (approved):

BYK:
I think this exists in nginx.conf too.

PR #928, opened 2021-04-22
fix: Fix .env path for test scripts

BYK:


chadwhitacre (approved):
Good catch.

BYK:
> ././_lib.sh: line 19: /home/runner/work/onpremise/onpremise/install/.env: No such file or directory

Apparently not enough.

chadwhitacre:
> Apparently not enough.

Where are you seeing that? I see an error on CI for the latest commit, but the initial commit is green afaict. No?

chadwhitacre:
Should we add a test for this? How did this come up in the first place?

install/_lib.sh
```diff
@@ -6,14 +6,13 @@ log_file="sentry_install_log-`date +'%Y-%m-%d_%H-%M-%S'`.txt"
...
-if [[ "$(basename $0)" = "install.sh" ]]; then
+if [[ "$(basename $0)" = "install.sh" || "$(basename $0)" = "test.sh" ]]; then
   cd "$(dirname $0)/install/"
 else
   cd "$(dirname $0)"  # assume we're a *-test.sh script
 fi
 
-_ENV="$(realpath ./.env)"
-
+_ENV="$(realpath ../.env)"
```
chadwhitacre: This is still a problem, though. Right? Don't we need the first commit that varies this with the base dir?

BYK:
>Should we add a test for this? How did this come up in the first place?

We definitely should. I saw this here by chance: https://github.com/getsentry/onpremise/runs/2404982826?check_suite_focus=true#step:4:683

chadwhitacre (changes requested):

install/_lib.sh
BYK: Nope, the logic there was valid :D

chadwhitacre: Got it, because dirname of $0 is `.../install/` if we assume it's a `*-test.sh` script. 👍

chadwhitacre (approved):

PR #964, opened 2021-05-26
Comment out mail.host option in config.example.yml

### LINKED ISSUE #965: SMTP Configuration overriden on install
## Important Details

How are you running Sentry?

On-Premise w/ Docker, version 21.5.1

## Description

I installed Sentry using the install.sh script, then I went on localhost:9000 and I configured my SMTP Settings. However after the installation succeeded I could not send any e-mail. I had Connexion Refused errors in the logs.

## Steps to Reproduce

1. Download Sentry on-premises
2. Install docker-compose
3. Run ./install.sh
4. Access the website at http://localhost:9000
5. Configure SMTP settings
6. Send invitation to join Sentry
7. (Email is not sent)

### What you expected to happen

I wanted the e-mail to be sent. I noticed that in sentry/config.yml I had the following :
```yaml
###############
# Mail Server #
###############

# mail.backend: 'smtp'  # Use dummy if you want to disable email entirely
mail.host: 'smtp'
# mail.port: 25
# mail.username: ''
# mail.password: ''
# mail.use-tls: false
# The email address to send on behalf of
# mail.from: 'root@localhost'
```
When I commented out the mail.host line, everything worked perfectly. It seems this line overrides the configuration on the website. It should be commented out by default.
Thanks.

RomainL972:
The default mail.host option overrides the SMTP configuration during the initial setup and prevents the configured SMTP server from working. It should be commented out so the configuration done on the website is automatically used.

Fixes #965.

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -8,7 +8,7 @@
 ###############
 
 # mail.backend: 'smtp'  # Use dummy if you want to disable email entirely
-mail.host: 'smtp'
+# mail.host: 'smtp'
 # mail.port: 25
 # mail.username: ''
 # mail.password: ''


BYK:
Ugh, this is not great as we should fallback to `smtp` for the bundled smtp server to work out of the box. Great find tho, I'll see how we can fix this in a backwards-compatible way.

chadwhitacre:
@RomainL972 For now we've added a note to the docs as low-hanging fruit. Hopefully that helps the next person.

https://github.com/getsentry/develop/pull/395#event-5097907177

Issue #950, opened 2021-05-13
Feedback about this repository


williamdes:
I use Sentry for quite some time and need to run it self-hosted.

I managed to make the old Docker official version run fine with a very minimalist docker-compose setup.
<details><summary>minimalist setup</summary>
<p>

```yml
version: '2.3'

services:
  redis:
    restart: on-failure:5
    image: redis:6.0-alpine
    healthcheck:
        test: ["CMD", "redis-cli", "ping"]
        interval: 1s
        timeout: 3s
        retries: 30

  postgres:
    restart: on-failure:5
    image: postgres:13-alpine
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U sentry"]
      interval: 10s
      timeout: 5s
      retries: 5
      start_period: 10s
    environment:
      POSTGRES_USER: sentry
      POSTGRES_PASSWORD: "${POSTGRES_PASSWORD}"
      POSTGRES_DB: sentry
    volumes:
     - ${POSTGRES_DATA_VOLUME}:/var/lib/postgresql/data

  sentry:
    image: sentry:9.1.2
    depends_on:
      postgres:
        condition: service_healthy
      redis:
        condition: service_healthy
    ports:
     - 9000:9000
    environment:
      SENTRY_SECRET_KEY: "${SENTRY_SECRET_KEY}"
      SENTRY_POSTGRES_HOST: postgres
      SENTRY_DB_USER: sentry
      SENTRY_DB_PASSWORD: "${POSTGRES_PASSWORD}"
      SENTRY_REDIS_HOST: redis

  cron:
    image: sentry:9.1.2
    depends_on:
      postgres:
        condition: service_healthy
      redis:
        condition: service_healthy
    command: "sentry run cron"
    environment:
      SENTRY_SECRET_KEY: "${SENTRY_SECRET_KEY}"
      SENTRY_POSTGRES_HOST: postgres
      SENTRY_DB_USER: sentry
      SENTRY_DB_PASSWORD: "${POSTGRES_PASSWORD}"
      SENTRY_REDIS_HOST: redis

  worker:
    image: sentry:9.1.2
    depends_on:
      postgres:
        condition: service_healthy
      redis:
        condition: service_healthy
    command: "sentry run worker"
    environment:
      SENTRY_SECRET_KEY: "${SENTRY_SECRET_KEY}"
      SENTRY_POSTGRES_HOST: postgres
      SENTRY_DB_USER: sentry
[...]

PR #973, opened 2021-05-29
ci: Add .gitattributes file to exclude unnecessary files from repo zips

williamdes:
Ref: #950


BYK (approved):
Thanks @williamdes!

jan-auer:
This seems to have regressed onpremise test builds which [require test.sh](https://github.com/getsentry/onpremise/blob/2ab05909248f563147ab1cd4ed7488475debb4d9/.github/workflows/test.yml#L46).

PR #974, opened 2021-05-29
ref: Use alpine for nginx and upgrade nginx from 1.16 to 1.21

williamdes:
- ~~Use alpine for postgres~~
- Use alpine for nginx
- Upgrade nginx from 1.16 to 1.21

NOTE: I did not dare to change the version of clickhouse, but it could use `21.3.12-alpine` if you allow the version to be changed.
Since there is no comment on why the version is so much specific (`20.3.9.70`) and that the clickhouse changelogs are way to large and need expertise I did not change anything.

Possible upgrades:
- PostgreSQL
- memcached
- redis

Ref: #950

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -272,7 +272,7 @@ services:
     <<: *restart_policy
     ports:
       - "$SENTRY_BIND:80/tcp"
-    image: "nginx:1.16"
+    image: "nginx:1.21-alpine"
     volumes:
       - type: bind
         read_only: true
--- a/install/wrap-up.sh
+++ b/install/wrap-up.sh
@@ -3,7 +3,7 @@ if [[ "$MINIMIZE_DOWNTIME" ]]; then
 
   # Start the whole setup, except nginx and relay.
   $dc up -d --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay)$')
-  $dc exec -T nginx service nginx reload
+  $dc restart nginx
 
   docker run --rm --network="${COMPOSE_PROJECT_NAME}_default" alpine ash \
     -c 'while [[ "$(wget -T 1 -q -O- http://web:9000/_health/)" != "ok" ]]; do sleep 0.5; done'


BYK:
>NOTE: I did not dare to change the version of clickhouse, but it could use 21.3.12-alpine if you allow the version to be changed.
Since there is no comment on why the version is so much specific (20.3.9.70) and that the clickhouse changelogs are way to large and need expertise I did not change anything.

We need to stick to this specific version as it is what we use everywhere else. We are aware of breaking changes in v21 so we cannot upgrade to that one.

I'm okay with the Nginx change but not so sure about Postgres. We'd also need to change some CDC stuff @fpacifici added for postrges so I'd make this PR only on Nginx.

@fpacifici - are you aware of any reason why we were using Alpine in some places? I remember talking about switching self-hosted to Alpine too but then deciding against it but don't remember why, do you remember? 😀

docker-compose.yml
```diff
@@ -70,7 +70,7 @@ services:
         hard: 10032
   postgres:
     <<: *restart_policy
-    image: "postgres:9.6"
+    image: "postgres:9.6-alpine"
```
BYK, line 73: Let's do `postrges` on a separate PR (as we may need to punt on it and we already need to change the `wal2json` setup.

BYK (changes requested):
Thanks a lot for this @williamdes. Just requesting changes regarding the Postgres change.

docker-compose.yml
BYK:
```suggestion
    image: "postgres:9.6"
```

williamdes: Done, thank you for your patience !

williamdes:
Seems like something is wrong with "service" I suspect it to be a command that needs to be adjusted for alpine

BYK:
>Seems like something is wrong with "service" I suspect it to be a command that needs to be adjusted for alpine

Eh, glad we have tests 😁

install/wrap-up.sh
```diff
@@ -3,7 +3,7 @@ if [[ "$MINIMIZE_DOWNTIME" ]]; then
 
   # Start the whole setup, except nginx and relay.
   $dc up -d --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay)$')
-  $dc exec -T nginx service nginx reload
+  $dc restart nginx
```
williamdes, line 6: just restart the container, it's the same as trying to reload the container
I did not find a quick alpine way to reload nginx without installing one more package into it just for that one time

BYK: This is the reason: https://github.com/getsentry/onpremise/pull/615#discussion_r477541160

BYK:
I think #1002 will be superseding this.

BYK:
Closing since #1002 landed and surpassed this.

Thanks a lot @williamdes for all your effort both on the issues and here with PRs. 🥳

williamdes:
> Closing since #1002 landed and surpassed this.
> 
> Thanks a lot @williamdes for all your effort both on the issues and here with PRs.

Thanks a lot for your reviews and patience !
I am glad more alpine could enter, looking forward to audit more of this to allow new Sentry to enter in our production env

PR #975, opened 2021-05-29
feat: Add healthchecks for redis, memcached and postgres

williamdes:
Ref: #950

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,9 +59,22 @@ services:
   memcached:
     <<: *restart_policy
     image: "memcached:1.5-alpine"
+    healthcheck:
+      # From: https://stackoverflow.com/a/31877626/5155484
+      test: echo stats | nc 127.0.0.1 11211
+      interval: 2s
+      timeout: 3s
+      retries: 30
+      start_period: 3s
   redis:
     <<: *restart_policy
     image: "redis:5.0-alpine"
+    healthcheck:
+      test: redis-cli ping
+      interval: 2s
+      timeout: 3s
+      retries: 30
+      start_period: 3s
     volumes:
       - "sentry-redis:/data"
     ulimits:
@@ -71,6 +84,13 @@ services:
   postgres:
     <<: *restart_policy
     image: "postgres:9.6"
+    healthcheck:
+      # Using default user "postgres" from sentry/sentry.conf.example.py or value of PING_USER if provided
+      test: ["CMD-SHELL", "pg_isready -U ${PING_USER:-postgres}"]
+      interval: 2s
+      timeout: 3s
+      retries: 30
+      start_period: 10s
     command: ["postgres", "-c", "wal_level=logical", "-c", "max_replication_slots=1", "-c", "max_wal_senders=1"]
     environment:
       POSTGRES_HOST_AUTH_METHOD: "trust"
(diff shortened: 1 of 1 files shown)


williamdes:
For postgres
```

    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U sentry"]
      interval: 10s
      timeout: 5s
      retries: 5
      start_period: 10s
```

But I would need to know the username

BYK:
>But I would need to know the username

Looks like we just use the default username:

https://github.com/getsentry/onpremise/blob/e20e6a3e5483ddfda6d3d9c07cb9ea53d3d22731/sentry/sentry.conf.example.py#L40

docker-compose.yml
```diff
@@ -59,9 +59,18 @@ services:
...
+    healthcheck:
+      test: echo stats | nc 127.0.0.1 11211
+      interval: 10s
+      retries: 60
   redis:
     <<: *restart_policy
     image: "redis:5.0-alpine"
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 1s
```
BYK, line 71: Why do you check redis way more frequently than memcached?

docker-compose.yml
```diff
@@ -59,9 +59,18 @@ services:
...
     image: "memcached:1.5-alpine"
+    healthcheck:
+      test: echo stats | nc 127.0.0.1 11211
+      interval: 10s
+      retries: 60
   redis:
     <<: *restart_policy
     image: "redis:5.0-alpine"
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
```
BYK, line 70: Any reason why we cannot do 
```suggestion
      test: redis-cli ping
```

like above?

BYK (approved):

docker-compose.yml
williamdes: Well this is because Redis has a real check and memcached uses some specific commands to preform some kind of check
That said I can research a bit more to find a better way to check memcached

BYK: Maybe we can add a comment about this?

williamdes: > Well this is because Redis has a real check and memcached uses some specific commands to preform some kind of check

I was stupid, this was indeed the right command as it asks stats to memcached and it would not respond if it was down
it's not an official way but enough for PerconaDB to post an article about it: https://www.percona.com/blog/2008/11/26/a-quick-way-to-get-memcached-status/

BYK: Great to hear, except for the part you call yourself stupid 😀

docker-compose.yml
```diff
@@ -71,6 +84,13 @@ services:
   postgres:
     <<: *restart_policy
     image: "postgres:9.6"
+    healthcheck:
+      # Using default user "postgres" from sentry/sentry.conf.example.py or value of PING_USER if provided
+      test: ["CMD-SHELL", "pg_isready -U ${PING_USER:-postgres}"]
```
BYK, line 89: Let's scope the variable to postgres only:
```suggestion
      # Using default user "postgres" from sentry/sentry.conf.example.py or value of POSTGRES_USER if provided
      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-postgres}"]
```

BYK (approved):

PR #978, opened 2021-05-31
fix(test): Export test script

jan-auer:
Follow-up to #973. Restores exporting test.sh so that integration tests can use
it after downloading the release archive.


jan-auer:
My intention is to unblock Sentry PRs as soon as possible, since they require the integration test suite to pass. There's probably better ways to pull the test script for the respective revision.

There was prior conversation in https://github.com/getsentry/onpremise/issues/950#issuecomment-840768818 on whether `test.sh` should be included or not.

PR #984, opened 2021-06-07
meta: Add some changelog

BYK:
@billyvg has a breaking change coming up soon so let's use this as an opportunity to add a changelog. This changelog will only capture important announcements (such as this breaking change over at Sentry) and changes in this repo for now.


armenzg (approved):
Excited!

For anyone looking at this, there's an example in [here](https://github.com/getsentry/craft#changelog-policies) that documents what `auto` does.

billyvg (approved):

PR #988, opened 2021-06-08
Make a reset button

chadwhitacre:
The purpose of this script is to make it easy to reset a local onpremise install to a clean state, optionally installing a particular version/sha. This is helpful for testing out upgrade scenarios such as in https://github.com/getsentry/sentry/issues/26134.

### DIFF
--- a/reset.sh
+++ b/reset.sh
@@ -25,7 +25,7 @@ confirm "☠️  Warning! 😳 This is highly destructive! 😱 Are you sure you
 echo "Okay ... good luck! 😰"
 
 # Hit the reset button.
-docker compose down --volumes --rmi local
+docker compose down --volumes --remove-orphans --rmi local
 
 # Remove any remaining (likely external) volumes with name matching '.*sentry.*'.
 for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
--- a/reset.sh
+++ b/reset.sh
@@ -34,10 +34,15 @@ for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
     || echo "Skipped volume: $volume"
 done
 
-# If we have a version (or other git ref) given, install it.
-ref="${1:-}"
-if [ -z "$ref" ]; then
+# If we have a version given, install it.
+
+# Note that arbitrary git refs won't work, because the *_IMAGE variables in
+# .env will almost certainly point to :latest. Release tags of onpremise are
+# generally the only refs where these component versions are pinned.
+
+version="${1:-}"
+if [ -z "$version" ]; then
   exit
 fi
-git checkout "$ref"
+git checkout "$version"
 exec ./install.sh
--- a/reset.sh
+++ b/reset.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # The purpose of this script is to make it easy to reset a local onpremise
-# install to a clean state, optionally installing a particular version/sha.
+# install to a clean state, optionally installing a particular version/ref.
 
 set -euo pipefail
 
@@ -34,7 +34,7 @@ for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
     || echo "Skipped volume: $volume"
 done
 
-# If we have a version (or other sha) given, install it.
+# If we have a version (or other git ref) given, install it.
 ref="${1:-}"
 if [ -z "$ref" ]; then
   exit
--- a/reset.sh
+++ b/reset.sh
@@ -34,15 +34,16 @@ for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
     || echo "Skipped volume: $volume"
 done
 
-# If we have a version given, install it.
+# If we have a version given, switch to it.
 
 # Note that arbitrary git refs won't work, because the *_IMAGE variables in
 # .env will almost certainly point to :latest. Release tags of onpremise are
 # generally the only refs where these component versions are pinned.
 
 version="${1:-}"
-if [ -z "$version" ]; then
-  exit
+if [ -n "$version" ]; then
+  git checkout "$version"
 fi
-git checkout "$version"
+
+# Install.
 exec ./install.sh
--- a/reset.sh
+++ b/reset.sh
@@ -46,8 +46,8 @@ echo "Okay ... good luck! 😰"
 # Hit the reset button.
 docker compose down --volumes --remove-orphans --rmi local
 
-# Remove any remaining (likely external) volumes with name matching '.*sentry.*'.
-for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
+# Remove any remaining (likely external) volumes with name matching 'sentry-.*'.
+for volume in $(docker volume list --format '{{ .Name }}' | grep '^sentry-'); do
   docker volume remove $volume > /dev/null \
     && echo "Removed volume: $volume" \
     || echo "Skipped volume: $volume"
--- a/reset.sh
+++ b/reset.sh
@@ -20,6 +20,25 @@ function confirm () {
   fi
 }
 
+
+# If we have a version given, validate it.
+# ----------------------------------------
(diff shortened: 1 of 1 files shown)


reset.sh
```diff
@@ -0,0 +1,43 @@
...
+function confirm () {
+  read -p "$1 [y/n] " confirmation
+  if [ "$confirmation" != "y" ]; then
+    echo "Canceled. 😅"
+    exit
+  fi
+}
+
+# Make sure they mean it.
+confirm "☠️  Warning! 😳 This is highly destructive! 😱 Are you sure you wish to proceed?"
```
BYK, line 24: 🤣

reset.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "Okay ... good luck! 😰"
+
+# Hit the reset button.
+docker compose down --volumes
```
BYK, line 28: Let's make it even more destructive:
```suggestion
docker compose down -v --remove-orphans --rmi local
```

reset.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Remove any remaining (likely external) volumes with name matching '.*sentry.*'.
+for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
```
BYK, line 31: Why not
```suggestion
for volume in $(docker volume ls | grep '^sentry\-'; do
```

Should be safer as we know all our volumes are prefixed with `sentry-`. We may even go more conservative and use the list from the volume creation script?

reset.sh
```diff
@@ -0,0 +1,43 @@
...
+  docker volume remove $volume > /dev/null \
+    && echo "Removed volume: $volume" \
+    || echo "Skipped volume: $volume"
```
BYK, line 34: Skipped? More like "couldn't"?

reset.sh
```diff
@@ -0,0 +1,43 @@
...
+done
+
+# If we have a version (or other git ref) given, install it.
+ref="${1:-}"
+if [ -z "$ref" ]; then
+  exit
+fi
+git checkout "$ref"
```
BYK, line 42: This wouldn't work for git refs as unless they are tagged releases, `SENTRY_IMAGE` etc values will be pointing to `latest`. It is hard to get this feature right unless people are using version numbers.

BYK (changes requested):
This is pretty useful thanks a lot! Rejected due to some concerns around volume removal and the git ref parts.

Once this lands, we should document it.

reset.sh
chadwhitacre: Yes but `skipped` has the same number of letters as `removed` 🙈 and presumably there will have been an error message emitted which gives additional context.

reset.sh
chadwhitacre: I had some volumes locally that were prefixed with `onpremise-sentry-`. Not sure where those came from but that is why I went with the more relaxed match. Since one of the main features of this script is to switch between versions locally and volumes can and do change over time, matching broadly makes more sense to me.

reset.sh
chadwhitacre: > `--remove-orphans`

[Docs](https://docs.docker.com/engine/reference/commandline/compose_down/):

> Remove containers for services not defined in the Compose file.

If I have other non-Sentry containers on my laptop I don't want those clobbered here, do I?

reset.sh
BYK: >I had some volumes locally that were prefixed with onpremise-sentry-. Not sure where those came from but that is why I went with the more relaxed match.

They come from docker-compose and are removed with the `docker-compose down -v` command.

>Since one of the main features of this script is to switch between versions locally and volumes can and do change over time, matching broadly makes more sense to me.

Even if we add/remove volumes in time the principles for those are:
1. Prefer local volumes so they are managed by `docker-compose`
2. Always prefix global volumes with `sentry-` to avoid issues like this.

I'd even argue we should use local volumes only but it is not easy to migrate.

reset.sh
BYK: That's a bit of a misrepresentation. They are trying to say containers under the same "project" umbrella. It wouldn't clean up all the containers on your laptop. `docker container prune` exists for that :)

reset.sh
chadwhitacre: Good point. Addressed in 9dbd0af4b53356ebffab312e3b3fd0f2d9b7e3e5. 👍

reset.sh
chadwhitacre: `--rmi local` in 387a72953a631ce0fd5540e0deb3cc299c48f8de.

chadwhitacre:
Back over to you @BYK. Note the semantic change in 5454714094aec4b9afcb549b4af3b515c7cad84b.

reset.sh
```diff
@@ -34,10 +34,16 @@ for volume in $(docker volume list --format '{{ .Name }}' | grep sentry); do
...
-ref="${1:-}"
-if [ -z "$ref" ]; then
-  exit
+# If we have a version given, switch to it.
+
+# Note that arbitrary git refs won't work, because the *_IMAGE variables in
+# .env will almost certainly point to :latest. Release tags of onpremise are
+# generally the only refs where these component versions are pinned.
+
+version="${1:-}"
```
BYK, line 43: Let's ensure that `$version` is actually a known tag here (in case people try to pass branches or refs):
```shell
git rev-parse --verify -q refs/tags/$version
```

reset.sh
```diff
@@ -25,7 +25,7 @@ confirm "☠️  Warning! 😳 This is highly destructive! 😱 Are you sure you
 echo "Okay ... good luck! 😰"
 
 # Hit the reset button.
-docker compose down --volumes
+docker compose down --volumes --rmi local
```
BYK, line 28: This is not a hard requirement but having `--remove-orphans` would help, when we add/remove services to the `docker-compose.yml` file. After a checkout, they will become "orphaned" by this terminology and unless that option is passed, they would stay running. This is not something we want ~99% of the time.

BYK (changes requested):

reset.sh
chadwhitacre: Done in 2f5c171059500b6b10a71659ee68c17bb0876e46 and 61b5de2e7aeef55c0a7c4444938ca393a118f97f.

reset.sh
chadwhitacre: Sorry, I missed https://github.com/getsentry/onpremise/pull/988#discussion_r648240411 earlier. Done in e405249807bd4d13bc09ef3a3025adb03e651830. 👍

reset.sh
chadwhitacre: Done in 50c2ab91b466d64e09d6755902d38884df114f0d.

reset.sh
```diff
@@ -0,0 +1,62 @@
...
+# ----------------------------------------
+# Note that arbitrary git refs won't work, because the *_IMAGE variables in
+# .env will almost certainly point to :latest. Tagged releases are generally
+# the only refs where these component versions are pinned, so enforce that
+# we're targeting a valid tag here. Do this early in order to fail fast.
+
+version="${1:-}"
+if [ -n "$version" ]; then
+  set +e
+  git rev-parse --verify --quiet "refs/tags/$version" > /dev/null
```
BYK, line 34: You shouldn't need `> /dev/null` here due to the `--quiet` option.

Love the early verification before doing anything ❤️

BYK (approved):

reset.sh
chadwhitacre: `--quiet` mutes stderr output, the stdout redirect suppresses the noise of a successful match.

```
$ git rev-parse --verify --quiet refs/tags/21.3.0
0dc51be35274db95c42c00f53d7f56c2aac4d84c
$
```

BYK: Ugh, you're right sorry. Turns out it only suppresses error messages 🤦🏻

PR #1011, opened 2021-06-29
Use the new GH issue forms

chadwhitacre:
No more janky GHA validation!

cf. https://github.com/getsentry/sentry/pull/26898


BYK (approved):
Yasss!

PR #1015, opened 2021-07-01
feat: Support custom CA roots

Spice-King:
Mount a certificate folder to local ca storage in containers,
and add update command to cron image's entrypoint.

Result of poking and prodding from getsentry/sentry#26851


cron/entrypoint.sh
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+if [ "$(ls -A /usr/local/share/ca-certificates/)" ]; then
+  update-ca-certificates
+fi
```
BYK, line 5: Shouldn't this change be reflected in Sentry entrypoint too?

Also, do you think we can test this somehow? Maybe with a fake CA we use for tests or something? I don't want this feature to get broken in the future without us noticing.

BYK (changes requested):

BYK:
Thanks so much for doing this and also submitting a patch for the docs!

BYK:
Well also, seeing getsentry/develop#369 reminded me that we now have a CHANGELOG here so it would be great to mention this here with a link to the new docs: https://github.com/getsentry/onpremise/blob/master/CHANGELOG.md

cron/entrypoint.sh
Spice-King: > Shouldn't this change be reflected in Sentry entrypoint too?
That's the point of the getsentry/sentry#27062 PR, and since the cron Dockerfile uses a dynamic base image, it overrides the entrypoint of the parent image and uses it's own script as the entrypoint, putting it in here would make sure any cron jobs get the same treatment for the CA roots in case they need to reach out.

> Also, do you think we can test this somehow? Maybe with a fake CA we use for tests or something?
Fair idea. The short of it would be to generate a CA root, do up a otherwise valid certificate signed by it, then spin up a HTTPS server and have Sentry try poking it, all while being reliable since we can't trust any hostnames nor IPs.

There is `host.docker.internal` and `host-gateway`, but they are not portable with the Win/Mac Docker versions having only one and Linux the other. Probably could solve IP/DNS by grabbing the the gateway IP from the docker network config and having an override to stuff a hostname in. That leaves figuring out a reliable way of testing Sentry can poke a URL. Could cheap out and just spawn a shell that uses python and the request lib in the `web` container and pokes the test HTTPS server and spits error codes.

Ironically this opens up a much larger can of worms than the original PR does. Probably worth making a sub folder to move the added test files into, since this will probably not all nicely fit in the existing `test.sh` file.

Spice-King:
> Well also, seeing [getsentry/develop#369](https://github.com/getsentry/develop/pull/369) reminded me that we now have a CHANGELOG here so it would be great to mention this here with a link to the new docs: https://github.com/getsentry/onpremise/blob/master/CHANGELOG.md

Did that, rebased and updated all the repos.

cron/entrypoint.sh
BYK: >That's the point of the getsentry/sentry#27062 PR, and since the cron Dockerfile uses a dynamic base image, it overrides the entrypoint of the parent image and uses it's own script as the entrypoint, putting it in here would make sure any cron jobs get the same treatment for the CA roots in case they need to reach out.

Ehh, sorry I missed the PR over at Sentry. Makes sense, that said we have an entrypoint override here in onpremise for Sentry so I'm thinking maybe it is cleaner to keep everything in this repo?

>There is host.docker.internal and host-gateway, but they are not portable with the Win/Mac Docker versions having only one and Linux the other. Probably could solve IP/DNS by grabbing the the gateway IP from the docker network config and having an override to stuff a hostname in. 

We can easily add a test service into the compose mix to be able to easily ping it. 

>That leaves figuring out a reliable way of testing Sentry can poke a URL. Could cheap out and just spawn a shell that uses python and the request lib in the web container and pokes the test HTTPS server and spits error codes.

I think this is a fair way to test it until we find a better way?

>Ironically this opens up a much larger can of worms than the original PR does. Probably worth making a sub folder to move the added test files into, since this will probably not all nicely fit in the existing test.sh file.

Agreed, willing to chip in with tests once we solidify the implementation part.

BYK (commented):
Looks pretty good, holding until we figure out what to do with sentry/onpremise repo split and tests.

cron/entrypoint.sh
Spice-King: > Makes sense, that said we have an entrypoint override here in onpremise for Sentry so I'm thinking maybe it is cleaner to keep everything in this repo?
I had overlooked that entrypoint as onpremise does not generate a custom image for sentry, but I still needed to make a PR to set the env vars in the image to get it to use the system CA bundle and the overridden entrypoint still invokes the original one. It ultimately made a little more sense to me to include updating the system bundle in the base image.

> We can easily add a test service into the compose mix to be able to easily ping it.
Probably the sane way to go about it, though I would think it would be more ideal to not mess with the compose config since that is kinda what is being tested. Probably fine though if we never do anything to the normal services. *Probably.*

> Agreed, willing to chip in with tests once we solidify the implementation part.
I have a vague idea on how I would set that up, but I'll chew on that after a late lunch. Breaking up `./test.sh` into at a pre-test script to isolate the test setup and all the other tests to their own logical unit would be a good start, and it's own PR.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 21.6.2
 
 - BREAKING CHANGE: The frontend bundle will be loaded asynchronously (via [#25744](https://github.com/getsentry/sentry/pull/25744)). This is a breaking change that can affect custom plugins that access certain globals in the django template. Please see https://forum.sentry.io/t/breaking-frontend-changes-for-custom-plugins/14184 for more information.
+- feat: Support custom CA roots ([#27062](https://github.com/getsentry/sentry/pull/27062)), see the [docs](https://develop.sentry.dev/self-hosted/custom-ca-roots/) for more details.
```
chadwhitacre, line 6: Is this the right place for this entry? Don't we want this in an "Unreleased" section or something? 21.6.2 is already out.

test/ca-setup.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# docker-compose up -d test
+# docker-compose restart test
+
+# # curl https://self.test --cacert ca.crt --resolve self.test:443:$(docker inspect sentry_onpremise_test_1| jq .[0].NetworkSettings.Networks.sentry_onpremise_default.IPAddress -r)
+# curl https://self.test  --cacert certificates/ca.crt --resolve self.test:443:$(docker inspect sentry_onpremise_test_1| jq .[0].NetworkSettings.Networks.sentry_onpremise_default.IPAddress -r)
+# cp ./test/cert-test.py ./sentry/
+# # docker-compose run --rm web python3 /etc/sentry/cert-test.py
+# docker-compose down
+# rm -f ca.* certificates/* test/nginx/*.test.* sentry/cert-test.py
```
chadwhitacre, line 41: Lots of commented-out code here. Could we remove those and add some meaningful comments instead? What exactly are we testing here? Or is this just setup for a `cert-test.py`?

test/core.sh
```diff
@@ -119,3 +120,7 @@ echo "${_endgroup}"
 echo "${_group}Ensure cleanup crons are working ..."
 $dc ps | grep -q -- "-cleanup_.\+[[:space:]]\+Up[[:space:]]\+"
 echo "${_endgroup}"
+
+echo "${_group}Test customs CAs work ..."
```
chadwhitacre, line 124:
```suggestion
echo "${_group}Test custom CAs work ..."
```

Spice-King:
Been busy, lost the notifications in the noise. Sorry about that. Going to quickly rebase and double check stuff.

CHANGELOG.md
Spice-King: Ah, by-product of me not checking what rebase did. Lesson learned.

test/ca-setup.sh
Spice-King: Left over test code mistakenly committed, I'll sprinkle in a few notes.

test/core.sh
Spice-King: Too much poking and prodding government paper work. >.> Fixed.

chadwhitacre:
I rebased this now that 21.7.0 is out.

chadwhitacre:
[Seeing](https://github.com/getsentry/onpremise/pull/1015/checks?check_run_id=3087706777#step:4:672) a test failure:

```
python3: can't open file '/etc/sentry/cert-test.py': [Errno 2] No such file or directory
```

Seeing a weirder test failure locally, not sure what's up. 🤔

```
./test/core.sh: line 90: DSN_PIECES[0]: unbound variable
```

chadwhitacre:
> Seeing a weirder test failure locally, not sure what's up.

Sorted that out in https://github.com/getsentry/onpremise/pull/1031, rebased again.

chadwhitacre:
> Seeing a test failure:

Did a local refactor, and now back to seeing this failure locally.

chadwhitacre:
Alright, I think I've almost wrapped my head around this via the refactor in 600becb29afed704cd56fcce353c9533c823454e. Not quite passing yet but I think I'm close. Gonna pause there for the weekend, will revisit on Monday and see where things stand. The approach seems straightforward, I appreciate the effort to test, would love to get that across the finish line.

chadwhitacre:
Did a little more refactoring, I'm not focused on achieving the same failure in both local and CI. From there I/we can figure out how to fix the failure. :)

chadwhitacre:
@Spice-King Okay, I've got the same result in CI and local. I'll keep poking at is as I'm able, but if you get to it before I do ... could you try out the latest locally for you and see if you get the same [result](https://github.com/getsentry/onpremise/pull/1015/checks?check_run_id=3163267878#step:4:801)? My hunch is that I boogered something up in all of my refactoring, hopefully it's a quick fix. :)

```
Custom CA cert failed to work.
  Traceback (most recent call last):
    File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 600, in urlopen
      chunked=chunked)
    File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 343, in _make_request
      self._validate_conn(conn)
    File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 839, in _validate_conn
      conn.connect()
    File "/usr/local/lib/python3.6/site-packages/urllib3/connection.py", line 344, in connect
      ssl_context=context)
    File "/usr/local/lib/python3.6/site-packages/urllib3/util/ssl_.py", line 347, in ssl_wrap_socket
      return context.wrap_socket(sock, server_hostname=server_hostname)
    File "/usr/local/lib/python3.6/ssl.py", line 407, in wrap_socket
      _context=self, _session=session)
    File "/usr/local/lib/python3.6/ssl.py", line 817, in __init__
      self.do_handshake()
    File "/usr/local/lib/python3.6/ssl.py", line 1077, in do_handshake
      self._sslobj.do_handshake()
    File "/usr/local/lib/python3.6/ssl.py", line 689, in do_handshake
      self._sslobj.do_handshake()
  ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)
```

chadwhitacre:
P.S. The repro steps are:

```
cd onpremise
git pull
./_integration-test/run.sh
```

Spice-King:
@chadwhitacre I have no clue why it's opted to ignore the environment variable when it accepted it before. Spent my while evening on it. I'll prod my work's instance I altered in a crude way using a pile of overrides. Last I checked, it works there. If all less failed, I do have an idea that should work, but is more ham fisted than what I'd have liked. The "It can't be tight if it's a liquid" kind of option on the Tool Expectations chart.

chadwhitacre:
Huzzah! I'm able to manually set up a successful request locally. Will see about test/CI ...

```
root@256dce7eda3d:/usr/local/share/ca-certificates# python /etc/sentry/test-custom-ca-roots.py 
ok
root@256dce7eda3d:/usr/local/share/ca-certificates#
```

chadwhitacre:
Alright, we're green! @BYK You had some feedback on this approach?

chadwhitacre:
Hrm, 5708586ed0af10c2a9d22fd3de93f3e755aebcd7 passes locally with this diff. 🤔

```diff
diff --git a/.env b/.env
index d49dae9..8ba24c3 100644
--- a/.env
+++ b/.env
@@ -3,7 +3,7 @@ SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
 SENTRY_BIND=9000
-SENTRY_IMAGE=getsentry/sentry:nightly
+SENTRY_IMAGE=us.gcr.io/sentryio/sentry:16f56945390034929c159ac0dab81584602eb65c
 SNUBA_IMAGE=getsentry/snuba:nightly
 RELAY_IMAGE=getsentry/relay:nightly
 SYMBOLICATOR_IMAGE=getsentry/symbolicator:nightly
```

sentry/entrypoint.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -e
 
+if [ "$(ls -A /usr/local/share/ca-certificates/)" ]; then
+  update-ca-certificates
+fi
+
```
chadwhitacre, line 7: Can we get [this logging](https://github.com/getsentry/sentry/pull/27062/commits/449412b925fbef18963f3b0bb95133e8caa82841) here to aid future debugging?

install/_lib.sh
```diff
@@ -6,7 +6,7 @@ log_file="sentry_install_log-`date +'%Y-%m-%d_%H-%M-%S'`.txt"
 exec &> >(tee -a "$log_file")
 
 # Work from /install/ for install.sh, project root otherwise
-if [[ "$(basename $0)" = "install.sh" || "$(basename $0)" = "test.sh" ]]; then
+if [[ "$(basename $0)" = "install.sh"  ]]; then
```
chadwhitacre, line 9: Was this necessary to get this working, or a sidecar?

docker-compose.yml
```diff
@@ -35,14 +35,16 @@ x-sentry-defaults: &sentry_defaults
...
+    GRPC_DEFAULT_SSL_ROOTS_FILE_PATH_ENV_VAR: "/etc/ssl/certs/ca-certificates.crt"
     # Leaving the value empty to just pass whatever is set
     # on the host system (or in the .env file)
     SENTRY_EVENT_RETENTION_DAYS:
   volumes:
     - "sentry-data:/data"
     - "./sentry:/etc/sentry"
     - "./geoip:/geoip:ro"
-    - "./certificates:/usr/local/share/ca-certificates:ro"
+    - "./certificates:/usr/local/share/ca-certificates"
```
chadwhitacre, line 47: Ah! Permissions after all. :) This is the heart of the fix, ya?

docker-compose.yml
```diff
@@ -35,14 +35,16 @@ x-sentry-defaults: &sentry_defaults
     PYTHONUSERBASE: "/data/custom-packages"
     SENTRY_CONF: "/etc/sentry"
     SNUBA: "http://snuba-api:1218"
+    REQUESTS_CA_BUNDLE: "/etc/ssl/certs/ca-certificates.crt"
+    GRPC_DEFAULT_SSL_ROOTS_FILE_PATH_ENV_VAR: "/etc/ssl/certs/ca-certificates.crt"
```
chadwhitacre, line 39: Why not `DEFAULT_CA_BUNDLE`? Why was that in the original and why do we not need it here?

install/_lib.sh
BYK: `test.sh` doesn't exist anymore so we should not refer to it?

sentry/entrypoint.sh
BYK: `update-ca-certificats` _always_ outputs something, hence the former debugging lines were not necessary.

docker-compose.yml
BYK: Not really, trying to bring this back TBH. The issue was using `$dc exec -T web` instead of `$dcr web`.

docker-compose.yml
BYK: That was defined as a "common variable" to not repeat the value, since we cannot do that here elegantly, opted to axe it and repeat the value.

docker-compose.yml
chadwhitacre: Interesting. Say more? Why did that cause the custom certificate to not be appended to `/ets/ssl/certs/ca-certificates.crt`?

install/_lib.sh
chadwhitacre: Ah, right. And we removed that way earlier on this PR. 😁

docker-compose.yml
BYK: `docker-compose exec` runs the command in the same container that is created by `docker-compose up`. From the logs, I deduced that `update-ca-certificates` were running successfully for all Sentry containers (including the derived `cleanup` cron containers) except for `web`. On `web`, it was getting cut off at the point where it was scanning the disk for new certs. Then I realized we were running `docker-compose exec <our test script>` too early, before the container fully starts up, and interrupt the certificate update step.

sentry/entrypoint.sh
chadwhitacre: I guess my argument would be that I don't know what `update-ca-certificates` outputs, and it's easier for me to find something grep for by looking in here than by learning more about `update-ca-certificates`. Not a blocker.

docker-compose.yml
chadwhitacre: Okay, and for the record `$dcr` [expands to](https://github.com/getsentry/onpremise/blob/3f713c27439e8e97cefb5d0a41e3e0ad99a3c0e6/install/_lib.sh#L29) `docker run --rm`, so we're running the tests in an isolated new container vs. running in the existing one. Makes sense, thanks! 👍 

https://github.com/getsentry/onpremise/blob/3f713c27439e8e97cefb5d0a41e3e0ad99a3c0e6/_integration-test/run.sh#L126

docker-compose.yml
Spice-King: @BYK Actually, no. botocore uses it.

chadwhitacre: Thanks for speaking up @Spice-King, I had a hunch but didn't follow it up.

BYK: Ah, great then I'll bring it back.

Spice-King: I'd not have used it if it was not needed, but it was a more descriptive one to try and be DRY with, so I used to define the rest.

chadwhitacre: We should document this, at least in the code here, possibly on `develop`.

BYK:
Okay, I think this is good for a final review @chadwhitacre and @Spice-King

docker-compose.yml
BYK: Done and done.

chadwhitacre (approved):
Ship it! 🚢

PR #1017, opened 2021-07-01
Require version on bug form

chadwhitacre:
Follow-up to b30e7ef94b3e5a4cbe04ff5c9c3d92aff75f93f7. This is not required in `sentry` because most folks are on SaaS and we don't care about version for that.


BYK (commented):
YES!

Shall we also add a note saying `latest on GitHub` is not a version? 🤣

chadwhitacre:
Well we have the placeholder text already that will justify closing the issue if they don't provide a real version. 😏

BYK (approved):
Sorry, I meant to approve this already.

PR #1026, opened 2021-07-07
fix(wal2json): Respect http_proxy/https_proxy/no_proxy when installing

fnkr:
`http_proxy`, `https_proxy` and `no_proxy` environment variables should be forwarded to the curl container.

### DIFF
--- a/install/install-wal2json.sh
+++ b/install/install-wal2json.sh
@@ -4,11 +4,15 @@ FILE_TO_USE="../postgres/wal2json/wal2json.so"
 ARCH=$(uname -m)
 FILE_NAME="wal2json-Linux-$ARCH-glibc.so"
 
-DOCKER_CURL="docker run --rm curlimages/curl"
+docker_curl() {
+    # The environment variables can be specified in lower case or upper case.
+    # The lower case version has precedence. http_proxy is an exception as it is only available in lower case.
+    docker run --rm -e http_proxy -e https_proxy -e HTTPS_PROXY -e no_proxy -e NO_PROXY curlimages/curl "$@"
+}
 
 if [[ $WAL2JSON_VERSION == "latest" ]]; then
     VERSION=$(
-        $DOCKER_CURL https://api.github.com/repos/getsentry/wal2json/releases/latest |
+        docker_curl https://api.github.com/repos/getsentry/wal2json/releases/latest |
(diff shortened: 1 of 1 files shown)


install/install-wal2json.sh
```diff
@@ -4,11 +4,13 @@ FILE_TO_USE="../postgres/wal2json/wal2json.so"
 ARCH=$(uname -m)
 FILE_NAME="wal2json-Linux-$ARCH-glibc.so"
 
-DOCKER_CURL="docker run --rm curlimages/curl"
+docker_curl() {
+    docker run --rm -e http_proxy="${http_proxy:-}" -e HTTPS_PROXY="${HTTPS_PROXY:-}" curlimages/curl "$@"
```
BYK, line 8: Just telling the names to Docker will make it pass the values defined in the local env to the container.

Btw. `http_proxy` and `HTTP_PROXY` are interchangeable, right? I want to be consistent with the naming/casing.

```suggestion
    docker run --rm -e HTTP_PROXY -e HTTPS_PROXY curlimages/curl "$@"
```

BYK (changes requested):

BYK:
Thanks so much for the PR!

install/install-wal2json.sh
fnkr: > The  environment  variables  can  be specified in lower case or upper case. The lower case version has precedence. http_proxy is an exception as it is only available in lower case. [Source](https://manpages.ubuntu.com/manpages/focal/en/man1/curl.1.html)

We can do all lowercase. I updated the PR and also added `no_proxy`. Or should we add all variants supported by curl?

* http_proxy
* https_proxy
* HTTPS_PROXY
* no_proxy
* NO_PROXY

BYK: I guess it is safest to add all supported?

BYK (approved):

install/install-wal2json.sh
BYK: Also, would be great to add this edge case about `http_proxy` as a comment into the code.

fnkr: Done.

BYK: You are fast! 🚤

BYK (approved):

PR #1040, opened 2021-07-20
Add mail.use-ssl and mail.list-namespace in config.example.yml

ldsink:
Cannot send emails after upgraded. After checking, I found that `mail.use-ssl` and `mail.list-namespace` must be set.
  
Add those to `config.example.yml`, make it easy to known.


chadwhitacre (approved):
Seems reasonable, thanks @ldsink. I did some grepping in the `sentry` codebase and I do find these two config keys in use. 👍

PR #1049, opened 2021-08-04
fix: Fix `curl` image to version 7.77.0

BYK:
Turns out the latest, `7.78.0` may have issues with DNS resolution from time to time (I experienced this locally). It is also a good practice to fix it to a specific version.


chadwhitacre (approved):
Seems reasonable, did not test.

PR #1051, opened 2021-08-04
upgrade: use --ansi never for docker-compose

BYK:
Starting from `docker-compose` v1.28.0, the `--no-ansi` option is deprecated and a new, `--ansi never` option is introduced instead. This PR makes the deprecation warnings around this go away but bumps the minimum docker-compose version required to `1.28.0` as the older versions don't support the new option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Official bootstrap for running your own [Sentry](https://sentry.io/) with [Docke
 ## Requirements
 
 * Docker 19.03.6+
-* Compose 1.24.1+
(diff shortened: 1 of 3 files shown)


chadwhitacre (approved):
Does this get us `docker compose` as well?

chadwhitacre:
> Does this get us `docker compose` as well?

Sorry, I guess that's on the `docker` side. 😊

BYK:
>Does this get us docker compose as well?

I'll look into that later 🙂

PR #1052, opened 2021-08-04
fix(tests): Django 2.0 and 2.1 compatible csrf token extraction

joshuarli:
See https://github.com/getsentry/sentry/pull/27555#issuecomment-893000179.

I also thought it would be more reliable to look for it in curl's cookiejar file, but it doesn't seem to be set.

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -57,7 +57,7 @@ login () {
     exit -1
   fi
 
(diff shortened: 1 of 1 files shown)


BYK:
>but it doesn't seem to be set.

Do you mean it is not in the cookiejar file?

BYK (approved):

joshuarli:
> > but it doesn't seem to be set.
> 
> Do you mean it is not in the cookiejar file?

Correct.

BYK:
That's pretty weird and unexpected but if this works, I'm good :)

PR #1053, opened 2021-08-05
fix(snuba-api): wait for clickhouse to be healthy

pharindoko:
to avoid a race condition in install.sh process at step
https://github.com/getsentry/onpremise/blob/7e7401a668987c44863c403548ba3dbf91c701da/install/bootstrap-snuba.sh#L4

closes #1033

PR #1057, opened 2021-08-06
feat: Leverage health checks for depends_on

BYK:
Expands the work on #1053 to all applicable services.


chadwhitacre (approved):
I like it!

PR #1061, opened 2021-08-16
Update check-minimum-requirements.sh

alex1702:
Updatet minimum docker-compose version to 1.29.2.

I saw the new version 21.8 and just tried to update on my server. Unfortunately the **install.sh** script broke without a proper error message. Then I saw that the docker-compose version must be higher on the release, updated it on my server and the installer then ran through.
So that a proper error message should come here is my patch for the 21.8.0 .

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -1,7 +1,7 @@
 echo "${_group}Checking minimum requirements ..."
 
 MIN_DOCKER_VERSION='19.03.6'
-MIN_COMPOSE_VERSION='1.28.0'
+MIN_COMPOSE_VERSION='1.29.2'
 MIN_RAM_HARD=3800 # MB
 MIN_RAM_SOFT=7800 # MB
 MIN_CPU_HARD=2


chadwhitacre:
> Unfortunately the `install.sh` script broke without a proper error message. 

My hunch is that the script broke due to #1062. Can you review that ticket and confirm, @alex1702?

If so, then I don't believe that requiring an even _newer_ version of `docker-compose` is the right solution.

chadwhitacre (changes requested):
Not sure we want to _increase_ `MIN_COMPOSE_VERSION` in light of #1062.

BYK:
@chadwhitacre I think there's a point here: we must have a test case with the minimum supported version of `docker-compose`. We may decide to have a matrix or just sticking with the oldest we support.

chadwhitacre:
@BYK So close this in favor of #1066, yes?

BYK:
>@BYK So close this in favor of #1066, yes?

Yup, let's do that. Thanks a lot @alex1702 for putting this up!

Issue #1062, opened 2021-08-16
Can't install on Ubuntu 20.04 LTS due to docker-compose version


arctica:
### Version

21.8.0

### Steps to Reproduce

Tried to upgrade from 20.12.1 to 21.8.0 but the install.sh script fails when checking the minimum docker-compose version and drops the -h output into the console.

Checking the source of install/check-minimum-requirements.sh seems to indicate that Sentry requires version 1.28 but Ubuntu only provides 1.25 (even in latest non-LTS releases). See https://packages.ubuntu.com/focal/docker-compose

This change seems to have been introduced recently as 21.7 still required docker-compose 1.24.1

### Expected Result

Able to upgrade/install Sentry on Ubuntu 20.04 LTS

### Actual Result

root@sentry:~/onpremise-21.8.0# ./install.sh
▶ Parsing command line ...

▶ Setting up error handling ...

▶ Checking minimum requirements ...
Define and run multi-container applications with Docker.

Usage:
  docker-compose [-f <arg>...] [options] [COMMAND] [ARGS...]
  docker-compose -h|--help

Options:
  -f, --file FILE             Specify an alternate compose file
                              (default: docker-compose.yml)
  -p, --project-name NAME     Specify an alternate project name
                              (default: directory name)
  --verbose                   Show more output
  --log-level LEVEL           Set log level (DEBUG, INFO, WARNING, ERROR, CRITICAL)
  --no-ansi                   Do not print ANSI control characters
  -v, --version               Print version and exit
  -H, --host HOST             Daemon socket to connect to

  --tls                       Use TLS; implied by --tlsverify
  --tlscacert CA_PATH         Trust certs signed only by this CA
  --tlscert CLIENT_CERT_PATH  Path to TLS certificate file
  --tlskey TLS_KEY_PATH       Path to TLS key file
  --tlsverify                 Use TLS and verify the remote
  --skip-hostname-check       Don't check the daemon's hostname against the
                              name specified in the client certificate
  --project-directory PATH    Specify an alternate working directory
                              (default: the path of the Compose file)
  --compatibility             If set, Compose will attempt to convert keys
                              in v3 files to their non-Swarm equivalent
  --env-file PATH             Specify an alternate environment file

Commands:
  build              Build or rebuild services
  bundle             Generate a Docker bundle from the Compose file
  config             Validate and view the Compose file
  create             Create services
  down               Stop and remove containers, networks, images, and volumes
  events             Receive real time events from containers
[...]

PR #1063, opened 2021-08-16
Complexifyingly relax MIN_COMPOSE_VERSION

chadwhitacre:
Possible solution to #1062.


BYK (changes requested):
As discussed in https://github.com/getsentry/onpremise/issues/1062#issuecomment-900265740 it is not as simple to going back to an older version and altering the ANSI setting 🙂

PR #1066, opened 2021-08-17
ci: Test with the required minimum docker-compose

BYK:
Refs #1061, refs #1062, refs #1063.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
       - "release/**"
   pull_request:
 env:
-  DOCKER_COMPOSE_VERSION: 1.29.2
+  DOCKER_COMPOSE_VERSION: 1.28.0
 defaults:
   run:
     shell: bash
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,8 +8,6 @@ on:
       - "master"
       - "release/**"
   pull_request:
-env:
-  DOCKER_COMPOSE_VERSION: 1.28.0
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
       - "release/**"
   pull_request:
 env:
-  DOCKER_COMPOSE_VERSION: 1.29.2
+  DOCKER_COMPOSE_VERSION: 1.28.0
```
chadwhitacre, line 12: How about a comment here and in `check-minimum-requirements.sh` that we want to keep these two values in sync?

https://github.com/getsentry/onpremise/blob/47c4a1df299fffb8340160d9633c05a6c8f75adc/install/check-minimum-requirements.sh#L4

chadwhitacre (changes requested):
Comments plz to avoid confusion in the future.

.github/workflows/test.yml
chadwhitacre: Sure, I'm fine to DRY up if you want.

.github/workflows/test.yml
```diff
@@ -29,15 +27,16 @@ jobs:
...
-          sudo mv docker-compose /usr/local/bin
-
       - name: Checkout
         uses: actions/checkout@v2
+  
+      - name: Pin docker-compose
+        run: |
+          COMPOSE_PATH=/usr/local/bin/docker-compose
+          source ./install/_min-requirements.sh
+          sudo rm $COMPOSE_PATH
```
chadwhitacre, line 37: Why is the `rm` necessary if we sudo the curl call?

chadwhitacre: I guess `curl` itself will complain. Could use a redirect but I guess this is fine, too.

chadwhitacre (approved):

.github/workflows/test.yml
BYK: 🤷🏻‍♂️

PR #1068, opened 2021-08-18
fix(requirements): Make compose version check bw-compatible

BYK:
Refs #1062.

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -8,7 +8,8 @@ MIN_CPU_HARD=2
 MIN_CPU_SOFT=4
 
 DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
-COMPOSE_VERSION=$($dc --version | sed 's/docker-compose version \(.\{1,\}\),.*/\1/')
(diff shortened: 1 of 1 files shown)


install/check-minimum-requirements.sh
```diff
@@ -8,7 +8,8 @@ MIN_CPU_HARD=2
 MIN_CPU_SOFT=4
 
 DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
-COMPOSE_VERSION=$($dc --version | sed 's/docker-compose version \(.\{1,\}\),.*/\1/')
+# Do NOT use $dc instead of `docker-compose` below as older versions don't support certain options and fail
+COMPOSE_VERSION=$(docker-compose --version | sed 's/docker-compose version \(.\{1,\}\),.*/\1/')
```
chadwhitacre, line 12: Does the "`docker-compose` → `docker compose`" warning/notice leak through here? If so, do we want to suppress it, perhaps by redirecting stderr?

BYK: Nope:

```shell
$ docker-compose --version
docker-compose version 1.29.2, build 5becea4c
```

chadwhitacre (approved):

PR #1076, opened 2021-08-25
feat(smtp): Add hostname to SMTP

### LINKED ISSUE #1045: Add a hostname to the container to avoid emails getting rejected
### Version

any

### Steps to Reproduce

Send email

### Expected Result

Mail is delivered

### Actual Result

```
Jul 30 07:11:18 mail postfix/smtpd[8831]: NOQUEUE: reject: RCPT from xxx.xxx.net[xx.xxx.xxx.xxx]: 504 5.5.2 <0391926cb167>: Helo command rejected: need fully-qualified hostname; from=<xxx@xxx.xxx> to=<williamdes@wdes.fr> proto=ESMTP helo=<0391926cb167>
```

BYK:
Fixes #1045.

### DIFF
--- a/.env
+++ b/.env
@@ -3,6 +3,8 @@ SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
 SENTRY_BIND=9000
+Set SENTRY_MAIL_FROM to a valid FQDN to be able to send emails!
+# SENTRY_MAIL_FROM=
 SENTRY_IMAGE=getsentry/sentry:nightly
 SNUBA_IMAGE=getsentry/snuba:nightly
 RELAY_IMAGE=getsentry/relay:nightly
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,7 @@ symbolicator/config.yml
 geoip/GeoIP.conf
 geoip/*.mmdb
 geoip/.geoipupdate.lock
+mailname
 
 # wal2json download
 postgres/wal2json
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -63,6 +63,7 @@ x-sentry-defaults: &sentry_defaults
     # Leaving the value empty to just pass whatever is set
     # on the host system (or in the .env file)
     SENTRY_EVENT_RETENTION_DAYS:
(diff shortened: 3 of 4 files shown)


BYK:
Will also update docs on https://develop.sentry.dev/self-hosted/email/

.env
```diff
@@ -3,8 +3,10 @@ SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
 SENTRY_BIND=9000
+Set SENTRY_MAIL_FROM to a valid FQDN to be able to send emails!
+# SENTRY_MAIL_FROM=
```
williamdes, line 7:
```suggestion
# Set SENTRY_MAIL_FROM to a valid FQDN to be able to send emails!
# SENTRY_MAIL_FROM=
```

williamdes (approved):
LGTM
With such a name, in docs, you need to make it very clear this is not to be filled by an email but rather than with the server name (FQDN) that is the one filled in the reverse DNS PTR.

Example:
- To find out the IP `dig sentry-server.company.tld +short` > IP
- Use the IP to see to what it reverse resolves `dig -x $IP +short` > result

Adding a disclaimer that people NEED to add the server to their SPF field would be great, nobody seems to care enough about it..

Use the $result to fill the value

This will do that the chain is resolvable
email>smtp>helo $result>spf check>name resolves the ip that we are receiving mail from>spf ok (I over simplified)

BYK:
@williamdes thanks a lot for the review! I'm okay renaming the variable, any suggestions? Should we add `_HOST` or `_FQDN` at the end?

Regarding docs, would you like to give it a shot yourself as you seem to be more knowledgeable than myself regarding mailing? 😄

williamdes:
> @williamdes thanks a lot for the review! I'm okay renaming the variable, any suggestions? Should we add `_HOST` or `_FQDN` at the end?
> 
> Regarding docs, would you like to give it a shot yourself as you seem to be more knowledgeable than myself regarding mailing? 😄

Sure, I would love to get all Sentry installations perfectly email compatible, let me know where I can contribute


_HOST would be perfect a I am not sure everyone knows what a FQDN is

Maybe _HOSTNAME but I am not sure where the python setting is actually used
Could you post me links on where is the source code the setting is finally used to evaluate the impacts?

PR #1116, opened 2021-10-12
feat: Support docker compose CLI

### LINKED ISSUE #962: Support docker compose CLI.
### Summary

There is a next generation docker compose CLI. https://github.com/docker/compose-cli

It's built-in in Docker Desktop (for mac and windows), and will coming soon for Linux. (For now, Linux users can install it manually)

The new compose CLI is a native program that write in GoLang, it won't need python anymore, and will built-in in docker release.

### Motivation

I think the install script should check if the `docker compose` command available, fallback to `docker-compose` only if `docker compose` is invalid.

And don't forget the version check, `docker compose` have a new version number that restart from 1.0.

EricsonMacedo:
Check if docker compose v2, CLI, is available and get semantic version
from it, or fallback to get semantic version out of docker-compose v1
when checking minimum requirements during install.sh script

Fixes #962

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -118,7 +118,7 @@ done
 echo "${_endgroup}"
 
 echo "${_group}Ensure cleanup crons are working ..."
-$dc ps | grep -q -E "-cleanup\s+running\s+|-cleanup_.+\s+Up\s+"
+$dc ps | grep -q -E "\-cleanup\s+running\s+|\-cleanup_.+\s+Up\s+"
 echo "${_endgroup}"
 
 echo "${_group}Test custom CAs work ..."


PMExtra:
Please note that: I met some trouble in `docker compose` with sentry-onpremise.
The first, the install output format will be wrong. They cannot wrap line correctly.
Then when I you used `docker compose` to make a clean install, I cannot make any interaction with the initialization script (It won't take any response however, just like without `-i / --interactive` flag.

I have tested in several weeks ago with my own patch, with `Docker version 20.10.8` and `Docker Compose version v2.0.0-rc2` in a SSH session of CentOS 8.

Issue #1133, opened 2021-10-31
install.sh hangs during superuser creation under Docker Compose v2


mowangjuanzi:
# Workaround

With Docker Compose < 2.2.1(?):

```
./install.sh --no-user-prompt
docker compose run --rm web createuser
```

Docker Compose >= v2.2.1(?) is [known to partially work](https://github.com/docker/compose/issues/8908#issuecomment-993022808).

# Update

Fixed in https://github.com/docker/compose/pull/9035! 💃 Awaiting a new Docker Compose [release](https://github.com/docker/compose/releases/).

# Original Report

Docker version 20.10.10, build b485636

Docker Compose version v2.0.1

### Version

21.10.0

### Steps to Reproduce

sudo ./install.sh

### Expected Result

Install Success

### Actual Result

Running hooks in /etc/ca-certificates/update.d...
done.
02:45:10 [WARNING] sentry.utils.geo: Error opening GeoIP database: /geoip/GeoLite2-City.mmdb
02:45:10 [WARNING] sentry.utils.geo: Error opening GeoIP database in Rust: /geoip/GeoLite2-City.mmdb
02:45:12 [INFO] sentry.plugins.github: apps-not-configured
Running migrations for default

                                                                               

Project Options: N/A% |                                        | ETA:  --:--:--
                                                                               

Project Options: 100% |#########################################| Time: 0:00:00
Operations to perform:
  Apply all migrations: admin, auth, contenttypes, nodestore, sentry, sessions, sites, social_auth
Running migrations:
  Applying sentry.0240_grouphistory_index... OK
  Applying sentry.0241_grouphistory_null_actor... OK
  Applying sentry.0242_delete_removed_plugin_data... OK
  Applying sentry.0243_delete_visualstudio_repo_data... OK

Would you like to create a user account now? [Y/n]: 

At this time, the terminal does not respond. There is no choice but to close the terminal. After closing, the log output is as follows

read /dev/stdin: input/output error

chadwhitacre:
D'oh! I'm able to repro. Will work on debugging ...

chadwhitacre:
Hrm, something funky with in [`setup-and-migrate-database.sh`](https://github.com/getsentry/onpremise/blob/8606348c020d5907e165ae9d0ac4ea695cb992fa/install/set-up-and-migrate-database.sh). Workaround is:

```
SKIP_USER_PROMPT=1 ./install.sh
```

chadwhitacre:
[...]

PR #1134, opened 2021-11-01
Fix #1079 - bug in reset.sh

### LINKED ISSUE #1079: reset.sh breaking after passing to install.sh during check-minimum-requirements.sh line 3: file not found
### Version

nightly

### Steps to Reproduce

on an existing onpremise setup run ./reset.sh

after the call to install.sh check-minimum-requirements breaks at line 3: with the call to _min_requirements.sh not found.

seems while running in reset.sh and then calling install.sh the working directory changes to onpremise; however if running strictly in install.sh the working directory is onpremise/install

image

### Expected Result

reset.sh and then install.sh should flow correctly

### Actual Result

exits with check-minimum-requirements.sh: line 3: /root/onpremise/_min-requirements.sh: No such file or directory

image

chadwhitacre:
We need $0 to evaluate as install.sh in order for check-requirements to work.

### DIFF
--- a/reset.sh
+++ b/reset.sh
@@ -59,4 +59,4 @@ if [ -n "$version" ]; then
 fi
 
 # Install.
-exec ./install.sh
+./install.sh


reset.sh
```diff
@@ -59,4 +59,4 @@ if [ -n "$version" ]; then
 fi
 
 # Install.
-exec ./install.sh
+./install.sh
```
armenzg, line 62: What was the original intent of using `exec`?

This piece of logic might be usable instead of `source "$(dirname $0)/_min-requirements.sh"`:
https://github.com/getsentry/sentry/blob/6776229fbde4ca8cd52884d369a1b498a43dc6b4/scripts/do.sh#L5-L10

chadwhitacre: I didn't have a strong reason to use `exec` originally. I did it that way because conceptually we are done with the one script so figured we'd hand off to the next, didn't think through the nuances.

I've started trying to whittle down a better repro in order to understand better what's going on and what the best solution is. That's turning out to take more time than I'd like, though. This `reset.sh` script is hacky and mostly for me. The real bug I want to focus on is #1133.

You okay to merge as-is?

armenzg (approved):

Issue #796, opened 2021-01-06
Rename onpremise to self-hosted


chadwhitacre:
In the [docs](https://develop.sentry.dev/self-hosted/) we're using "self-hosted" to name what we're doing here. For branding purposes I would like to standardize on that and stop referring to "onpremise."

(Short URL, fwiw: https://git.io/JSDlX)

### Todo

- [x] rename repo
- [x] [code](https://github.com/search?q=org%3Agetsentry+onpremise&type=code)
  - [x] easy
      - [x] `sentry.io`—archived repo
      - [x] `sentry-release`—moribund repo, don't care
      - [x] https://github.com/getsentry/labels/commit/9d35accca4af33a52ab2cbb91b65a33adc077deb
      - [x] https://github.com/getsentry/deps/commit/5962f6b603d73095b9e72ca68e417f15a7094d78
      - [x] https://github.com/getsentry/demo/pull/11
      - [x] https://github.com/getsentry/develop/pull/469
      - [x] https://github.com/getsentry/eng-pipes/pull/273
      - [x] https://github.com/getsentry/etl/pull/671
      - [x] https://github.com/getsentry/lookml/pull/318
      - [x] https://github.com/getsentry/publish/pull/670
      - [x] https://github.com/getsentry/relay/pull/1136
      - [x] https://github.com/getsentry/self-hosted/pull/1169
      - [x] `sentry`
          - [x] https://github.com/getsentry/sentry/pull/30482
          - [x] https://github.com/getsentry/sentry/pull/30483
          - [x] https://github.com/getsentry/sentry/pull/30486
      - [x] https://github.com/getsentry/sentry-dart/pull/642
      - [x] https://github.com/getsentry/sentry-docs/pull/4437
      - [x] https://github.com/getsentry/snuba/pull/2240
      - [x] https://github.com/getsentry/static-sites/pull/773
  - [x] [hard](https://github.com/getsentry/self-hosted/issues/796#issuecomment-990355748)
    - [x] backend
      - [x] https://github.com/getsentry/getsentry/pull/6841
      - [x] https://github.com/getsentry/single-tenant/pull/429
      - [x] https://github.com/getsentry/sentry/pull/30973
      - [x] https://github.com/getsentry/sentry/pull/31066
    - [x] frontend
      - [x] add `isSelfHosted` in `sentry` (done in [#30973](https://github.com/getsentry/sentry/pull/30973))
      - [x] https://github.com/getsentry/sentry/pull/31065
[...]

PR #1169, opened 2021-11-24
Rename onpremise to self-hosted

chadwhitacre:
https://github.com/getsentry/self-hosted/issues/796

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
-COMPOSE_PROJECT_NAME=sentry_self_hosted
+COMPOSE_PROJECT_NAME=sentry-self-hosted
 SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
-COMPOSE_PROJECT_NAME=sentry_onpremise
+COMPOSE_PROJECT_NAME=sentry_self_hosted
 SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -357,6 +357,7 @@ services:
       web:
         <<: *depends_on-healthy
 volumes:
+  # These store application data that should persist across restarts.
   sentry-data:
     external: true
   sentry-postgres:
@@ -371,6 +372,8 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
+
+  # These store ephemeral data that needn't persist across restarts.
   sentry-secrets:
   sentry-smtp:
   sentry-zookeeper-log:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ x-healthcheck-defaults: &healthcheck_defaults
   # Related issues:
   # https://github.com/moby/moby/issues/39102
   # https://github.com/moby/moby/issues/39388
-  # https://github.com/getsentry/onpremise/issues/1000
+  # https://github.com/getsentry/self-hosted/issues/1000
   interval: 30s
   timeout: 5s
   retries: 5
@@ -252,7 +252,7 @@ services:
     command: subscriptions --auto-offset-reset=latest --consumer-group=snuba-transactions-subscriptions-consumers --topic=events --result-topic=transactions-subscription-results --dataset=transactions --commit-log-topic=snuba-commit-log --commit-log-group=transactions_group --delay-seconds=60 --schedule-ttl=60
   snuba-cleanup:
     <<: *snuba_defaults
-    image: snuba-cleanup-onpremise-local
+    image: snuba-cleanup-self-hosted-local
     build:
       context: ./cron
       args:
@@ -260,7 +260,7 @@ services:
     command: '"*/5 * * * * gosu snuba snuba cleanup --storage errors --dry-run False"'
   snuba-transactions-cleanup:
     <<: *snuba_defaults
-    image: snuba-cleanup-onpremise-local
+    image: snuba-cleanup-self-hosted-local
     build:
       context: ./cron
       args:
@@ -278,7 +278,7 @@ services:
     command: run -c /etc/symbolicator/config.yml
   symbolicator-cleanup:
     <<: *restart_policy
-    image: symbolicator-cleanup-onpremise-local
+    image: symbolicator-cleanup-self-hosted-local
     build:
       context: ./cron
       args:
@@ -317,7 +317,7 @@ services:
     command: run query-subscription-consumer --commit-batch-size 1 --topic transactions-subscription-results
   sentry-cleanup:
     <<: *sentry_defaults
-    image: sentry-cleanup-onpremise-local
+    image: sentry-cleanup-self-hosted-local
     build:
       context: ./cron
       args:
--- a/install/turn-things-off.sh
+++ b/install/turn-things-off.sh
@@ -5,9 +5,6 @@ if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
   $dc rm -fsv $($dc config --services | grep -v -E '^(nginx|relay)$')
 else
   # Clean up old stuff and ensure nothing is working while we install/update
-  # This is for older versions of on-premise:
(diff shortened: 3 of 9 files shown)


install/turn-things-off.sh
```diff
@@ -5,9 +5,6 @@ if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
   $dc rm -fsv $($dc config --services | grep -v -E '^(nginx|relay)$')
 else
   # Clean up old stuff and ensure nothing is working while we install/update
-  # This is for older versions of on-premise:
-  $dc -p onpremise down -t $STOP_TIMEOUT --rmi local --remove-orphans
-  # This is for newer versions
```
chadwhitacre, line 10: The older/newer distinction is from https://github.com/getsentry/self-hosted/pull/316 which is almost 2 years old. I think we can do away with the older version by now.

chadwhitacre:
Mind giving this a once-over when you get a chance @aminvakil? Pretty sure there are no breaking changes here but would value a second opinion.

chadwhitacre:
@BYK Happy for your input as well if you're up for it. 🙇‍♂️ ☺️

.env
```diff
@@ -1,4 +1,4 @@
-COMPOSE_PROJECT_NAME=sentry_onpremise
+COMPOSE_PROJECT_NAME=sentry_self_hosted
```
chadwhitacre, line 1: Will this trip people up if they are upgrading an existing installation?

docker-compose.yml
```diff
@@ -252,15 +252,15 @@ services:
     command: subscriptions --auto-offset-reset=latest --consumer-group=snuba-transactions-subscriptions-consumers --topic=events --result-topic=transactions-subscription-results --dataset=transactions --commit-log-topic=snuba-commit-log --commit-log-group=transactions_group --delay-seconds=60 --schedule-ttl=60
   snuba-cleanup:
     <<: *snuba_defaults
-    image: snuba-cleanup-onpremise-local
+    image: snuba-cleanup-self-hosted-local
```
chadwhitacre, line 255: This might leave some extraneous images hanging around people's environments but I don't think that's any great problem?

.env
BYK: yes, this is likely a breaking change.

docker-compose.yml
BYK: This should be fine. They can be removed with `--remove-orphans`

install/turn-things-off.sh
BYK: Yes as people need to go through some other hard-stops anyway.

BYK (approved):

.env
chadwhitacre: In that case it seems as if we should keep this as `sentry_onpremise` for now, put a note in the changelog for 21.12.0 and possibly a warning inline on `install.sh`, and then make the breaking change in 22.1.0.

aminvakil: It would be better to output this on `install.sh`, users may probably not read the changelog on every release.

aminvakil:
@chadwhitacre Thanks for doing this!

.env
chadwhitacre: One mention in our codebase:

https://github.com/getsentry/self-hosted/blob/459323354ef8f370828636cba29e36e90613525f/install/wrap-up.sh#L9-L10

... and then:

https://docs.docker.com/compose/reference/envvars/#compose_project_name

chadwhitacre: Okay so how do I test this? I want to launch self-hosted using the old value, then update to the new value, then relaunch self-hosted.

chadwhitacre: Why do we use `--network="${COMPOSE_PROJECT_NAME}_default"` in `wrap-up.sh`? 🤔 

> Your app’s network is given a name based on the “project name”, which is based on the name of the directory it lives in. You can override the project name with either the `--project-name` flag or the `COMPOSE_PROJECT_NAME` environment variable.

https://docs.docker.com/compose/networking/

👍

```
$ docker network ls
$ docker network ls
NETWORK ID     NAME                       DRIVER    SCOPE
36ae04d3f89c   bridge                     bridge    local
b14723786ff7   host                       host      local
71053c5e70ae   none                       null      local
b5034c179e6c   sentry_onpremise_default   bridge    local
$
```

chadwhitacre: Afaict containers, networks, and volumes all use project name. 🤔

chadwhitacre: [No volume rename](https://github.com/moby/moby/issues/31154), [best option is manual](https://serverfault.com/a/1022691):

```
docker volume create --name <new_volume>
docker run --rm -it -v <old_volume>:/from:ro -v <new_volume>:/to alpine \
    ash -c "cd /from ; cp -av . /to"
docker volume rm <old_volume>
```

chadwhitacre: Proposal:

1. Containers—Don't care. They end up with orphans that they can clean up on their own time.
2. Network—Don't care. Spin down old, spin up new. They can clean up if they want, same as (1).
3. Volumes—This is the tricky one. The user needs to take some action in order for this to work, either they need to rename their volumes manually as just described, or they need to override `COMPOSE_PROJECT_NAME` in `.env.custom` to retain the old `sentry_onpremise` value. It sorta feels to me like adapting to config changes like this are a normal part of the upgrade process, so we are within bounds to put this burden on users. Thoughts?

aminvakil: How about not changing current `onpremise` volumes and create `self-hosted` in case there aren't already volumes created?

chadwhitacre: Good news! We create most of our volumes by [calling `docker volume create` ourselves](https://github.com/getsentry/self-hosted/blob/459323354ef8f370828636cba29e36e90613525f/install/create-docker-volumes.sh), marking them [`external`](https://docs.docker.com/compose/compose-file/compose-file-v3/#external) [in `docker-compose.yml`](https://github.com/getsentry/self-hosted/blob/459323354ef8f370828636cba29e36e90613525f/docker-compose.yml#L360-L373):

> If set to `true`, specifies that this volume has been created outside of Compose. `docker-compose up` does not attempt to create it, and raises an error if it doesn’t exist.

It [seems](https://forums.docker.com/t/docker-compose-prepends-directory-name-to-named-volumes/32835/2) that `docker-compose` prepends the project name when creating a volume, but `docker volume create` does not. 💃 

```
$ docker volume list
DRIVER    VOLUME NAME
$ ./install.sh --no-user-prompt
[...]
$ docker volume list
DRIVER    VOLUME NAME
local     f1d00624e9452e7075f7224f8eeb08744005c64938d6d2d92812d463926429b8
local     sentry-clickhouse
local     sentry-data
local     sentry-kafka
local     sentry-postgres
local     sentry-redis
local     sentry-symbolicator
local     sentry-zookeeper
local     sentry_onpremise_sentry-clickhouse-log
local     sentry_onpremise_sentry-kafka-log
local     sentry_onpremise_sentry-secrets
local     sentry_onpremise_sentry-smtp
local     sentry_onpremise_sentry-smtp-log
local     sentry_onpremise_sentry-zookeeper-log
$
```

Sooooooooo ... can we get away with orphaning the log/secrets/smtp volumes? 🧐

chadwhitacre: The external volumes are ancient: 5d064c2224aa73a45ee0e29063be85c8810f4707.
The non-external are also ancient: #291.

chadwhitacre: I read #291 as saying that these are ephemeral volumes that we can recreate on down/up (the reason for #291 was to avoid littering with random volumes, not to address an operational issue with non-persistent volumes).

I've confirmed that the prefixed volumes are ephemeral with this method:

1. `install.sh` on `master`
2. `dcm up`
3. [Populate](https://github.com/getsentry/self-hosted/issues/1182#issue-1068600193) a couple events.
    1.  `export SENTRY_DSN="foo"`
    2. `sentry-cli send-event -m "[Random-Error] $(echo $RANDOM | md5sum | head -c 20; echo;)" > /dev/null 2>&1 &`
4. `dcm down`
5. switch to this branch
6. `install.sh`
7. `dcm up`
8. Did events persist? Yes.
9. Populate a few more events. Did it work? Yes.

chadwhitacre: Okay! I'm ready to say that this change is non-breaking. I've pushed 2573e87fa4738dc7ef50423c67333d14dbb1f073 but otherwise I am ready to go with this PR. Last chance to stop me @BYK @aminvakil et al. :)

ayozemr:
Waiting for this for a new deployment! Thanks for the work guys

chadwhitacre:
Wait no more, @ayozemr! :) Now that this is merged, would you mind trying it out and reporting back whether it breaks anything for you? Some orphaned containers/volumes/network are expected, but the upgrade is should otherwise be seamless.

Issue #1132, opened 2021-10-30
Make Docker Compose version check more forgiving


mowangjuanzi:
### Version

21.10.0

### Steps to Reproduce

1. https://github.com/getsentry/onpremise.git
2. sudo SENTRY_IMAGE=getsentry/sentry:21.10.0 ./install.sh

### Expected Result

Install Success.

### Actual Result

▶ Parsing command line ...

▶ Setting up error handling ...

▶ Checking minimum requirements ...
FAIL: Expected minimum docker-compose version to be 1.28.0 but found Docker Compose version v2.0.1

mowangjuanzi:
this is a docker-compose question ?

chadwhitacre:
@mowangjuanzi We recently started supporting both v1 and v2 of Docker Compose, it's possible you've found a bug in how we implement that. Can you debug this further? Look in `install/check-requirements.sh` and see how the environment variables are ending up in there.

mowangjuanzi:
```bash
$ echo $(docker-compose --version | sed 's/docker-compose version \(.\{1,\}\),.*/\1/')
Docker Compose version v2.0.1
```

chadwhitacre:
> i not found `install/check-requirements.sh`
>
> but i found `install/check-minimum-requirements.sh`

Sorry, that's what I meant, yes. :)

### Architectural Sidebar

```
sh install/check-minimum-requirements.sh
```

The non-executable files in `install` are designed to be sourced by `install.sh`, not run directly, hence the following output:

```
install/check-minimum-requirements.sh: 3: source: not found
```

### Debugging the Problem

Going over your initial report more carefully, I see a clue:

```
Expected minimum docker-compose version to be 1.28.0 but found Docker Compose version v2.0.1
```

I hypothesize that the `v` in `v2.0.1` is causing this check to fail:

https://github.com/getsentry/onpremise/blob/c16ae024dc3e8259b03679f950e80e1cbfb5dae8/install/check-minimum-requirements.sh#L24

And I hypothesize that the `v` is there because you're hitting this line:

https://github.com/getsentry/onpremise/blob/c16ae024dc3e8259b03679f950e80e1cbfb5dae8/install/check-minimum-requirements.sh#L11

Instead of this one:

https://github.com/getsentry/onpremise/blob/c16ae024dc3e8259b03679f950e80e1cbfb5dae8/install/check-minimum-requirements.sh#L8

Can you confirm these hypotheses, please? :)

[...]

Issue #1176, opened 2021-11-30
Error checking minimum Docker Compose version


thenets:
### Version

master branch

### Steps to Reproduce

1. Clone the repo
2. Run the installer:
   - `./install.sh`
3. Got error at "Checking minimum requirements"
   - Probably is an issue related to a regex to check the min version that now has a `v` at the beginning.

### Expected Result

Installation should work, cause I have the minimum required version.

#### Error message

```
[luiz@fedora self-hosted]$ ./install.sh 
▶ Parsing command line ...

▶ Setting up error handling ...

▶ Checking minimum requirements ...
FAIL: Expected minimum docker-compose version to be 1.28.0 but found Docker Compose version v2.0.1
```

#### Docker Compose version

I installed from the official GitHub project:
- https://github.com/docker/compose

```
[luiz@fedora self-hosted]$ docker-compose -v
Docker Compose version v2.0.1
```

aminvakil:
I thought something has been changed since https://github.com/getsentry/self-hosted/pull/1116, but this takes `v` prefix into account before checking already.

BYK:
So the issue here is we expect version `< 2` when `docker-compose` exists and use `docker compose` if we get an error or something. Can you confirm that you cannot run `docker compose` but can run `docker-compose`?

aminvakil:
Maybe the issue here is `/usr/local/bin/docker-compose` which is `< 2` and `/usr/libexec/docker/cli-plugins/docker-compose` which is `> 2` both exist.

BYK:
@aminvakil yeah that might be the issue but in that case, we should just be using `docker compose`. At least what I think the code does.

aminvakil:
> @aminvakil yeah that might be the issue but in that case, we should just be using `docker compose`. At least what I think the code does.

Yes, it appears so.

I just downgraded `docker-compose` to 2.0.1 and tested this:
```
docker compose version --short | sed 's/v\{0,1\}\(.\{1,\}\)/\1/'
2.0.1
```
So probably `2.0.1` specifically does not have problem.

chadwhitacre:
This is a dupe of https://github.com/getsentry/self-hosted/issues/1132, right?

chadwhitacre:
Gonna take that as a yes. 😛

PR #1177, opened 2021-11-30
Fix Docker Compose min ver

thenets:
Fixes:
- https://github.com/getsentry/self-hosted/issues/1132

That's my attempt to fix the issue.
Probably exist much better ways to solve it. This PR can be at least an inspiration to someone else to fix the problem.

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -9,6 +9,12 @@ if docker compose version &>/dev/null; then
 else
   # Do NOT use $dc instead of `docker-compose` below as older versions don't support certain options and fail
   COMPOSE_VERSION=$(docker-compose --version | sed 's/docker-compose version \(.\{1,\}\),.*/\1/')
+
+  # Another check if is the newer version with additional prefix
+  if [[ $COMPOSE_VERSION =~ v ]]; then
+    # Extract only the version number
+    COMPOSE_VERSION=$(echo ${COMPOSE_VERSION} | rev | cut -d 'v' -f 1 | rev)
+  fi
 fi
 RAM_AVAILABLE_IN_DOCKER=$(docker run --rm busybox free -m 2>/dev/null | awk '/Mem/ {print $2}');
 CPU_AVAILABLE_IN_DOCKER=$(docker run --rm busybox nproc --all);


install/check-minimum-requirements.sh
```diff
@@ -9,6 +9,12 @@ if docker compose version &>/dev/null; then
 else
   # Do NOT use $dc instead of `docker-compose` below as older versions don't support certain options and fail
   COMPOSE_VERSION=$(docker-compose --version | sed 's/docker-compose version \(.\{1,\}\),.*/\1/')
+
+  # Another check if is the newer version with additional prefix
+  if [[ $COMPOSE_VERSION =~ v ]]; then
+    # Extract only the version number
+    COMPOSE_VERSION=$(echo ${COMPOSE_VERSION} | rev | cut -d 'v' -f 1 | rev)
```
aminvakil, line 16: What do you think about this?
```suggestion
    COMPOSE_VERSION=$(echo ${COMPOSE_VERSION} | sed 's/^v//')
```

aminvakil (commented):
Just for simplicty.

install/check-minimum-requirements.sh
thenets: @aminvakil doesn't quite work. The content is `version v2.0.1`. That `v` is the last one.

The full output is:
```bash
[luiz@fedora self-hosted]$ docker-compose -v
Docker Compose version v2.0.1
```

So my strategy was ignoring 100% of what's before `v`, because doesn't matter if they change the prefix again in the future.

What you can do if you want a cleaner approach is to ignore the entire `Docker Compose version v` prefix.

aminvakil: Per my comment on the issue about #1116 , the scripts checks for `docker-compose` version using `docker compose version --short` which returns:
```
$ docker compose version --short
2.2.0
```
They changed lots of things even after `2.0.0`, I cannot test using versions > `2` < `2.2.0` right now, but it drops the `v` prefix already here:
`COMPOSE_VERSION=$(docker compose version --short | sed 's/v\{0,1\}\(.\{1,\}\)/\1/')`

thenets: @aminvakil Well, this is not the current code. The `docker compose version [--short]` command repeats 3 times:

https://github.com/getsentry/self-hosted/pull/1177#diff-1b2cb955f165ce00dbc0caa8eb2f9f54032b41f6f52d144f9a1313f53a8b5e1fR10-R11

I'll try to change all of them.

aminvakil: Please read https://github.com/getsentry/self-hosted/pull/1116#issuecomment-944900571 completely. As once mentioned to me :) setting `docker-compose` is not mandatory anymore.
And we don't want to change the behaviour which was working for `docker-compose` < 2.
Also I'm not sure if `--short` existed for `docker-compose` < 2.

thenets: @aminvakil sorry. I just read it in the entire issue and I don't know what to do.
The first solution solved the problem and my second attempt broke some behavior.
I also don't know the `docker-compose` internals.

I'll wait for some recommendations. 

My recommendation still is the first one. Fix the problem without compromising the initial asserts.

thenets: Reverted again to the initial PR state.

BYK:
See my comment on https://github.com/getsentry/self-hosted/issues/1176#issuecomment-982553697

install/check-minimum-requirements.sh
chadwhitacre: The challenge is that, while `docker compose` is guaranteed(?) to be v2, `docker-compose` can be either v1 or v2. I have some ideas ... going to post back on #1132.

thenets: Hello @chadwhitacre. Now I understood the problem but I can't understand why this current PR state isn't enough.
It solves all scenarios that I can think about.

chadwhitacre:
Thanks for the PR, @thenets. How does https://github.com/getsentry/self-hosted/pull/1179 look to you?

install/check-minimum-requirements.sh
chadwhitacre: @thenets I feel bad for turning down your contribution ... how about we merge this one _and_ #1179 (which refactors `check-minimum-requirements.sh` for legibility, and also addresses additional instances of `docker-compose` throughout the codebase)? 😁

thenets: @chadwhitacre no problem at all. As I said in the beginning, probably exists some better way to solve the problem. You can discard my PR, no problem.
But I still think that until now, my PR is the only solution for my specific case without compromising existent scenarios.

Issue #1196, opened 2021-12-13
log4j vulnerability (CVE-2021-44228) - not vulnerable


aminvakil:
# Official Statement

https://blog.sentry.io/2021/12/15/sentrys-response-to-log4j-vulnerability-cve-2021-44228

# Update

The two Java-based components used in self-hosted Sentry are Kafka and Zookeeper. Both of those projects have concluded that they are not vulnerable, therefore our best understanding is that [**self-hosted Sentry is not vulnerable to log4shell**](https://github.com/getsentry/self-hosted/issues/1196#issuecomment-993626831) and does not need to be patched.

# Original

I have a self-hosted installation of Sentry 21.9.0 (sorry for not updating it).

I know that some components of Sentry are written in Java. Today I have read the news about a critical log4j vulnerability (https://www.lunasec.io/docs/blog/log4j-zero-day/, CVE-2021-44228). Are Sentry installations using scripts from this repository affected? If so, what's the proper way to apply a fix or workaround?

_Originally posted by @patrakov in https://github.com/getsentry/self-hosted/issues/1175#issuecomment-990674501_

aminvakil:
@patrakov Sentry does not have any Java code in it. That said it uses Kafka and Zookeeper so it'd be better to look there. Self-hosted uses [Confluent](https://docs.confluent.io/platform/current/release-notes/index.html) packages so you may check them and submit a PR to upgrade the packages if you see something.

_Originally posted by @BYK in https://github.com/getsentry/self-hosted/issues/1175#issuecomment-990726002_

aminvakil:
Would also love confirmation on the vulernability as @patrakov mentioned, since it does look like some old log4j versions (1.x) were used when I searched a self-hosted server for log4j with `sudo find / -iname '*log4j*.jar'`. But I'm not very familiar with Docker/Kafka so wouldn't know where to check next. Thank you 😊

_Originally posted by @bryceadams in https://github.com/getsentry/self-hosted/issues/1175#issuecomment-992083915_

acherifi:
It seems that log4j 1.x is impacted only if some features are enabled  [check here](https://github.com/apache/logging-log4j2/pull/608#issuecomment-990494126).
On the latest Self Hosted Sentry kafka container 
```
root@cf46a1c8a75e:/# find / -name 'log4j*'
/usr/share/java/kafka/log4j-1.2.17.jar
/etc/kafka/log4j.properties
/etc/confluent/docker/log4j.properties.template
```
[...]

PR #1203, opened 2021-12-15
Self-hosted Sentry is not vulnerable to log4shell

chadwhitacre:
This is a dummy PR to work around limitations with the automated changelog system we are using. Please see https://github.com/getsentry/self-hosted/issues/1196 for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,3 +58,4 @@ Where you replace `83b1380` with the sha you want to use.
 
 [build-status-image]: https://github.com/getsentry/self-hosted/workflows/test/badge.svg
 [build-status-url]: https://git.io/JUYkh
+


README.md
```diff
@@ -58,3 +58,4 @@ Where you replace `83b1380` with the sha you want to use.
 
 [build-status-image]: https://github.com/getsentry/self-hosted/workflows/test/badge.svg
 [build-status-url]: https://git.io/JUYkh
+
```
BYK, line 61: 🤣 Why not literally add a disclaimer in the readme: Sentry is not vulnerable to log4shell

chadwhitacre: Because nobody is going to care in a month? 0, 1, infinity.

PR #1179, opened 2021-11-30
Do more migrating to Docker Compose v2

chadwhitacre:
The big thing here is a fix for #1132. I did a grep for `docker-compose` and cleaned up the rest of the instances. The one question I had is in CI tests ... as I read it we still need `docker-compose` there so we can test v1, which we do still want to support for now.

PR #1209, opened 2021-12-20
Fix reset.sh docker-compose call

matti:


BYK (approved):

aminvakil:
This has been changed in #1179 .

aminvakil (changes requested):
This should detect docker-compose version, some users (I won't say few or many as I have no idea) have migrated to `docker compose > 2.0.0` and definitely not all of them have symlinked `/usr/local/bin/docker-compose` to plugin to make it compatible with this change.

chadwhitacre:
Seems like it's time to teach `reset.sh` about `install/_lib.sh`, or factor out the `dc` stuff and have both source that.

https://github.com/getsentry/self-hosted/blob/28e413c5dd04e1170ee8201dd5e11b9b3165b65d/install/_lib.sh#L33-L36

aminvakil:
I don't think of any particular reason not use the whole `install/_lib.sh`, but maybe later we want to add something to it, which breaks the `reset.sh`, so it's better to just copy `docker-compose` detection from it IMO.

matti:
fyi I am *not* going to do this extra changes, I just fixed the most obvious bug.

aminvakil:
> This should detect docker-compose version, some users (I won't say few or many as I have no idea) have migrated to `docker compose > 2.0.0` and definitely not all of them have symlinked `/usr/local/bin/docker-compose` to plugin to make it compatible with this change.

@matti What part of this you disagree with?

matti:
I don't diasagree with anything - I just wont do anything else for this PR, feel free to do as much as you want.

matti:
I've seen this about 100 times - somebody makes a PR for one thing and somebody in the project starts listing completely unrelated backlog and the the thing itself stays broken

aminvakil:
Fine, thanks, I'll take it from here in #1215 .

PR #1213, opened 2021-12-20
Upgrade to Postgres 10

### LINKED ISSUE #1097: Upgrade PostgreSQL from 9.6 to 10.x
# Original Report

### Problem Statement

Hello

I have been using Sentry 21.9.0 for production use for a python based api , and I am using AWS RDS Postgres 9.6 as the database server for the sentry application. I have been getting a warning message from AWS regarding the EOL date of Postgres9.6 and its deprecation is nearby but your docker-compose and other tools still list postgres:9.6 as the db server. Are there any plans for upgrade to a newer version of postgres ?

### Solution Brainstorm

Here's an excerpt :  https://forums.aws.amazon.com/ann.jspa?annID=8499

# Update from Sentry

1. Hosting on Postgres [10](https://github.com/getsentry/self-hosted/issues/1097#issuecomment-1004341364), [11](https://github.com/getsentry/sentry/issues/12673#issuecomment-528274149), [12](https://github.com/getsentry/sentry/issues/12673#issuecomment-889610630), [13](https://github.com/getsentry/self-hosted/issues/1097#issuecomment-1069723230) and [14](https://github.com/getsentry/self-hosted/issues/1097#issuecomment-1102981785) seems to work, but should be considered experimental and second-class until we upgrade SaaS away from Postgres 9.6, which is not currently prioritized or scheduled.
1. We run `sentry` CI against Postgres 9.6 and 14.
1. We do not officially support Sentry on RDS, but it's an obvious and reasonable customization, so we'd rather not break it if we don't have to.

chadwhitacre:
Is it as simple as this? It can't be as simple as this, can it?

Closes #1097.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,7 +118,7 @@ services:
         hard: 10032
   postgres:
     <<: *restart_policy
-    image: "postgres:9.6"
+    image: "postgres:10.19"
     healthcheck:
       <<: *healthcheck_defaults
       # Using default user "postgres" from sentry/sentry.conf.example.py or value of POSTGRES_USER if provided


aminvakil:
I'm not sure if sticking to a specific minor version of postgres has more upsides than downsides.

https://www.postgresql.org/support/versioning/:

We always recommend that all users run the latest available minor release for whatever major version is in use.

Major versions usually change the internal format of system tables and data files. These changes are often complex, so we do not maintain backward compatibility of all stored data. A dump/reload of the database or use of the pg_upgrade module is required for major upgrades. We also recommend reading the upgrading section of the major version you are planning to upgrade to. You can upgrade from one major version to another without upgrading to intervening versions, but we recommend reading the release notes of all intervening major versions prior to doing so.

Upgrading to a minor release does not normally require a dump and restore; you can stop the database server, install the updated binaries, and restart the server. For some releases, manual changes may be required to complete the upgrade, so always read the release notes before upgrading.

While upgrading will always contain some level of risk, PostgreSQL minor releases fix only frequently-encountered bugs, security issues, and data corruption problems to reduce the risk associated with upgrading. **For minor releases, the community considers not upgrading to be riskier than upgrading.**

glensc (commented):
This is definitely not simple as that, see:
- https://github.com/getsentry/self-hosted/blob/21.12.0/install/upgrade-postgres.sh

I have fresh experience from last week (sentry upgrade) of 9.5->9.6 postgresql not starting up unless the upgrade 9.5 to 9.6 was performed, and since you are performing major upgrade here (9 to 10), the upgrade script is a must!

chadwhitacre:
Gosh. I'm actually surprised to see that we actually do automate the 9.5 to 9.6 upgrade. I guess we only do that if they're using the docker compose service container for PostgreSQL. For any more complex setup the upgrade is in the operator's able hands. :)

I'm closing this for now as the conclusion at https://github.com/getsentry/self-hosted/issues/1097#issuecomment-999160152 is to hold off on upgrading to Postgres 10 here in self-hosted until SaaS is upgraded.

Thanks for surfacing that script, @glensc!

BYK:
>Gosh. I'm actually surprised to see that we actually do automate the 9.5 to 9.6 upgrade. I guess we only do that if they're using the docker compose service container for PostgreSQL. 

You know who to thank for that 😝

chadwhitacre:
https://github.com/getsentry/self-hosted/pull/283 🌮 👏 @BYK 😁 🙇

Issue #1171, opened 2021-11-24
CI flakes under "Ensure cleanup crons are working"


chadwhitacre:
Seeing integration tests fail quite frequently, rerun sometimes works. Noticing a few failure modes, here is one (cf. #1170):

https://github.com/getsentry/self-hosted/runs/4317568879?check_suite_focus=true

```
Ensure cleanup crons are working ...
  + grep -q -E -e '\-cleanup\s+running\s+' -e '\-cleanup[_-].+\s+Up\s+'
  + docker compose --ansi never --env-file /home/runner/work/self-hosted/self-hosted/.env ps
  ++ teardown ERR 121
  ++ '[' 0 -eq 1 ']'
  ++ DID_TEAR_DOWN=1
  ++ '[' ERR '!=' EXIT ']'
  ++ echo 'An error occurred, caught SIGERR on line 121'
  An error occurred, caught SIGERR on line 121
  ++ echo 'Tearing down ...'
  Tearing down ...
  ++ rm /tmp/tmp.9Qws9UthXM
  ++ echo Done.
  Done.
  + teardown EXIT 1
  + '[' 1 -eq 1 ']'
  + return 0
  Error: Process completed with exit code 255.
  ##[debug]Finishing: Integration Test
```

Not consistent between 1.28 and 2.0.1.

Line:

https://github.com/getsentry/self-hosted/blob/da8f4903e4cfe47ca4fa5bf68265b62f795c60a0/_integration-test/run.sh#L123

chadwhitacre:
https://github.com/getsentry/self-hosted/runs/4372607126?check_suite_focus=true#step:4:3336

aminvakil:
I was trying to understand what has happened in times prior to regression starting to happen yesterday and could it be because of https://github.com/getsentry/self-hosted/pull/1119?

I couldn't understand why this should fail sometimes, maybe others have some idea?

chadwhitacre:
Another on #1289:

https://github.com/getsentry/self-hosted/runs/5058157966?check_suite_focus=true#step:4:3617

My hunch is that cleanup runs but the test hits either before or after.

chadwhitacre:
https://github.com/getsentry/self-hosted/runs/5116818258?check_suite_focus=true#step:4:3641

aminvakil:
https://github.com/getsentry/self-hosted/runs/5163491240?check_suite_focus=true

On https://github.com/getsentry/self-hosted/commit/e0a8ae commit.

[...]

Issue #418, opened 2020-03-24
Config permissions: !! Configuration error: `system.secret-key` MUST be set.


PhillWide:
## Important Details
UPDATED after I figured out the actual issue

I'm running the Docker on-premise commit a7c5a32e36282f513e6dd5f9461f50114cb0597d

## Description

I get the error in issue name if I rewrite the config file with permissions 640 instead of 644.

## Steps to Reproduce

1. ./install.sh
2. change permissions for the config
3. ./install.sh

### What you expected to happen

I'm not really sure why it requires the permissions but at least it would be nice to get correct error.
Additionally it would be nice if the install added the secret key when system.secret-key is not specified at all instead of only with !!changeme!!.

stayallive:
You need to do it the other way around, it looks like you ran `./install.sh` before modifying the `config.yml`, the install script builds docker images and copies the configuration into them. So to update the config you need to re-run `install.sh` to activate those changes.

PhillWide:
Thanks! I must have been the combination of handling the installation as before with the new script.

PhillWide:
Ah, the issue returns if I use CI=yes which I need for my ansible.

PhillWide:
I had to remove the configs and cleanup docker cache to fix the issue

PhillWide:
Ok, so it ended up being a permissions issue: for some reason, I need 644 permissions for the configuration file

PR #1222, opened 2021-12-25
Install: setup umask

glensc:
Setup default [`umask`](https://en.wikipedia.org/wiki/Umask), otherwise files created with permissions that containers later can't read.

```
    !! Configuration error: ConfigurationError("PermissionError: [Errno 13]
    Unable to load configuration file (Permission denied):
    '/etc/sentry/sentry.conf.py'",)
```

To reproduce, try:

```
git checkout 21.6.3 
umask 0007
./install.sh
```

```
root@docker2 sentry/21.6.3# ls -l sentry/sentry.conf.py
-rw-rw---- 1 root root 7.4K Dec 25 03:18 sentry/sentry.conf.py
```

also known to be broken with `21.5.0` and `9.1.2`, i.e the versions noted in hard stops:
- https://develop.sentry.dev/self-hosted/releases/#hard-stops

so, perhaps carry the fix to those branches as well

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,6 +5,8 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+umask 002
+
 source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other things
 
 source dc-detect-version.sh


aminvakil:
Do you think it's good to show users that their `umask` is changing temporarily with executing `install.sh`?

Or at least to the ones who have disabled reading for `world` and therefore without this fix sentry won't get installed properly.

glensc:
It doesn't change `umask` temporarily, it changes for the install script. `umask` outside the script says as it is.

aminvakil:
> It doesn't change `umask` temporarily, it changes for the install script. `umask` outside the script says as it is.

Sorry for not being clear, by temporarily I mean the same, do you think it's good to tell users this is changing created files by `install.sh`?

glensc:
No I don't think users need to be aware of any of this.

ideally, this should be solved (also) by doing proper chown/chmod after file creation. or even better at each install.sh invocation.

aminvakil:
> No I don't think users need to be aware of any of this.
> 
> ideally, this should be solved (also) by doing proper chown/chmod after file creation. or even better at each install.sh invocation.

That can be taken care of in another PR.

This LGTM for now.

aminvakil (approved):

glensc:
An older issue, that have been likely caused by this problem: https://github.com/getsentry/self-hosted/issues/418

install.sh
```diff
@@ -5,6 +5,8 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+umask 002
```
chadwhitacre, line 8: Refreshing my memory:

> new directories will have permission 775 and new files permission of 664

Right?

glensc: uhm, yes...

easy tester:

```
➜ (umask 002; mkdir dir1; touch file1; ls -ld dir1 file1)
drwxrwxr-x 2 glen staff 64 dets  28 23:35 dir1/
-rw-rw-r-- 1 glen staff  0 dets  28 23:35 file1
```

chadwhitacre:
Rerunning CI to dodge #1171 ... set to auto-merge once green.

Thanks @glensc @aminvakil. :)

> so, perhaps carry the fix to those branches as well

Hrm, that would be a first. We don't really ever backport to older versions, especially now that we're on CalVer. Why are you hitting this (why is this the first we're hearing about it?)? Presumably you were able to work around the problem, yes?

glensc:
@chadwhitacre it's not common to change umask. also, I'm guessing people who hit any problem just do `chmod -R 777 .` and go on instead of taking time figuring out the problem cause and reporting it :D

glensc:
@chadwhitacre as for the problem workaround, I ran `chmod a+rX -R .` and re-run the install script. and that as many times as many new files the install script creates. rather annoying and time-waster, but was able to finish install/upgrade like this.

and once I knew it's about umask, the workaround is to set umask once prior script run.

BYK (approved):

chadwhitacre:
> it's not common to change umask

So a user has to have set a umask in order to hit this issue? I guess that is indicated in your repro steps:

```
git checkout 21.6.3 
umask 0007
./install.sh
```

In that case I guess the user could also manually set a more relaxed umask before running 21.5.0 or 9.1.2. With these workarounds available, I don't see a strong enough case for issuing our first x.x.x.1 releases.

chadwhitacre (approved):

chadwhitacre:
Thanks for taking the time to figure out the root cause and report it, @glensc! 😁

PR #1241, opened 2022-01-05
Take some actions to avoid unhealthy containers

### LINKED ISSUE #1178: 'ERROR: for snuba-api Container "ee666f7f2cdd" is unhealthy.' during install.sh
### Version

21.11.0

### Steps to Reproduce

1. Run ./install.sh

### Expected Result

Installation succeed

### Actual Result

▶ Bootstrapping and migrating Snuba ...
Creating sentry_onpremise_clickhouse_1 ... 
Creating sentry_onpremise_zookeeper_1  ... 
Creating sentry_onpremise_redis_1      ... 
Creating sentry_onpremise_redis_1      ... done
Creating sentry_onpremise_zookeeper_1  ... done
Creating sentry_onpremise_clickhouse_1 ... done
Creating sentry_onpremise_kafka_1      ... 
Creating sentry_onpremise_kafka_1      ... done

ERROR: for snuba-api  Container "ee666f7f2cdd" is unhealthy.
Encountered errors while bringing up the project.
An error occurred, caught SIGERR on line 3
Cleaning up...

Can anyone help me with this ?

chadwhitacre:
Closes #1178

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,9 +11,9 @@ source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other 
 
 source parse-cli.sh
 source dc-detect-version.sh
+source error-handling.sh
 source turn-things-off.sh
 source check-latest-commit.sh
-source error-handling.sh
 source check-minimum-requirements.sh
 source create-docker-volumes.sh
 source ensure-files-from-examples.sh
--- a/install.sh
+++ b/install.sh
@@ -9,8 +9,9 @@ umask 002
 
 source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other things
 
-source dc-detect-version.sh
 source parse-cli.sh
+source dc-detect-version.sh
+source turn-things-off.sh
 source check-latest-commit.sh
 source error-handling.sh
 source check-minimum-requirements.sh
@@ -20,7 +21,6 @@ source generate-secret-key.sh
 source replace-tsdb.sh
 source update-docker-images.sh
 source build-docker-images.sh
-source turn-things-off.sh
 source set-up-zookeeper.sh
 source install-wal2json.sh
 source bootstrap-snuba.sh
--- a/install.sh
+++ b/install.sh
@@ -1,20 +1,23 @@
 #!/usr/bin/env bash
 set -e
+
+# Pre-pre-flight? 🤷
 if [[ -n "$MSYSTEM" ]]; then
   echo "Seems like you are using an MSYS2-based system (such as Git Bash) which is not supported. Please use WSL instead.";
   exit 1
 fi
 
-umask 002
-
 source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other things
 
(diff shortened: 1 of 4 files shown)


install/dc-detect-version.sh
```diff
@@ -1,3 +1,5 @@
+echo "${_group}Initializing Docker Compose ..."
```
chadwhitacre, line 1: sidecar

install.sh
```diff
@@ -9,8 +9,9 @@ umask 002
 
 source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other things
 
-source dc-detect-version.sh
 source parse-cli.sh
+source dc-detect-version.sh
```
chadwhitacre, line 13: sidecar to move this below CLI parsing, conceptually I think CLI parsing should always be first

aminvakil:
Are we sure changing all healthcheck timeouts for all users to 60s is a good idea?

My question is are there any situations in self-hosted where if something does not respond in 5 seconds, it's broken?

install.sh
```diff
@@ -9,8 +9,9 @@ umask 002
...
+source turn-things-off.sh
```
BYK, line 14: `error-handling` should come as early as possible?

chadwhitacre:
> My question is are there any situations in self-hosted where if something does not respond in 5 seconds, it's broken?

If it's broken after 5s it will still be broken after 60s, right?

aminvakil:
> > My question is are there any situations in self-hosted where if something does not respond in 5 seconds, it's broken?
> 
> If it's broken after 5s it will still be broken after 60s, right?

Well if something is wrong and it responds in 30s or so, it's broken in my POV.

install.sh
chadwhitacre: Fair point. It depends on `$dc` so moving just after that in c48efe096ad614e7860b48f7e6bc5a187d4ecede.

aminvakil:
> > > My question is are there any situations in self-hosted where if something does not respond in 5 seconds, it's broken?
> > 
> > 
> > If it's broken after 5s it will still be broken after 60s, right?
> 
> Well if something is wrong and it responds in 30s or so, it's broken in my POV.

Like I said in https://github.com/getsentry/self-hosted/issues/1178#issuecomment-999424878 best if it could be set in `.env` IMO, for this PR changing default to 60s is better than current state though.

install.sh
chadwhitacre: Hmm ... thought through this a little further and created 2b9a179270763a2919304fc0337fb2c307a499c0. Eh?

chadwhitacre:
> best if it could be set in `.env` IMO, for this PR changing default to 60s is better than current state though.

Okay, let's run with this for a while and if we find that it's causing problems in the other direction we can look at making it configurable, would rather no introduce additional complexity if we don't have to.

chadwhitacre:
@aminvakil I mean, if "someone else" wants to make a PR for configurability ... ;)

Issue #1283, opened 2022-01-31
Reset script doesn't work to unbound variable _group


lovetodream:
### Version

22.1.0

### Steps to Reproduce

1. Download the latest version
2. Execute the install script (not really necessary)
3. Execute the reset script

### Expected Result

The reset script should execute successfully.

### Actual Result

install/dc-detect-version.sh: line 1: _group: unbound variable

chadwhitacre:
Good catch @lovetodream, thanks. Should be fixed in https://github.com/getsentry/self-hosted/pull/1285.

chadwhitacre:
lol oops seeing #1284 now :P

lovetodream:
I am glad to help. 
Next time I'll highlight that I already made a PR.

lovetodream:
Fixed in #1284

PR #1273, opened 2022-01-24
Debug CI

chadwhitacre:
CI is failing on master. Why?

#1272

PR #1272, opened 2022-01-24
Revert "Add CI compose version 1.29.2 / 2.2.3"

chadwhitacre:
Reverts getsentry/self-hosted#1251

We're seeing a failure on an internal CI build that seems to be caused by this, still investigating ...

PR #1284, opened 2022-01-31
fix: unbound variable _group in reset/dc-detect-version script (#1283)

lovetodream:
This would fix the reset script. [#1283 –_ed._]

Because the `dc-detect-version` is only used in the reset script, I think it is fine to add the _group and _endgroup declarations directly.


chadwhitacre (approved):

chadwhitacre:
ref #1283

chadwhitacre:
Blech #1273. 😖

lovetodream:
> Blech #1273. 😖

I Just noticed that #1273 was closed. Does this mean that this can be merged?

chadwhitacre:
> Does this mean that this can be merged?

Almost. #1273 turned into #1289, which (it turns out) basically dupes #1290, which revives #1251, which we had to revert in #1272 because it effed up CI over in `sentry`. Kind of a tangle right now but I think we are making progress. 😁

Hopefully we can get back here soon ... 🤞

chadwhitacre:
Still getting a garbled relay config. 🤔

https://github.com/getsentry/self-hosted/runs/5009812723?check_suite_focus=true#step:5:121

aminvakil:
As https://github.com/getsentry/self-hosted/pull/1315 has been merged, should this fix also reflect GCB too?

chadwhitacre:
Yeah, we need to rebase/update.

install/dc-detect-version.sh
```diff
@@ -1,3 +1,11 @@
+if [ "${GITHUB_ACTIONS:-}" = "true" ]; then
```
aminvakil, line 1: > Yeah, we need to rebase/update.

@chadwhitacre I mean is there some environment variable or something like that in GCB which should be checked like here?

chadwhitacre: This can stay as-is, only GHA parses the `::` in a special way (right?), other envs can all use the arrow output.

aminvakil (approved):
LGTM unless squashing commits is not an option and this should be rebased without a merge commit.

chadwhitacre:
Thanks for your patience @lovetodream. 🙏

PR #1289, opened 2022-02-02
Clean up relay credentials generation

chadwhitacre:
This is a follow-up to #1273, and it should get CI back on track on `master`.

### DIFF
--- a/install/ensure-relay-credentials.sh
+++ b/install/ensure-relay-credentials.sh
@@ -1,0 +1,34 @@
+echo "${_group}Ensuring Relay credentials ..."
+
+RELAY_CONFIG_YML="../relay/config.yml"
+RELAY_CREDENTIALS_JSON="../relay/credentials.json"
+
+ensure_file_from_example $RELAY_CONFIG_YML
+
+if [[ -f "$RELAY_CREDENTIALS_JSON" ]]; then
+  echo "$RELAY_CREDENTIALS_JSON already exists, skipped creation."
+else
+
+  # There are a couple gotchas here:
+  #
+  # 1. We need to use a tmp file because if we redirect output directly to
+  #    credentials.json, then the shell will create an empty file that relay
+  #    will then try to read from (regardless of options such as --stdout or
+  #    --overwrite) and fail because it is empty.
+  #
+  # 2. We need to use -T to avoid additional garbage output cluttering
+  #    credentials.json under Docker Compose 1.x and 2.2.3+. Note that the
+  #    long opt --no-tty doesn't exist in Docker Compose 1.
+
+  creds="$dcr --no-deps -T relay credentials"
+  $creds generate --stdout > "$RELAY_CREDENTIALS_JSON".tmp
+  mv "$RELAY_CREDENTIALS_JSON".tmp "$RELAY_CREDENTIALS_JSON"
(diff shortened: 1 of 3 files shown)


install/ensure-relay-credentials.sh
```diff
@@ -0,0 +1,33 @@
...
+  #
+  # 1. We need to use a tmp file because if we redirect output directly to
+  #    credentials.json, then the shell will create an empty file that relay
+  #    will then try to read from (regardless of options such as --stdout or
+  #    --overwrite) and fail because it is empty.
+  #
+  # 2. We need to use -T to avoid additional garbage output cluttering
+  #    credentials.json under Docker Compose 1.x and 2.2.3+.
+
+  creds="$dcr --no-deps -T relay credentials"
```
chadwhitacre, line 22: @aminvakil It looks like there is no `--no-tty` option in Docker Compose 1, only in 2. That could explain why merging #1251 caused CI to start failing.

```
$ docker-compose run -T --no-deps relay credentials show      
Creating sentry-self-hosted_relay_run ... done
Credentials:
  relay id: ed128877-ceb6-47e0-8a4e-e3df4bed162f
  public key: EyGAkE-3wBlBQuSbKJLsXdhQWEMsaZIDSnOCvnVE5zo
$
$
$ docker-compose run --no-tty --no-deps relay credentials show
Run a one-off command on a service.

For example:

    $ docker-compose run web python manage.py shell

By default, linked services will be started, unless they are already
running. If you do not want to start linked services, use
`docker-compose run --no-deps SERVICE COMMAND [ARGS...]`.

Usage:
    run [options] [-v VOLUME...] [-p PORT...] [-e KEY=VAL...] [-l KEY=VALUE...] [--]
        SERVICE [COMMAND] [ARGS...]

Options:
    -d, --detach          Detached mode: Run container in the background, print
                          new container name.
    --name NAME           Assign a name to the container
    --entrypoint CMD      Override the entrypoint of the image.
    -e KEY=VAL            Set an environment variable (can be used multiple times)
    -l, --label KEY=VAL   Add or override a label (can be used multiple times)
    -u, --user=""         Run as specified username or uid
    --no-deps             Don't start linked services.
    --rm                  Remove container after run. Ignored in detached mode.
    -p, --publish=[]      Publish a container's port(s) to the host
    --service-ports       Run command with the service's ports enabled and mapped
                          to the host.
    --use-aliases         Use the service's network aliases in the network(s) the
                          container connects to.
    -v, --volume=[]       Bind mount a volume (default [])
    -T                    Disable pseudo-tty allocation. By default `docker-compose run`
                          allocates a TTY.
    -w, --workdir=""      Working directory inside the container
$
$
$ docker-compose --version
docker-compose version 1.29.2, build 5becea4c
$
```

chadwhitacre: Bah ... but in that case why was CI green on #1251? 🤔

chadwhitacre:
Hrm, red CI ... 🧐

BYK (approved):

aminvakil (approved):

chadwhitacre:
I reticketed the CI flakiness as https://github.com/getsentry/self-hosted/issues/1293. Rebased here and aiming to merge ...

PR #1290, opened 2022-02-02
Add CI compose version 1.29.2 / 2.0.1 / 2.2.3

chadwhitacre:
Redo of #1251 after #1272.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,11 @@ jobs:
         include:
           - compose_version: '1.28.0'
             compose_path: '/usr/local/bin'
+          - compose_version: '1.29.2'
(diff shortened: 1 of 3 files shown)


chadwhitacre:
This and #1289 are pointed in basically the same direction, over there there is more refactoring but no addition of additional Compose versions to the CI matrix. I think if this is green we merge this first to keep close to the original plan with #1251, and then rebase over there to pick up the refactor as well.

BYK (approved):

aminvakil (approved):

chadwhitacre:
Flakiness reticketed as https://github.com/getsentry/self-hosted/issues/1294. Merging ...

PR #1308, opened 2022-02-07
Update set-up-and-migrate-database.sh

drmrbrewer:
It's `docker-compose` not `docker compose`

### DIFF
--- a/install/set-up-and-migrate-database.sh
+++ b/install/set-up-and-migrate-database.sh
@@ -6,7 +6,7 @@ if [[ -n "${CI:-}" || "${SKIP_USER_PROMPT:-0}" == 1 ]]; then
   echo "Did not prompt for user creation due to non-interactive shell."
   echo "Run the following command to create one yourself (recommended):"
   echo ""
-  echo "  docker-compose run --rm web createuser"
+  echo "  $dc_base run --rm web createuser"
   echo ""
 else
   $dcr web upgrade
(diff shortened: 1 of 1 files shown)


BYK:
It _is_ `docker compose` since a while ago 🙂: https://docs.docker.com/compose/cli-command/

drmrbrewer:
@BYK Oh 🤦‍♂️ but I get:

    # docker compose run --rm web createuser
    unknown flag: --rm
    See 'docker --help'.

BYK:
@drmrbrewer we do have some version detection code around which then defines a `$dc_base` here: https://github.com/getsentry/self-hosted/blob/master/install/dc-detect-version.sh

I think you can make this file use `$dc_base` instead of `docker compose` directly and that should take case of the issue?

BYK (approved):

BYK:
@chadwhitacre nudge

chadwhitacre (approved):

chadwhitacre (approved):

chadwhitacre:
Thanks @drmrbrewer @BYK! :)

chadwhitacre:
I like this PR so much I approved it twice. 🤩

drmrbrewer:
I am sure this will go down as a significant day in the history of self-hosted sentry.  This was the equivalent of laying down a single mosaic tile (or maybe two!) in a luxurious Roman palace.  Well, Rome wasn't built in a day, as they say!

Issue #1294, opened 2022-02-03
CI flakes with "Renaming a container with the same name as its current name"


chadwhitacre:
Reticketing from https://github.com/getsentry/self-hosted/pull/1290#issuecomment-1028902426. [Log link](https://github.com/getsentry/self-hosted/runs/5040186466?check_suite_focus=true#step:4:2459)

Happens at:

https://github.com/getsentry/self-hosted/blob/da8f4903e4cfe47ca4fa5bf68265b62f795c60a0/_integration-test/run.sh#L45

Possibly https://github.com/docker/compose/issues/6704? But why? And how to avoid? 🤔

chadwhitacre:
https://github.com/getsentry/self-hosted/runs/5100619223?check_suite_focus=true#step:4:2609

chadwhitacre:
https://github.com/getsentry/self-hosted/runs/5517601020?check_suite_focus=true#step:4:2698

aminvakil:
https://github.com/getsentry/self-hosted/runs/5522537135?check_suite_focus=true#step:4:2360

chadwhitacre:
From https://github.com/getsentry/self-hosted/pull/1365#issuecomment-1067074853 ...

> multiple versions in same docker instance

@glensc Do you understand what's going on here and how to fix it?

chadwhitacre:
@BYK Confirming ... did you intend to fix this ticket by setting `COMPOSE_PROJECT_NAME` in CI?

BYK:
>@BYK Confirming ... did you intend to fix this ticket by setting COMPOSE_PROJECT_NAME in CI?

No I did not. I intended to fix another one with "container has the same name" or something like that and that one seems to be fixed.

BYK:
>Possibly https://github.com/docker/compose/issues/6704? But why? And how to avoid? 🤔

Adding `--no-recreate` to that command should fix it. Also adding `--wait` may make the `waiting for Sentry to be up` step redundant.

aminvakil:
https://github.com/getsentry/self-hosted/runs/5620519447?check_suite_focus=true#step:4:2170

aminvakil:
https://github.com/getsentry/self-hosted/runs/6953387418?check_suite_focus=true#step:4:3973

aminvakil:
https://github.com/getsentry/self-hosted/runs/7158445726?check_suite_focus=true#step:4:3806

aminvakil:
[...]

Issue #1293, opened 2022-02-03
CI flakes with "Cannot restart container"


chadwhitacre:
Reticketing from https://github.com/getsentry/self-hosted/pull/1289#issuecomment-1028149691. [Log link](https://github.com/getsentry/self-hosted/runs/5040239745?check_suite_focus=true#step:4:2101)

Happens at:

https://github.com/getsentry/self-hosted/blob/da8f4903e4cfe47ca4fa5bf68265b62f795c60a0/install/wrap-up.sh#L6

[Old SO](https://stackoverflow.com/questions/31365827/cannot-stop-or-restart-a-docker-container)

Considering adding a `|| true`?

chadwhitacre:
https://github.com/getsentry/self-hosted/runs/5116648918?check_suite_focus=true#step:4:2096

aminvakil:
https://github.com/getsentry/self-hosted/runs/7267174574?check_suite_focus=true#step:4:3809

PR #1315, opened 2022-02-08
Add a cloudbuild.yaml for GCB

### LINKED ISSUE #1311: Build on GCB during CI
We broke CI in `sentry` and `snuba` with #1251 and #1289, due to (not fully understood) discrepancies in the environment between local dev, `self-hosted` CI, and `sentry`/`snuba` CI. The proposal here is to trigger a GCB runner similar to the ones in `sentry` and `snuba` to run as part of `self-hosted` CI to ensure that we're not breaking CI over there with our changes here. This should cover us in the future even if we don't fully understand the root cause for the problem(s) with #1251 and #1289.

This is an action item for [INC-98](https://getsentry.atlassian.net/browse/INC-98) (private).

chadwhitacre:
Closes #1311.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,57 @@
+steps:
+  - name: "gcr.io/$PROJECT_ID/docker-compose"
+    id: clone-and-configure 
+    entrypoint: "bash"
+    args:
+      - "-e"
+      - "-c"
+      - |
+        git clone https://github.com/getsentry/self-hosted.git
+        echo '{"version": "3.4", "networks":{"default":{"external":{"name":"cloudbuild"}}}}' > self-hosted/docker-compose.override.yml
+    timeout: 60s
+  - name: "gcr.io/$PROJECT_ID/docker-compose"
+    id: install 
+    waitFor:
+      - clone-and-configure
+    dir: self-hosted
+    entrypoint: "bash"
+    args:
+      - "-e"
+      - "-c"
+      - |
+        ./install.sh
+    timeout: 600s
+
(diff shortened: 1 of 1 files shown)


chadwhitacre:
GCB is green, slogging through #1171 #1293 #1294 ...

BYK:
Why not go the other way around and replace those GCB runs with GitHub Actions? You can now use OIDC for securely pushing images to Google Cloud: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-google-cloud-platform

Would be less maintenance and better DX as we all know nobody looks at (or understands) GCB logs 😀

aminvakil (approved):
If I have understood correctly, this would happen in a private environment where non-employed contributors without access cannot see the logs over there or even if it had been broken or not, but if this helps sentry guys, sure, why not?

chadwhitacre:
Gosh, with 2x the CI builds post-#1290 it's even harder to get a clean build. :-/

chadwhitacre:
> Why not go the other way around and replace those GCB runs with GitHub Actions?

That's not a bad idea, not sure I want to set the plow that deep though since it involves getting buy-in from other teams and this is an action item for a CI incident, I'd like to resolve sooner than later to prevent future bugs from taking down CI.

BYK:
Happy to give you a hand as the GCB config for Sentry itself is already public (along with Snuba and Relay). All I'd need from your end would be setting up OIDC which should be a breeze.

chadwhitacre:
Thanks @BYK. Merging this for now. I don't have bandwidth to drive GCB→GHA right now but if you start making PRs I will try to keep up. 😁

PR #1316, opened 2022-02-09
Link to SaaS

chadwhitacre:


README.md
```diff
@@ -1,7 +1,9 @@
-# Self-Hosted Sentry nightly
+# Self-Hosted Sentry
```
BYK, line 1: FYI this will be added back on the next release automatically.

chadwhitacre: [Interesting](https://github.com/getsentry/self-hosted/blob/22.1.0/scripts/bump-version.sh#L16). Thanks for the tip! :)

PR #1335, opened 2022-02-20
Proof of Concept: Use Vectorized Redpanda instead of Kafka + Zookeeper

skunkworker:
This is still a Proof of Concept, but I set out to see how difficult it would be to use Vectorize Redpanda instead of Kafka, so far this has worked without any real issues, and the overall cpu and memory requirements would be lower than they are now, with the advantage of removing Java from the application.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,7 @@ x-sentry-defaults: &sentry_defaults
   depends_on:
     redis:
       <<: *depends_on-healthy
-    redpanda:
+    kafka:
       <<: *depends_on-healthy
     # kafka:
     #   <<: *depends_on-healthy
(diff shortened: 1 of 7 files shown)


chadwhitacre:
Nice work, @skunkworker! I've shared this internally with our Ops team in case they have any interest in making the switch at some point. Until then this is great to have as an example for anyone willing to live on the edge. :)

chadwhitacre:
FYI

> I did not do an investigation on how different in terms of semantics kafka and redpanda are. Though, assuming we are not changing anything in prod, we do use some specific features of kafka to make the product work (like alerts depend on monotonic broker timestamp for messages. Remove that and they break). Redpanda may just not even work

skunkworker:
> FYI
> 
> > I did not do an investigation on how different in terms of semantics kafka and redpanda are. Though, assuming we are not changing anything in prod, we do use some specific features of kafka to make the product work (like alerts depend on monotonic broker timestamp for messages. Remove that and they break). Redpanda may just not even work

Interesting, do you have more details on these specific features? I'm sure I could share them over at the redpanda repo

chadwhitacre:
I guess that's a no I don't have more details. :P The internal thread kind of trailed off. I would expect the relevant code to be somewhere under https://github.com/getsentry/snuba.

PR #1340, opened 2022-02-22
Add some debug logging

chadwhitacre:
For #1171

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -120,7 +120,7 @@ done
 echo "${_endgroup}"
 
 echo "${_group}Ensure cleanup crons are working ..."
-$dc ps | grep cleanup  # to debug https://github.com/getsentry/self-hosted/issues/1171
+$dc ps -a | grep cleanup  # to debug https://github.com/getsentry/self-hosted/issues/1171
 $dc ps | grep -q -E -e '\-cleanup\s+running\s+' -e '\-cleanup[_-].+\s+Up\s+'
 echo "${_endgroup}"
 
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -120,6 +120,7 @@ done
 echo "${_endgroup}"
 
 echo "${_group}Ensure cleanup crons are working ..."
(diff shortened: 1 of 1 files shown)


BYK (approved):

_integration-test/run.sh
```diff
@@ -120,6 +120,7 @@ done
 echo "${_endgroup}"
 
 echo "${_group}Ensure cleanup crons are working ..."
+$dc ps | grep cleanup  # to debug https://github.com/getsentry/self-hosted/issues/1171
```
aminvakil, line 123: Why not try to see exited containers too while we're at it?
```suggestion
$dc ps -a | grep cleanup  # to debug https://github.com/getsentry/self-hosted/issues/1171
```

aminvakil (commented):
Now that we're just debugging I think we can have this.

We can discuss it later if the actual line should have `-a` flag or not later.

_integration-test/run.sh
chadwhitacre: Sounds good, we need to rerun CI anyway. ;)

chadwhitacre:
The logging [is working](https://github.com/getsentry/self-hosted/runs/5304727139?check_suite_focus=true#step:4:3569). :)

```
  sentry-self-hosted-sentry-cleanup-1                             "/entrypoint.sh '0 0…"   sentry-cleanup                             running             9000/tcp
  sentry-self-hosted-snuba-cleanup-1                              "/entrypoint.sh '*/5…"   snuba-cleanup                              running             1218/tcp
  sentry-self-hosted-snuba-transactions-cleanup-1                 "/entrypoint.sh '*/5…"   snuba-transactions-cleanup                 running             1218/tcp
  sentry-self-hosted-symbolicator-cleanup-1                       "/entrypoint.sh '55 …"   symbolicator-cleanup                       running             3021/tcp
```

chadwhitacre:
Okay so why does the grep not match those lines? Or is the call the feeds into grep spitting out something different? 🤔

aminvakil:
Closing and reopening, hoping that this time it will break :) and we can see our logging.

chadwhitacre:
And actually it's good to have the green case documented. Merging this so we can start getting debug on red cases ...

Issue #1341, opened 2022-02-22
CI flakes with "The operation was canceled."


chadwhitacre:
Integration test is timing out after 360 minutes:

https://github.com/getsentry/self-hosted/runs/5266047420?check_suite_focus=true#step:4:775

aminvakil:
I'm not sure after two runs , but it seems to happen with 1.28.0, 1.29.2 and 2.2.3 (2.0.1 is fine)
https://github.com/getsentry/self-hosted/runs/5300821867?check_suite_focus=true

chadwhitacre:
Is that because it's the first one to kick off? Is there some resource constraint, like the 2.0.1 job is grabbing a lock and not releasing it?

aminvakil:
https://github.com/getsentry/self-hosted/runs/5307522338?check_suite_focus=true
Again only 2.0.1 was fine.

aminvakil:
https://github.com/getsentry/self-hosted/runs/5311060854?check_suite_focus=true
from https://github.com/getsentry/self-hosted/pull/1347

aminvakil:
https://github.com/getsentry/self-hosted/runs/5343799600?check_suite_focus=true
OK, I was hoping if this was a temporary issue somehow with GHA, unfortunately it's not :(

aminvakil:
https://github.com/getsentry/self-hosted/runs/5461913213?check_suite_focus=true
Again, 2.0.1 is only working.

BYK:
[>https://github.com/getsentry/self-hosted/runs/5339631065?check_suite_focus=true#step:4:1207](https://github.com/getsentry/self-hosted/runs/5339631065?check_suite_focus=true#step:4:1207) shows the issue

BYK:
This is because https://github.com/getsentry/sentry/commit/53e38f42bcb01078a78169f4223894920e29276c - you should not have an interactive prompt when `CI=1` or [`--no-input`](https://github.com/getsentry/sentry/blob/b5bf8da514bef723e7e11940c869a9e203a1f86e/src/sentry/runner/commands/createuser.py#L56) is passed. 🤦🏻

chadwhitacre:
Reopening as fix was deemed "temporary."

chadwhitacre:
> This is because https://github.com/getsentry/sentry/commit/53e38f42bcb01078a78169f4223894920e29276c

Good find @BYK. 👍

PR #1365, opened 2022-03-08
Docker-Compose: Avoid setting hostname to ''

glensc:
The hostname is literally set to `''`:

Before:
```
# docker-compose exec smtp hostname
''
```

After:
```
# docker-compose exec smtp hostname
c531c19841de
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -93,7 +93,7 @@ services:
   smtp:
     <<: *restart_policy
     image: tianon/exim4
-    hostname: ${SENTRY_MAIL_HOST:-''}
+    hostname: ${SENTRY_MAIL_HOST:-}
     volumes:
       - "sentry-smtp:/var/spool/exim4"
       - "sentry-smtp-log:/var/log/exim4"


aminvakil (changes requested):
I think there was a reason this has been added in https://github.com/getsentry/self-hosted/pull/1116/commits/6c92774fae3a76df266cc035a7f5c7466c441b6f which was in https://github.com/getsentry/self-hosted/pull/1116.

Investigating further...

aminvakil:
I just tested your change with docker-compose 2.3.2 and after this, `snuba` container keeps restarting without any log.

glensc:
@aminvakil in any case, the used syntax is incorrect. it's already well known (to me at least) that docker compose doesn't like extra quoting. it's not shell script. for example if you place a variable with spaces to .env YOU MAU NOT QUOTE THE VALUE, or you get quotes in env value itself.

anyway, after my fix, if env is missing the valuer is set to empty (rather apostrhophes), and then default hostname will be set, which is generated name. as seen in example I provided

glensc:
I fail to see how "smtp" container "hostname" is related to "snuba" container :D

If you insist on quoting, then this might work (not tested):

```yml
  smtp:
    <<: *restart_policy
    image: tianon/exim4
    hostname: "${SENTRY_MAIL_HOST:-}"
```

but that's just yaml syntax, shouldn't affect the variables expansion.

aminvakil:
> I just tested your change with docker-compose 2.3.2 and after this, `snuba` container keeps restarting without any log.

Ok, reverting this PR change results in `sentry-self-hosted-snuba-subscription-consumer-transactions-1` container restarting over and over again.

aminvakil:
> I fail to see how "smtp" container "hostname" is related to "snuba" container :D
> 
> If you insist on quoting, then this might work (not tested):
> 
> ```yaml
>   smtp:
>     <<: *restart_policy
>     image: tianon/exim4
>     hostname: "${SENTRY_MAIL_HOST:-}"
> ```
> 
> but that's just yaml syntax, shouldn't affect the variables expansion.

Yeah, it was a mistake, it's not related. But I still want to understand why it has been added at the first place.

BYK:
GitHub did something to how they use Docker across matrix jobs and I think this is causing issues. We may want to scope the project name with the job name in docker-compose setups to avoid the issues. The failure on 1.29.2 seems related to that.

aminvakil:
@glensc As https://github.com/getsentry/self-hosted/commit/de81299bafe2e36e9660f31837569f93b5788da1 has been merged, can you please rebase your branch to retrigger the CI?

It should pass now.

glensc:
@aminvakil, you can do that yourself, really:

image

aminvakil:
@glensc Do you see "maintainer" here?

image

BYK (approved):

aminvakil (approved):
LGTM.

Closing and reopening to trigger the CI.

docker-compose.yml
```diff
@@ -93,7 +93,7 @@ services:
   smtp:
     <<: *restart_policy
     image: tianon/exim4
-    hostname: ${SENTRY_MAIL_HOST:-''}
+    hostname: ${SENTRY_MAIL_HOST:-}
```
chadwhitacre, line 96: lgtm

> Additionally when using the [2.1 file format](https://docs.docker.com/compose/compose-file/compose-versioning/#version-21), it is possible to provide inline default values using typical shell syntax:
>
> - `${VARIABLE:-default}` evaluates to `default` if `VARIABLE` is unset or empty in the environment.

https://docs.docker.com/compose/environment-variables/

I'll take your word for it on quoting, @glensc. :P

chadwhitacre (approved):

chadwhitacre:
Sadly needs another rebase once #1375 lands.

chadwhitacre:
I landed #1375. Can you rebase again @glensc? 🙏

glensc:
image

Why do you need me to do the rebases? Makes only useless ping-pongs!

docker-compose.yml
glensc: No need to take on anything, just see the proof in PR body.

glensc:
Build flackyness due to multiple versions in same docker instance still not resolved:

- https://github.com/getsentry/self-hosted/runs/5540753414?check_suite_focus=true

```
 Container self-hosted-0-nginx-1  Recreated
  Error response from daemon: Renaming a container with the same name as its current name
  ++ teardown ERR 45
```

and unrelated to this PR.

chadwhitacre:
> Build flakiness due to multiple versions in same docker instance still not resolved:

https://github.com/getsentry/self-hosted/issues/1294 ftr

> The fix didn't apparently work

de81299bafe2e36e9660f31837569f93b5788da1 was for #1341, not #1294. We have [several open CI issues](https://github.com/getsentry/self-hosted/issues?q=is%3Aissue+is%3Aopen+ci+flakes). 😞

glensc:
I thought this was to solve container name collisions:

> `COMPOSE_PROJECT_NAME: self-hosted-${{ strategy.job-index }}`

glensc:
hey. CI green. merge fast before someone requests rebase! :D

chadwhitacre:
Close one! Thanks @glensc! 😁

PR #1370, opened 2022-03-10
ci: Temporary fix for interactive prompt on createuser

### LINKED ISSUE #1341: CI flakes with "The operation was canceled."

BYK:
Fixes #1341 temporarily.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
           - compose_version: "v2.2.3"
             compose_path: "/usr/local/lib/docker/cli-plugins"
     env:
-      COMPOSE_PROJECT_NAME: self-hosted-${{ matrix.compose_version }}
+      COMPOSE_PROJECT_NAME: self-hosted-${{ strategy.job-index }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,14 +30,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - compose_version: '1.28.0'
-            compose_path: '/usr/local/bin'
-          - compose_version: '1.29.2'
-            compose_path: '/usr/local/bin'
-          - compose_version: 'v2.0.1'
-            compose_path: '/usr/local/lib/docker/cli-plugins'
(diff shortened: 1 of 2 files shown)


aminvakil (approved):
Thank you very much!
I'm not gonna bother you why this worked on `v2.0.1` :)

BYK:
>I'm not gonna bother you why this worked on v2.0.1 :)

Very likely because of the earlier `stdin` related issues they have on `v2.0.1`. Note that it says `Aborted` on the logs for that line so it wasn't really working, it was just quitting at that step since `stdin` was not available.

aminvakil (changes requested):
`COMPOSE_PROJECT_NAME` is not working.

If you want to, we can remove it from this PR and re-open #1367 for that.

BYK:
@aminvakil 
>COMPOSE_PROJECT_NAME is not working.

It _is_ working: `Container self-hosted-0-postgres-1`

aminvakil (approved):
Oops! I saw [this](https://github.com/getsentry/self-hosted/runs/5494674968?check_suite_focus=true#step:4:1754):
`  Successfully tagged sentry-cleanup-self-hosted-local:latest`
and I was mistaken that images created should have `-0-` too.

It _is_ working indeed: https://github.com/getsentry/self-hosted/runs/5494674968?check_suite_focus=true#step:4:1853

chadwhitacre:
Ok but _why_ does it work?

chadwhitacre (approved):
I don't understand why this works but it seems to work and doesn't look that bad.

BYK:
>I don't understand why this works but it seems to work and doesn't look that bad.

This works because I feed a `y` to the interactive prompt here: https://github.com/getsentry/self-hosted/pull/1370/files#diff-e386fde36977f06e32a6c6dbe63595743f0721ca78fcae81414fbdc0ca837b24R44

So instead of waiting for input indefinitely, it just moves along. The real fix is making the code I mentioned over https://github.com/getsentry/self-hosted/issues/1341#issuecomment-1063811848 not ask for a prompt when `--no-input` is passed or the `stdin` is not a `tty`.

PR #1384, opened 2022-03-15
Turn off containers under old name as well

chadwhitacre:
Aims to fix https://github.com/getsentry/self-hosted/issues/1178.


chadwhitacre:
I haven't test this, btw. Not sure if it actually works for the intended case, upgrade from pre-21.12.0 (pre-rename) to 21.12.0+.

BYK (approved):

PR #1397, opened 2022-03-22
fix(createuser): Use --no-input switch

rgibert:
--staff switch was added to createuser in https://github.com/getsentry/sentry/pull/32802 for usage in Single Tenant, this fixes the test failing after this addition.

Not sure the best approach to manage the chicken & egg problem of requiring this test to pass to merge my PR to sentry master, but merging this check will cause others to fail until I've merged PR.

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -41,7 +41,7 @@ echo "${_endgroup}"
 echo "${_group}Starting Sentry for tests ..."
 # Disable beacon for e2e tests
 echo 'SENTRY_BEACON=False' >> $SENTRY_CONFIG_PY
-echo y | $dcr web createuser --force-update --superuser --staff --email $TEST_USER --password $TEST_PASS
+$dcr web createuser --no-input --force-update --superuser --email $TEST_USER --password $TEST_PASS
 $dc up -d
 printf "Waiting for Sentry to be up"; timeout 90 bash -c 'until $(curl -Isf -o /dev/null $SENTRY_TEST_HOST); do printf '.'; sleep 0.5; done'
 echo ""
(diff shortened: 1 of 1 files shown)


BYK:
@rgibert there are two things you can (and probably should) do:

1. Never prompt when `--no-input` flag is passed.
2. Default to `staff` when we have `settings.SENTRY_SELF_HOSTED` set to `True`.

BYK:
Once you do item one, please come back here to remove that horrible `echo y` hack that I had to introduce from the line you modified 😅

rgibert:
I believe this test should work both before and after https://github.com/getsentry/sentry/pull/32802 is merged.

`staff` is defaulting whatever `superuser` is set to so will continue to behave as it did prior to my PR.

BYK:
@rgibert left a comment: https://github.com/getsentry/sentry/pull/32802/files#r833635984

_integration-test/run.sh
```diff
@@ -41,7 +41,7 @@ echo "${_endgroup}"
 echo "${_group}Starting Sentry for tests ..."
 # Disable beacon for e2e tests
 echo 'SENTRY_BEACON=False' >> $SENTRY_CONFIG_PY
-echo y | $dcr web createuser --force-update --superuser --email $TEST_USER --password $TEST_PASS
+$dcr web createuser --no-input --force-update --superuser --email $TEST_USER --password $TEST_PASS
```
BYK, line 44: I thought this was set through the `CI` variable 🤔

BYK (approved):

aminvakil (approved):
LGTM

chadwhitacre:
Thanks @github-actions.

chadwhitacre:
Shoot. Did we still need this?

_integration-test/run.sh
chadwhitacre: Thinking of this?

https://github.com/getsentry/self-hosted/blob/518a9ec1c0f0272650765a8265db52baa38a4382/install/set-up-and-migrate-database.sh#L3-L4

PR #1408, opened 2022-04-01
Use better API key when available

chadwhitacre:
Explanation:

https://github.com/getsentry/action-enforce-license-compliance/blob/236d528c324632485f9e8adb02b515ce01e81d12/action.yml#L16-L26


mwarkentin (approved):
LGTM.

Issue #1415, opened 2022-04-06
CI flakes with "no such service: web"


aminvakil:
It's been a couple of days I'm seeing this regression on docker-compose 1.28.0 and 1.29.2 on my fork:

https://github.com/aminvakil/self-hosted/actions/runs/2068250465
https://github.com/aminvakil/self-hosted/actions/runs/2074457629
https://github.com/aminvakil/self-hosted/actions/runs/2080468698
https://github.com/aminvakil/self-hosted/actions/runs/2083753352

This is a clean installation of course :)

Here is the error:
```
Bootstrapping and migrating Snuba ...
  Container self-hosted-0-clickhouse-1  Creating
  Container self-hosted-0-redis-1  Creating
  Container self-hosted-0-zookeeper-1  Creating
  Container self-hosted-0-clickhouse-1  Created
  Container self-hosted-0-zookeeper-1  Created
  Container self-hosted-0-kafka-1  Creating
  Container self-hosted-0-redis-1  Created
  Container self-hosted-0-kafka-1  Created
  Container self-hosted-0-redis-1  Starting
  Container self-hosted-0-zookeeper-1  Starting
  Container self-hosted-0-clickhouse-1  Starting
  Container self-hosted-0-redis-1  Started
  Container self-hosted-0-zookeeper-1  Started
  Container self-hosted-0-zookeeper-1  Waiting
  Container self-hosted-0-clickhouse-1  Started
  Container self-hosted-0-zookeeper-1  Healthy
  Container self-hosted-0-kafka-1  Starting
  Container self-hosted-0-kafka-1  Started
  Container self-hosted-0-kafka-1  Waiting
  Container self-hosted-0-redis-1  Waiting
  no such service: web
  An error occurred, caught SIGERR on line 3
  *NOT* cleaning up, to clean your environment run "docker compose stop".
  Error: Process completed with exit code 1.
```

This started to happen since last nightly run on https://github.com/getsentry/self-hosted CI too:

https://github.com/getsentry/self-hosted/runs/5843062307?check_suite_focus=true
https://github.com/getsentry/self-hosted/runs/5843062379?check_suite_focus=true

chadwhitacre:
This is plaguing the 22.5.0 release:

https://github.com/getsentry/publish/issues/1013

chadwhitacre:
[...]

PR #1423, opened 2022-04-15
Debug #1415

chadwhitacre:
https://github.com/getsentry/self-hosted/issues/1415

### DIFF
--- a/.env
+++ b/.env
@@ -5,10 +5,10 @@ SENTRY_EVENT_RETENTION_DAYS=90
 SENTRY_BIND=9000
 # Set SENTRY_MAIL_HOST to a valid FQDN (host/domain name) to be able to send emails!
 # SENTRY_MAIL_HOST=example.com
-SENTRY_IMAGE=getsentry/sentry:nightly
-SNUBA_IMAGE=getsentry/snuba:nightly
-RELAY_IMAGE=getsentry/relay:nightly
-SYMBOLICATOR_IMAGE=getsentry/symbolicator:nightly
+SENTRY_IMAGE=getsentry/sentry:22.3.0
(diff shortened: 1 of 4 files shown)


chadwhitacre:
Repro in 8.5 min:

https://github.com/getsentry/self-hosted/runs/6043241790?check_suite_focus=true#step:4:2573

chadwhitacre:
Hrm now with whittled down versions seeing "no service relay"

https://github.com/getsentry/self-hosted/runs/6043393674?check_suite_focus=true#step:4:1602

chadwhitacre:
Seems like `--minimize-downtime` is implicated?

chadwhitacre:
Confirmed only manifests under `--minimize-downtime`.

chadwhitacre:
This is where the failure occurs:

https://github.com/getsentry/self-hosted/blob/7cdb9d82578116b1df3e23ce5bb04c7305d47bd9/install/bootstrap-snuba.sh#L3

This is not itself wrapped in `MINIMIZE_DOWNTIME`.

chadwhitacre:
With e6ace90f6245f3d2a20a9b0629a2befc8eb097b9 ... `web` service seems to be there:

https://github.com/getsentry/self-hosted/runs/6043667634?check_suite_focus=true#step:4:3584

chadwhitacre:
Nothing has changed on master for 2 weeks, and this bug started showing up after that. Did something change in the an upstream container that is having a downstream effect here? Could test by pinning versions.

chadwhitacre:
Interesting! Fails when pinned to 22.3.0.

chadwhitacre:
Continued in https://github.com/getsentry/self-hosted/pull/1424.

PR #1424, opened 2022-04-15
Debug #1415

chadwhitacre:
#1423 redux

#1415

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,17 +14,6 @@ defaults:
   run:
     shell: bash
 jobs:
-  unit-test:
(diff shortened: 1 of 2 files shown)


chadwhitacre:
aef6128d3aac54122284ffc167835a682888273a is essentially the same as 22.3.0

chadwhitacre:
Okay! Same failure against 22.3.0. So this is due to nothing that we introduced either in `self-hosted` or upstream. Hmmmmmmm ... 🤔

In that case do we just turn off the two Docker Compose v1 checks for now in order to get 22.4.0 out the door, with a big note at the top? I think that's what we do ...

PR #1425, opened 2022-04-15
Disable Docker Compose v1 checks

chadwhitacre:
Workaround for https://github.com/getsentry/self-hosted/issues/1415.


chadwhitacre:
[Nevermind](https://github.com/getsentry/publish/issues/1013#issuecomment-1100479500).

PR #1430, opened 2022-04-26
fix: replace git.io links with redirect targets

asottile-sentry:
see: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Committed via https://github.com/asottile/all-repos


aminvakil (approved):
@asottile-sentry Thanks for fixing these!

PR #1446, opened 2022-05-11
Document From email display name

chadwhitacre:
To prevent recurrences of https://github.com/getsentry/sentry/issues/34387

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -21,7 +21,8 @@ mail.host: 'smtp'
 
 
 # The email address to send on behalf of
-# mail.from: 'root@localhost'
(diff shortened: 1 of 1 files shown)


BYK (approved):

chadwhitacre:
Is this accurate to what you did @marguspeet?

chadwhitacre:
I considered the self-hosted email docs but "set a valid address for mail.from setting in config.yml" is good enough there, points to the file and then with this PR you have what you need once you get there. The more general doc is probably worth updating ... https://github.com/getsentry/develop/pull/575.

glensc:
Agree with @BYK should also update web page.

PR #1469, opened 2022-05-13
ref: reset user to root for installation

asottile-sentry:
I'm looking to replace `gosu` with a `USER` instruction in https://github.com/getsentry/snuba/pull/2696 -- this requires us to reset the user back to root before proceeding with `apt` installation


chadwhitacre (approved):

chadwhitacre:
Close/reopen to jog CI

PR #1473, opened 2022-05-17
Omit explicit entrypoint in sentry-cleanup

spawnia:
I was wondering why that line was there and found that it is unnecessarily repeating the default set in `cron/Dockerfile`.
Other uses of the same image do not set the entrypoint explicitly too, so this makes the definition more concise and self-consistent.

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -322,7 +322,6 @@ services:
       context: ./cron
       args:
         BASE_IMAGE: "$SENTRY_IMAGE"
-    entrypoint: "/entrypoint.sh"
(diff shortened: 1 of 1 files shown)


aminvakil:
This has been added in https://github.com/getsentry/self-hosted/pull/861 to prevent https://github.com/getsentry/self-hosted/issues/860.

Can you please check again to see if it's unnecessary or not?

spawnia:
Oh I see, there are multiple levels of overwrites going on. cron/Dockerfile -> x-sentry-defaults -> sentry-cleanup`.

PR #1479, opened 2022-05-18
Ease modification of base image

### LINKED ISSUE #882: Add required libraries for sentry-ldap-auth to base Sentry image
### Summary

Since the Dockerfile is being removed and we can no longer build with these libraries locally, we're hoping they can be added to the base image. We're referencing a comment left by another user on the PR that took out the Sentry Dockerfile: https://github.com/getsentry/onpremise/pull/834#issuecomment-778915871

### Motivation

It allows the use of the ldap plugin for authentication.

### Additional Context

N/A

### LINKED ISSUE #1477: Install plugins in web and sentry-cleanup
### Version

22.4.0

### Steps to Reproduce

1. Use https://github.com/Banno/getsentry-ldap-auth
2. Install dependencies through `sentry/entrypoint.sh` (necessary because of https://github.com/getsentry/self-hosted/pull/834)
   
   ```sh
   apt-get update
   apt-get install -y gcc libsasl2-dev python-dev libldap2-dev libssl-dev
   ```

3. Add plugins through `sentry/requirements.txt`

   ```
   # Add plugins here
   pyldap
   sentry-ldap-auth
   ```

### Expected Result

`sentry-cleanup` should have the necessary dependencies and plugins to periodically clear old events.

### Actual Result

The cronjob within `sentry-cleanup` fails with an error related to importing the module LDAP. I am afraid I don't have the logs for this anymore to show the exact error, but it was pretty clear it was due to missing the plugins.

Prior to https://github.com/getsentry/self-hosted/pull/834, the dependencies and plugins were installed in a locally built base image, which was then used for both `web` and `sentry-cleanup`. Now, `sentry/entrypoint.sh` is only used for `web` - not `sentry-cleanup` - hence it is not working.

The following diff shows how I adapted my setup to revert back to building a local base image for both:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index b01e8e3..d7acd51 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,11 @@ x-healthcheck-defaults: &healthcheck_defaults
   start_period: 10s
 x-sentry-defaults: &sentry_defaults
   <<: *restart_policy
-  image: "$SENTRY_IMAGE"
+  build:
+    context: ./sentry
+    args:
+      - SENTRY_IMAGE
+  image: sentry-self-hosted-local
   depends_on:
     redis:
       <<: *depends_on-healthy
@@ -321,7 +325,7 @@ services:
     build:
       context: ./cron
       args:
-        BASE_IMAGE: "$SENTRY_IMAGE"
+        BASE_IMAGE: sentry-self-hosted-local
     entrypoint: "/entrypoint.sh"
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
   nginx:
diff --git a/install/build-docker-images.sh b/install/build-docker-images.sh
index 4bb96b5..616d667 100644
--- a/install/build-docker-images.sh
+++ b/install/build-docker-images.sh
@@ -1,6 +1,8 @@
 echo "${_group}Building and tagging Docker images ..."
 
 echo ""
+# Build the sentry self-hosted image first as it is needed for the cron image
+$dc build --force-rm web
 $dc build --force-rm
 echo ""
 echo "Docker images built."
diff --git a/sentry/Dockerfile b/sentry/Dockerfile
new file mode 100644
index 0000000..74b4f59
--- /dev/null
+++ b/sentry/Dockerfile
@@ -0,0 +1,7 @@
+ARG SENTRY_IMAGE
+FROM ${SENTRY_IMAGE}
+
+RUN apt-get update \
+    && apt-get install -y gcc libsasl2-dev python-dev libldap2-dev libssl-dev \
+    && rm -r /var/lib/apt/lists/* \
+    && pip install pyldap sentry-ldap-auth
```

I think this issue is related to https://github.com/getsentry/self-hosted/issues/882, but not necessarily asking the same. I just wanted to describe a specific problem I am having without shoeboxing our thinking and prescribing a specific solution.

spawnia:
Resolves #1477
Resolves #882

Provides a simple and canonical way to make modifications of the Sentry base image, such as installing plugins or their dependencies.

In contrast to previous solutions, this method:
- applies to all services using the Sentry base image (see https://github.com/getsentry/self-hosted/issues/1477)
- runs exactly once during `install.sh` for all services
- does not require modification of setup files within the repository

### Legal Boilerplate


.github/workflows/test.yml
```diff
@@ -59,14 +59,7 @@ jobs:
...
         run: |
-          echo "Testing initial install"
-          ./install.sh
-          ./_integration-test/run.sh
-          echo "Testing in-place upgrade"
-          # Also test plugin installation here
-          echo "sentry-auth-oidc" >> sentry/requirements.txt
-          ./install.sh --minimize-downtime
-          ./_integration-test/run.sh
+          ./test.sh
```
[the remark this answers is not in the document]
aminvakil: Can you please move this change to another PR?

sentry/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG SENTRY_IMAGE
+FROM ${SENTRY_IMAGE}
+
+COPY . /usr/src/sentry
```
aminvakil, line 4: Why should we copy all contents of self-hosted to sentry image?

test.sh
```diff
@@ -1,4 +1,19 @@
...
+apt-get update
+apt-get install -y gcc libsasl2-dev python-dev libldap2-dev libssl-dev
+EOT
+printf "python-ldap" > sentry/requirements.txt
+
+echo "Testing initial install"
+./install.sh
+./_integration-test/run.sh
+
+echo "Testing in-place upgrade"
```
aminvakil, line 17: What do you think about moving writing commands into `enhance-image.sh` in second run with `--minimize-downtime`?

_integration-test/run.sh
```diff
@@ -132,3 +132,8 @@ source ./custom-ca-roots/setup.sh
 $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
+
+echo "${_group}Ensure customizations work"
```
aminvakil, line 136: This should change as well if installing ldap moves before `minimize-downtime`.

aminvakil (changes requested):
I would say let's keep this PR to moving from using sentry image to building sentry image and do stuff after.

And open another PRs for other modifications.

.github/workflows/test.yml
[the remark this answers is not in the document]
spawnia: I also modified the steps, it was not just a pure refactoring. If you insist I can revert the move, but I think it is more convenient for testing to keep it.

sentry/Dockerfile
spawnia: It is just copying the contents of `sentry`, since that is the build context defined in `docker-compose.yml`. It is necessary to copy the entire folder because `sentry/requirements.txt` and `sentry/enhance-image.sh` are optional. It is not possible to copy single files conditionally in a `Dockerfile`.

aminvakil: > It is just copying the contents of `sentry`,

Oops, right.

> since that is the build context defined in `docker-compose.yml`. It is necessary to copy the entire folder because `sentry/requirements.txt` and `sentry/enhance-image.sh` are optionally present and influence the build.

Right.

> It is not possible to copy single files conditionally in a `Dockerfile`.

And right again!

test.sh
spawnia: Good idea, I will extract the validation for it and add a before-after comparison to ensure the test itself is sound.

_integration-test/run.sh
```diff
@@ -133,7 +133,7 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
 
-echo "${_group}Ensure customizations work"
-$dcr web bash -c "if [ ! -e /created-by-enhance-image ]; then exit 1; fi"
-$dcr web python -c "import ldap"
+echo "${_group}Ensure customizations are not present yet"
+! $dcr web bash -c "if [ ! -e /created-by-enhance-image ]; then exit 1; fi"
```
aminvakil, line 137: Please avoid negative negatives :)
It does not help my eyes, but let's hear from others too before changing.

spawnia: I generally agree, but decided that just adding the `!` in front would be better for maintenance. If the condition were to change we can just replace it in both files without thinking about how it can be negated.

```suggestion
$dcr web bash -c "if [ -e /created-by-enhance-image ]; then exit 1; fi"
```

aminvakil: What other conditional?

.github/workflows/test.yml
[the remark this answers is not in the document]
BYK: We use `test.sh` over here:

https://github.com/getsentry/sentry/blob/fe4795f5eae9e0d7c33e0ecb736c9d1369535eca/docker/cloudbuild.yaml#L59

So let's keep changes to that isolated as it can break on the other side of the world.

_integration-test/run.sh
spawnia: I clarified the symmetry between the positive and negative version of those test statements and added a hint to ensure it is maintained in the future.

.github/workflows/test.yml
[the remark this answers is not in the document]
spawnia: Especially now that the complexity of steps has grown, I think it is much preferrable to have a dedicated file for it.

README.md
```diff
@@ -57,7 +63,3 @@ sudo SENTRY_IMAGE=us.gcr.io/sentryio/sentry:83b1380 ./install.sh
 ```
 
 Where you replace `83b1380` with the sha you want to use.
-
-[build-status-image]: https://github.com/getsentry/self-hosted/workflows/test/badge.svg
```
BYK, line 61: Why remove these?

sentry/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+RUN if [ -s /usr/src/sentry/enhance-image.sh ]; then /usr/src/sentry/enhance-image.sh; fi
+RUN if [ -s /usr/src/sentry/requirements.txt ]; then pip install -r /usr/src/sentry/requirements.txt; fi
```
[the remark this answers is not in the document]
BYK: +1 for deprecation and then removing in the next release with a big warning in the changelog.

The deprecation warning should also be put in the entrypoint which uses and detects this file and should loudly complain during the install process.

test.sh
```diff
@@ -1,4 +1,21 @@
 #!/usr/bin/env bash
-set -e
+set -ex
 
-./_integration-test/run.sh
\ No newline at end of file
+echo "Testing initial install"
+./install.sh
```
BYK, line 5: Yeah, this is a no go for the getsentry/sentry repo. This file should be a new file named like `test-gha.sh` or something to avoid breakage from the other side.

BYK (changes requested):
Looks pretty great so far, thanks for putting this effort in. I'll leave the final approval to @chadwhitacre (he needs to approve to land anyway).

That said please address my comments regarding `test.sh` file as otherwise you'll break builds in all other Sentry repos :) -- @chadwhitacre we should probably add a simple GHA check for that here instead of using me as a human e2e/lint check 😅

README.md
spawnia: They are just references that don't do anything by themselves. I am guessing they were used at some point and then forgotten - this is why I never use them and just inline links.

.github/workflows/test.yml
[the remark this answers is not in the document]
spawnia: Got it, I extracted the more involved tests to a separate file and documented the dependency to the other repository.

test.sh
spawnia: Got it, solved in [`004d41b` (#1479)](https://github.com/getsentry/self-hosted/pull/1479/commits/004d41bc4d64ff3af7eb970474c04ccb1a703b86)

sentry/entrypoint.sh
```diff
@@ -5,16 +5,8 @@ if [ "$(ls -A /usr/local/share/ca-certificates/)" ]; then
...
-checksum_file="$plugins_dir/.checksum"
-
-if [[ -s "$req_file" ]] && ! cat "$req_file" | grep '^[^#[:space:]]' | shasum -s -c "$checksum_file" 2>/dev/null; then
-    echo "Installing additional dependencies..."
-    mkdir -p "$plugins_dir"
-    pip install --user -r "$req_file"
-    cat "$req_file" | grep '^[^#[:space:]]' | shasum > "$checksum_file"
-    echo ""
+if [ -e /etc/sentry/requirements.txt ]; then
+  echo "sentry/requirements.txt is deprecated, use sentry/enhance-image.sh"
```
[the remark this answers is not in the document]
spawnia: I added links everywhere and improved the README.

CONTRIBUTING.md
```diff
@@ -0,0 +1,7 @@
+## Testing
```
aminvakil, line 1: @chadwhitacre Do we have some guidelines already for `CONTRIBUTING.md`?

sentry/enhance-image.example.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+# Enhance the base $SENTRY_IMAGE with additional dependencies, plugins - see https://github.com/getsentry/self-hosted#enhance-sentry-image
```
aminvakil, line 3: What do you think about adding an example like this here?
```suggestion
# Enhance the base $SENTRY_IMAGE with additional dependencies, plugins - see https://github.com/getsentry/self-hosted#enhance-sentry-image
# For example:
# apt-get update
# apt-get install -y gcc libsasl2-dev python-dev libldap2-dev libssl-dev
# pip install python-ldap
```

chadwhitacre:
Thanks for the contribution @spawnia and for reviewing @BYK @aminvakil!

Kicking off CI ...

/gcbrun

CONTRIBUTING.md
chadwhitacre: Looks like we [don't have](https://github.com/getsentry/.github) a global CONTRIBUTING.md that this would override, wfm to start one here. 👍

install/build-docker-images.sh
```diff
@@ -1,6 +1,8 @@
 echo "${_group}Building and tagging Docker images ..."
 
 echo ""
+# Build sentry-self-hosted-local first as it is needed for sentry-cleanup-self-hosted-local
+$dc build --force-rm web
```
chadwhitacre, line 5: Comment says "Build sentry-self-hosted-local first" but then we "build ... web"?

sentry/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+COPY . /usr/src/sentry
+
+RUN if [ -s /usr/src/sentry/enhance-image.sh ]; then \
+    /usr/src/sentry/enhance-image.sh; \
+fi
+
+RUN if [ -s /usr/src/sentry/requirements.txt ]; then \
+    echo "sentry/requirements.txt is deprecated, use sentry/enhance-image.sh - see https://github.com/getsentry/self-hosted#enhance-sentry-image"; \
+    pip install -r /usr/src/sentry/requirements.txt; \
```
chadwhitacre, line 12: This covers the functionality removed from `entrypoint.sh`, yes? The checksum shenanigans are not necessary to bring forward?

integration-test.sh
```diff
@@ -0,0 +1,26 @@
...
+set -ex
+
+echo "Reset customizations"
+rm -f sentry/enhance-image.sh
+rm -f sentry/requirements.txt
+
+echo "Testing initial install"
+./install.sh
+./_integration-test/run.sh
+./_integration-test/ensure-customizations-not-present.sh
```
chadwhitacre, line 11: Why is this a separate script vs. called in/under `run.sh`? Same below with `ensure-customizations-work.sh`.

.github/workflows/test.yml
```diff
@@ -58,15 +58,7 @@ jobs:
...
-        run: |
-          echo "Testing initial install"
-          ./install.sh
-          ./_integration-test/run.sh
-          echo "Testing in-place upgrade"
-          # Also test plugin installation here
-          echo "sentry-auth-oidc" >> sentry/requirements.txt
-          ./install.sh --minimize-downtime
-          ./_integration-test/run.sh
+        run: ./integration-test.sh
```
chadwhitacre, line 61: The main purpose of factoring this out is to make it easier to run locally, yes? (cf. CONTRIBUTING.md). In that case, how does `./integration-test.sh` relate to `./test.sh`? Looks like there has been some back and forth on this with 004d41bc4d64ff3af7eb970474c04ccb1a703b86. :)

README.md
```diff
@@ -20,15 +30,13 @@ By default, there exists no `.env.custom` file. In this case, you can manually a
...
+To install plugins and their dependencies or make other modifications to the Sentry base image,
+copy `sentry/enhance-image.example.sh` to `sentry/enhance-image.sh` and add necessary steps there.
+For example, you can use `apt-get` to install dependencies and use `pip` to install plugins.
 
-During the install, a prompt will ask if you want to create a user account. If you require that the install not be blocked by the prompt, run `./install.sh --skip-user-prompt`.
-
-Thinking of not managing this yourself? Check out the [SaaS migration docs](https://docs.sentry.io/product/sentry-basics/migration/) or [contact us](https://sentry.io/from/self-hosted) for help.
-
-Please visit [our documentation](https://develop.sentry.dev/self-hosted/) for everything else.
+After making modifications to `sentry/enhance-image.sh`, run `./install.sh` again to apply them.
```
chadwhitacre, line 39: Glad to see docs for this, I'm torn between README and develop.sentry.dev/self-hosted. I think as with the .env customization we should dupe over there for now and we can dedupe in some glorious future where we fold all README content into dev docs.

_integration-test/ensure-customizations-work.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -ex
+
+source "$(dirname $0)/../install/_lib.sh"
+
+source ../install/dc-detect-version.sh
+
+# Negated version of ensure-customizations-not-present.sh, make changes in sync
```
chadwhitacre, line 8: Almost certainly a way to DRY this up.

chadwhitacre (changes requested):
Thanks for the PR! Some initial questions, not sure we're there yet on how test code is organized between the different scripts ... but I think we can get there. :)

sentry/Dockerfile
spawnia: As far as I can tell, they were only added with https://github.com/getsentry/self-hosted/pull/834 to maintain some level of caching. Since the image build is naturally run only once, it is no longer necessary to cache.

integration-test.sh
spawnia: Because they are logical opposites.

.github/workflows/test.yml
spawnia: > The main purpose of factoring this out is to make it easier to run locally, yes?

Exactly.

> In that case, how does ./integration-test.sh relate to ./test.sh?

I am not sure how to answer this. I think their purpose follows from their contents and usage in the project. Would you like to rename them or comment them?

_integration-test/ensure-customizations-work.sh
spawnia: I am sure there is, I am not convinced it is worth the effort though. I do follow DRY, but also WET - write everything twice. Given this is a negation, we will never need to maintain more than two versions of this.

sentry/Dockerfile
BYK: Yup, exactly as @spawnia said. I just poorly replicated what Docker layered file system automatically does.

chadwhitacre (approved):

chadwhitacre:
Thanks @spawnia! Merging as-is, I think there are some slight improvements possible but don't want to hold this up further. :)

spawnia:
Thanks to everyone involved. Looking forward to simplifying our setup with this.

README.md
chadwhitacre: Duping in https://github.com/getsentry/develop/pull/590.

PMExtra:
Gread job!

But why do we have to migrate from `requirements.txt` to `enhance-image.sh` ?

I think they can coexist. In some cases, use `requirements.txt` is simpler than shell script.

I think we can drop the `requirements.txt is deprecated` message, then everything will be awesome.

spawnia:
I think the simplicity of having a single shell script is preferrable.

`requirements.txt` pretty much only saves the addition of `pip install`, but at the added cost of:
- having an extra file
- having to document two ways of achieving the same thing
- missing clarity as to the order of execution - one more thing to think about or document

chadwhitacre:
Nice hack @PMExtra. :)

PR #1507, opened 2022-06-08
snuba: New subscriptions infrastucture rollout

lynnagara:
We have rebuilt the Snuba subscriptions infrastructure to help with scaling
subscriptions in SaaS. Hopefully it will be a bit more stable for self hosted
as well. It's configured to be able to recover more quickly from any downtimes or
backlogs as it ignores stale subscriptions and only executes and delivers alerts
on recent ones.

The ability to run the old subscriptions process will be removed in an upcoming
Snuba release.


lynnagara:
Hi @BYK 👋 Yep, it should be backwards compatible.

Issue #1381, opened 2022-03-15
Logs flooded with client request body is buffered to a temporary file messages


glensc:
### Version

21.21.0

### Steps to Reproduce

1. have some app write their data
2. `docker-compose logs --tail=30 -tf nginx`

### Expected Result

clean logs ;)

### Actual Result

```
[warn] 23#23: *24007 a client request body is buffered to a temporary file /var/cache/nginx/client_temp/0000010755, client: 192.168.24.21, server: , request: "POST /api/54/store/ HTTP/1.1"
```

glensc:
I think `/var/cache/nginx/client_temp/0000010755` should at least be stored on a volume. writing to image paths directly is ugly and likely slow (overlayfs layers come to play).

glensc:
Here's an example how to increase buffers:
- https://github.com/eventum/docker/commit/58c06696f42c6ad76f4f2fa7e517fbf1df1ac1c4

chadwhitacre:
> 21.21.0

Month 21, eh? :P

For reference here are the [docs for Nginx proxy_buffering](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering).

But we seem to have this off by default? 🤔

https://github.com/getsentry/self-hosted/blob/48dd5eb144858b47c7631dffcd325021c4daa7a2/nginx/nginx.conf#L39

aminvakil:
By default sentry's nginx is fine with `proxy_buffering off;` and it should better be off anyway as cached responses are unnecessary and cannot help IMO.

Do you see a reason to include this in default configuration?

glensc:
`proxy_buffering` is irrelevant here, you need `client_body_buffer_size` to remove those warnings:
- https://github.com/getsentry/self-hosted/issues/1381#issuecomment-1068121074

aminvakil:
For future reference: https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size

Do you think this should be documented somewhere? Add it to configured using environment variable? or something else?

glensc:
why not up the value? or at least add the default to confg for easy reference

aminvakil:
> why not up the value? or at least add the default to confg for easy reference

I agree with adding the default value commented to `nginx.conf`.

glensc:
16k still too small for `client_body_buffer_size`, upping to 32k:

```
[...]

PR #1511, opened 2022-06-12
Add volume for nginx cache

glensc:
Use volume for writable data, as it's supposedly faster because no need to update overlayfs layers on volumes.

NOTE: `nginx` itself does `chown nginx` on startup according to my tests:

```
# docker-compose exec nginx ls -la /var/cache/nginx
total 0
drwxr-xr-x    1 root     root            98 Jun 12 17:05 .
drwxr-xr-x    1 root     root            19 Nov 13  2021 ..
drwxr-xr-x    2 nginx    root             6 Jun 12 17:05 client_temp
drwx------    2 nginx    root             6 Jun 12 17:05 fastcgi_temp
drwx------    2 nginx    root             6 Jun 12 17:05 proxy_temp
drwx------    2 nginx    root             6 Jun 12 17:05 scgi_temp
drwx------    2 nginx    root             6 Jun 12 17:05 uwsgi_temp
```

Refs:
- https://github.com/getsentry/self-hosted/issues/1381#issuecomment-1068061553

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -338,6 +338,7 @@ services:
         read_only: true
         source: ./nginx
         target: /etc/nginx
+      - sentry-nginx-cache:/var/cache/nginx
     depends_on:
       - web
       - relay
@@ -380,6 +381,7 @@ volumes:
   # These store ephemeral data that needn't persist across restarts.
   sentry-secrets:
   sentry-smtp:
+  sentry-nginx-cache:
   sentry-zookeeper-log:
   sentry-kafka-log:
   sentry-smtp-log:


docker-compose.yml
```diff
@@ -338,6 +338,7 @@ services:
         read_only: true
         source: ./nginx
         target: /etc/nginx
+      - sentry-nginx:/var/cache/nginx/client_temp
```
aminvakil, line 341:
```suggestion
      - sentry-nginx-cache:/var/cache/nginx/client_temp
```

docker-compose.yml
```diff
@@ -380,6 +381,7 @@ volumes:
   # These store ephemeral data that needn't persist across restarts.
   sentry-secrets:
   sentry-smtp:
+  sentry-nginx:
```
aminvakil, line 384:
```suggestion
  sentry-nginx-cache:
```

aminvakil (commented):
Thanks for doing this!

I think `sentry-nginx-cache` is a more appropriate name for this volume and will reduce the effort later if we wanted to have another volume for a different part of our `nginx`.

Out of curiosity, is there a reason not to mount the entire `/var/cache/nginx` instead of only `client_temp` folder, even though it will not give us any value right now, it may do in future and also it's more suitable for our naming :)

glensc:
> Out of curiosity, is there a reason not to mount the entire `/var/cache/nginx` instead of only `client_temp` folder, even though it will not give us any value right now, it may do in future and also it's more suitable for our naming :)

My concern was that with `VOLUME` I was sure docker creates the empty directory, so with mounting the volume may end up missing dirs, but seems nginx creates (and sets permissions) the directories if they are missing:

```
root@docker2 sentry/21.12.0# docker-compose exec nginx ls -la /var/cache/nginx/
total 0
drwxr-xr-x    7 root     root            98 Jun 13 08:47 .
drwxr-xr-x    1 root     root            31 Nov 13  2021 ..
drwx------    2 nginx    root             6 Jun 13 08:47 client_temp
drwx------    2 nginx    root             6 Jun 13 08:47 fastcgi_temp
drwx------    2 nginx    root             6 Jun 13 08:47 proxy_temp
drwx------    2 nginx    root             6 Jun 13 08:47 scgi_temp
drwx------    2 nginx    root             6 Jun 13 08:47 uwsgi_temp
```

glensc:
Changed to add volume for whole `/var/cache/nginx`.

BYK (commented):
I think it would be better to use `tmpfs` for this then: https://docs.docker.com/compose/compose-file/compose-file-v3/#tmpfs

glensc:
not `tmpfs`, as if you want to keep in memory, set `client_body_buffer_size`:
- https://github.com/getsentry/self-hosted/issues/1381#issuecomment-1068121074

this is for cases when `client_body_buffer_size` is exceeded and need to persist on disk.

aminvakil:
@glensc `client_body_buffer_size` does not have to do anything with this, we can or cannot change the value of `client_body_buffer_size` by default in repository or manually.

But we can use `tmpfs` "for cases when `client_body_buffer_size` is exceeded and need to persist on disk" anyway, right?

BYK (approved):
I think @glensc has a good point there. Since the speed of `tmpfs` comes from it being based on the memory, using that here doesn't make much sense. Based on that I think this patch is good to go.

aminvakil (approved):
Aside from being `tmpfs` or not, this is definitely an improvement.

@glensc Thanks!

chadwhitacre (approved):

Issue #914, opened 2021-04-02
Support arm64


vova-lantsov-dev:
### Summary

Support for arm64 processors (including arm64v8).

### Motivation

Users can run Sentry on mini PCs like Raspberry Pi or Asus Tinker [edit: as well as on Apple M1s].

### Additional Context

There are 13 images used in official Sentry docker-compose configuration.
4 images support arm64 officially:
- memcached
- redis
- postgres
- nginx.

5 images do not support arm64 at all:
- smtp (tianon/exim4)
- geoipupdate (maxmindinc/geoipupdate:latest)
- snuba
- symbolicator
- relay

4 images support arm64 unofficially:
- cp-zookeeper (watershine/cp-zookeeper:6.1.0-arm64) - only 6.1.0 version
- cp-kafka (watershine/cp-kafka:6.1.0-arm64) - only 6.1.0 version
- clickhouse-server (altinity/clickhouse-server:20.10.1.4844-testing-arm) - looks like a test unstable build, that has 20.10.x version instead of 20.3.9.70 suggested by your config
- sentry (daprio/sentry:latest) - I guess this is 1.1.0 version of Sentry

chadwhitacre:
`geoipupdate` should (could) be optional, and probably `smtp` as well. Gonna need `snuba`, `symbolicator`, and `relay`, however. Fortunately those are Sentry-owned, so theoretically easier to achieve arm64 support there. Version `1.1.0` of Sentry is hecka ancient. Sure that's right?

vova-lantsov-dev:
Not familiar with Sentry versioning, maybe you're right :) Maybe I am wrong, because this docker hub repo is actively supported and gets updates regularly. Anyway, sentry service is supported by Sentry team as well, it would be good to have official arm64 builds for it too.

BYK:
As much as I find being able to use arm64 on production fascinating, I don't think we can commit to support this for the time being due to following reasons:

1. Our main cloud provider, Google Cloud Platform, does not support arm64 architecture so we don't have any incentives or real-world testing ability for this architecture
2. Our preferred CI platform, GitHub Actions, does not support arm64 either so even if we implemented something, testing these continously would be a challenge.

qiutian00:
> I've created patches for on-premise deploy and all related docker images, it may work while Sentry hasn't introduced official support yet: https://github.com/Sentry-ARM/onpremise

Great!

TuringKi:
I have

[...]

PR #1538, opened 2022-06-22
Add ARM support

chadwhitacre:
#914


BYK:
Looking forward to LEG support.

chadwhitacre:
I have to admit I googled "leg computer architecture" before remembering that you are entering into your prime with the dad jokes. 😏 😁 🤦‍♂️

BYK:
@chadwhitacre this is only the beginning but I'll save Sentry's public repos of the horrors.

aminvakil (commented):
So with this change there isn't any other stuff which does not work on ARM?

chadwhitacre:
> So with this change there isn't any other stuff which does not work on ARM?

Unclear. We have ARM working for main Sentry dev env so should be able to audit that ...

.env
```diff
@@ -9,6 +9,26 @@ SENTRY_IMAGE=getsentry/sentry:nightly
...
+
+CLICKHOUSE_IMAGE=yandex/clickhouse-server:20.3.9.70
+#CLICKHOUSE_IMAGE=altinity/clickhouse-server:20.10.1.4844-testing-arm
+# tl;dr ^^^ Uncomment this on ARM.
+
+# Images built on arm need to be tagged to use linux/arm64 on Apple silicon Macs.
+# To work around this issue we configure the platform here
+
+# See https://github.com/docker/cli/issues/3286 for the docker bug.
+DOCKER_PLATFORM=linux/amd64
```
ethanhs, line 28: A potentially nicer solution than putting this and the clickhouse image in the `.env` would be to detect the docker architecture via `docker info --format '{{.Architecture}}'` and set them accordingly. I can add this if you think it is a good idea @chadwhitacre

BYK: The less the manual work there is, the better this is so I'm a big 👍🏻

ethanhs: @chadwhitacre https://github.com/getsentry/self-hosted/pull/1538/commits/8ef61db0a439c13c9bafdb30539174acb973da87 includes the changes I was working on for the auto-detection, I am not sure why the `turning-things-off.sh` script can't see the environment variables I set...

install/detect-platform.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# Images built on ARM also need to be tagged to use linux/arm64 on Apple
+# silicon Macs to work around an issue where they are built for
+# linux/amd64 by default due to virtualization.
+# See https://github.com/docker/cli/issues/3286 for the Docker bug.
+
+if [[ "$(docker info --format '{{.Architecture}}')" = "x86_64" ]]; then
+    export DOCKER_PLATFORM="linux/amd64"
+    export CLICKHOUSE_IMAGE="yandex/clickhouse-server:20.3.9.70"
+else
```
aminvakil, line 19: What do you think about improving these statements?
I mean currently `arm64` is being considered for every architecture and error which does not return `x86_64`.

I would say check for explicit `x86_64` and explicit `xxx` (I don't know what docker would output in arm) and if none of them was returned just drop and echo your system is not supported.

For example on a system which docker service is not started it returns:
```
$ docker info --format '{{.Architecture}}'

template: :1:2: executing "" at <.Architecture>: reflect: indirection through nil pointer to embedded struct field Info
```

aminvakil (commented):
Hey @ethanhs !

install/detect-platform.sh
ethanhs: Yeah, this definitely could be better. I will make the changes you suggest, thanks!

aminvakil (approved):
I'm not sure if it needs more change for ARM or not, but I've tested this in a clean local machine and it works fine.

ethanhs:
> I'm not sure if it needs more change for ARM or not, but I've tested this in a clean local machine and it works fine.

Well, it works on my machine :) (which is an ARM/M1 Pro Mac).

I haven't tested this on ARM Linux, but it should theoretically work.

install/check-minimum-requirements.sh
```diff
@@ -6,6 +6,12 @@ source "$(dirname $0)/_min-requirements.sh"
 function ver () { echo "$@" | awk -F. '{ printf("%d%03d%03d", $1,$2,$3); }'; }
 
 DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
+
+if [[ -z "$DOCKER_VERSION" ]]; then
+  echo "FAIL: Unable to get docker version, is the docker daemon running?"
+  exit 1
+fi
+
```
chadwhitacre, line 14: Looks like a sidecar, yes? Or is this required for this PR?

install/check-minimum-requirements.sh
```diff
@@ -36,15 +42,4 @@ elif [[ "$RAM_AVAILABLE_IN_DOCKER" -lt "$MIN_RAM_SOFT" ]]; then
...
-# On KVM, cpuinfo could falsely not report SSE 4.2 support, so skip the check. https://github.com/ClickHouse/ClickHouse/issues/20#issuecomment-226849297
-IS_KVM=$(docker run --rm busybox grep -c 'Common KVM processor' /proc/cpuinfo || :)
-if [[ "$IS_KVM" -eq 0 ]]; then
-  SUPPORTS_SSE42=$(docker run --rm busybox grep -c sse4_2 /proc/cpuinfo || :)
-  if [[ "$SUPPORTS_SSE42" -eq 0 ]]; then
-    echo "FAIL: The CPU your machine is running on does not support the SSE 4.2 instruction set, which is required for one of the services Sentry uses (Clickhouse). See https://github.com/getsentry/self-hosted/issues/340 for more info."
-    exit 1
-  fi
-fi
-
```
chadwhitacre, line 49: Why are we getting rid of this here? I have a general sense it's related to ARM but I don't fully understand how. Why can get rid of this now?

.env
```diff
@@ -9,6 +9,7 @@ SENTRY_IMAGE=getsentry/sentry:nightly
 SNUBA_IMAGE=getsentry/snuba:nightly
 RELAY_IMAGE=getsentry/relay:nightly
 SYMBOLICATOR_IMAGE=getsentry/symbolicator:nightly
+
```
chadwhitacre, line 12: ?

install/detect-platform.sh
```diff
@@ -0,0 +1,30 @@
...
+# See https://github.com/docker/cli/issues/3286 for the Docker bug.
+
+DOCKER_ARCH=$(docker info --format '{{.Architecture}}')
+
+if [[ "$DOCKER_ARCH" = "x86_64" ]]; then
+    export DOCKER_PLATFORM="linux/amd64"
+    export CLICKHOUSE_IMAGE="yandex/clickhouse-server:20.3.9.70"
+elif [[ "$DOCKER_ARCH" = "aarch64" ]]; then
+    export DOCKER_PLATFORM="linux/arm64"
+    export CLICKHOUSE_IMAGE="altinity/clickhouse-server:20.10.1.4844-testing-arm"
```
chadwhitacre, line 23: Do you think it's worth trying to keep this in sync with [our dev envs](https://github.com/getsentry/sentry/blob/2619db88ab9eacfdda77bb362813ce20536a5667/src/sentry/conf/server.py#L1931-L1935)?

At least with comments in both places. Also this isn't _quite_ in sync off the bat. 🙈 I picked this version because it's the closes to 20.3.9.70 I could find. Maybe decide with Anthony?

chadwhitacre:
@ethanhs lolsob I'm the PR author so I can't approve. Left some comments inline.

.env
ethanhs: Oops, this is a remnant from an earlier iteration of my changes where I had the image/platform info in .env, I'll remove this.

install/check-minimum-requirements.sh
ethanhs: This is in response to the review by Amin here: https://github.com/getsentry/self-hosted/pull/1538#discussion_r912368681

I wanted to check the daemon was running as early as possible so later commands don't fail.

install/check-minimum-requirements.sh
ethanhs: This was in your original PR 😁 

The reason is ARM uses a different kind of SIMD, so it won't have SSE 4.2. I could change this check to only run on x86_64...

install/detect-platform.sh
```diff
@@ -0,0 +1,30 @@
...
+# support ARM, which is relevant especially for Apple M1 laptops, Sentry's
+# standard developer environment. As a workaround, we use an altinity image
+# targeting ARM.
+#
+# See https://github.com/getsentry/self-hosted/issues/1385#issuecomment-1101824274
+#
+# Images built on ARM also need to be tagged to use linux/arm64 on Apple
+# silicon Macs to work around an issue where they are built for
+# linux/amd64 by default due to virtualization.
+# See https://github.com/docker/cli/issues/3286 for the Docker bug.
```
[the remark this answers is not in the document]
ethanhs: Thanks! Spent a good few hours pulling my hair out over this issue, rather annoying...

install/detect-platform.sh
ethanhs: Yeah I think that is probably a good idea, it at least means any issues can be tested with the dev environment. I'll discuss it with Anthony.

ethanhs:
> lolsob I'm the PR author so I can't approve

@chadwhitacre Ha, yeah I was going to request review from you but you can't do that on your own PR :P

ethanhs (approved):
I think this is good to merge :)

Issue #1550, opened 2022-06-29
FAIL: Expected minimum docker-compose version to be 1.24.1 but found Docker Compose version 2.5.0


jamincollins:
### Version

any version prior to 21.5.0

### Steps to Reproduce

Any `docker-compose` 2.x release installed (essentially any current `docker-compose`).

```bash
cd ${self-host-clone}
docker-compose down
git fetch --all
git checkout 21.5.0
./install.sh
```

### Expected Result

Normal upgrade would happen.

### Actual Result

upgrade fails with:

```
FAIL: Expected minimum docker-compose version to be 1.24.1 but found Docker Compose version 2.5.0
```

In order to proceed, the check must be corrected or commented out:
```bash
$ git diff
diff --git a/install/check-minimum-requirements.sh b/install/check-minimum-requirements.sh
index 4527a22..ba764c9 100644
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -20,10 +20,10 @@ if [[ "$(ver $DOCKER_VERSION)" -lt "$(ver $MIN_DOCKER_VERSION)" ]]; then
   exit 1
 fi
 
-if [[ "$(ver $COMPOSE_VERSION)" -lt "$(ver $MIN_COMPOSE_VERSION)" ]]; then
-  echo "FAIL: Expected minimum docker-compose version to be $MIN_COMPOSE_VERSION but found $COMPOSE_VERSION"
-  exit 1
-fi
+#if [[ "$(ver $COMPOSE_VERSION)" -lt "$(ver $MIN_COMPOSE_VERSION)" ]]; then
+#  echo "FAIL: Expected minimum docker-compose version to be $MIN_COMPOSE_VERSION but found $COMPOSE_VERSION"
+#  exit 1
+#fi
 
 if [[ "$CPU_AVAILABLE_IN_DOCKER" -lt "$MIN_CPU_HARD" ]]; then
   echo "FAIL: Required minimum CPU cores available to Docker is $MIN_CPU_HARD, found $CPU_AVAILABLE_IN_DOCKER"
```

ethanhs:
Hi!

Could you please provide what the output of `$(ver $COMPOSE_VERSION)` and `$(ver MIN_COMPOSE_VERSION)` (you can just add e.g. `echo $(ver $COMPOSE_VERSION)` above/in place of the if check).

[...]

PR #1551, opened 2022-06-29
ref: Replace regex with --short flag to get compose version

ethanhs:
While looking at how we get the compose version for https://github.com/getsentry/self-hosted/issues/1550 I realized there was a cleaner way to do this. This flag works all the way back to compose 1.4, when the `version` argument was introduced.


ethanhs:
> I remember trying to use --short and then not being able to use it for some reason.

Hmm OK. Well hopefully there aren't any more unusual version differences than just the `v` prefix.

I did check that Debian does *not* change the format of the output to reflect it has been patched, which I was worried about.

chadwhitacre:
Reminder that we're no longer running CI against Docker Compose v1 so that could be the delta.

chadwhitacre (approved):

PR #1553, opened 2022-06-30
ref: use sort -V to check minimum versions

ethanhs:
The [sort manpage](https://man7.org/linux/man-pages/man1/sort.1.html) might be handy when reviewing this if you are not familiar with the flags.

tl;dr: 
 - --version-sort sorts by version number
 - --reverse because we want the larger version first, as we pass $COMPOSE_VERSION first
 - --check=quiet will have sort return 1 if the order changes in sorting (aka, $COMPOSE_VERSION is less than $MIN_COMPOSE_VERSION)


BYK (approved):
LGTM. I guess I either couldn't discover the `-V` option or it was not widely available. I remember the Docker Engine version being the actual problem with their weird semver + calver combo but I tested some examples and all seem to pass.

aminvakil (approved):
@ethanhs This is great, I will shamelessly use your function in other places for myself too :smiley: .

I've also tested the function correctly and it works great.

ethanhs:
> This is great, I will shamelessly use your function in other places for myself too

Thanks :)

> I guess I either couldn't discover the -V option or it was not widely available.

I think it isn't very well known. It's been part of coreutils since 2014, idk about on macOS however, it may be a more recent addition there.

chadwhitacre (approved):

Issue #1556, opened 2022-07-03
/bin/bash: line 1: gosu: command not found


goganchic:
### Version

22.6.0

### Steps to Reproduce

1. Start Sentry with docker-compose
2. Wait for a bit
3. Check `docker-compose logs`

### Expected Result

No errors

### Actual Result

```
sentry-self-hosted-snuba-transactions-cleanup-1                | /bin/bash: line 1: gosu: command not found
sentry-self-hosted-snuba-cleanup-1                             | /bin/bash: line 1: gosu: command not found
```

aminvakil:
```
$ docker exec -ti sentry-self-hosted-snuba-cleanup-1 whereis gosu
gosu: /usr/local/bin/gosu
$ docker exec -ti sentry-self-hosted-snuba-cleanup-1 gosu --help
Usage: gosu user-spec command [args]
   eg: gosu tianon bash
       gosu nobody:root bash -c 'whoami && id'
       gosu 1000:1 id

gosu version: 1.12 (go1.13.10 on linux/amd64; gc)
gosu license: GPL-3 (full text at https://github.com/tianon/gosu)
```
It's working fine for me, are you sure your docker has been built correctly?

goganchic:
You run this command using shell directly, but docker-compose run this command using cron, so the command will be something like `cron "0 0 * * * /usr/local/bin/gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"`. I don't think that in this case `$PATH` variable will be the same as for shell.

goganchic:
The problem is not with paths: sentry-cleanup and symbolicator-cleanup have gosu, but snuba-transactions-cleanup and snuba-cleanup. I think, it's because snuba-transactions-cleanup and snuba-cleanup use another base image. Does it really contains gosu?

I use such versions:

```
SENTRY_IMAGE=getsentry/sentry:22.6.0
SNUBA_IMAGE=getsentry/snuba:22.6.0
RELAY_IMAGE=getsentry/relay:22.6.0
SYMBOLICATOR_IMAGE=getsentry/symbolicator:0.5.1
```

goganchic:
I'd checked the problem with nightly images - nothing changed: sentry-cleanup and symbolicator-cleanup have gosu, but snuba-transactions-cleanup and snuba-cleanup do not have.

If I add `gosu` into `cron/Dockerfile` - everything works as expected.

dlouzan:
[...]

PR #1560, opened 2022-07-03
Remove unnecessary env variables passing

goganchic:
PR into docs: https://github.com/getsentry/develop/pull/628

### Legal Boilerplate

### DIFF
--- a/install/build-docker-images.sh
+++ b/install/build-docker-images.sh
@@ -3,8 +3,8 @@ echo "${_group}Building and tagging Docker images ..."
 echo ""
 # Build any service that provides the image sentry-self-hosted-local first,
 # as it is used as the base image for sentry-cleanup-self-hosted-local.
-$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm web
-$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm
+$dc build --force-rm web
+$dc build --force-rm
 echo ""
 echo "Docker images built."
 
--- a/install/install-wal2json.sh
+++ b/install/install-wal2json.sh
@@ -5,9 +5,7 @@ ARCH=$(uname -m)
 FILE_NAME="wal2json-Linux-$ARCH-glibc.so"
 
 docker_curl() {
-    # The environment variables can be specified in lower case or upper case.
-    # The lower case version has precedence. http_proxy is an exception as it is only available in lower case.
-    docker run --rm -e http_proxy -e https_proxy -e HTTPS_PROXY -e no_proxy -e NO_PROXY curlimages/curl:7.77.0 "$@"
+    docker run --rm curlimages/curl:7.77.0 "$@"
 }
 
 if [[ $WAL2JSON_VERSION == "latest" ]]; then


docker-compose.yml
```diff
@@ -264,15 +264,15 @@ services:
       context: ./cron
       args:
         BASE_IMAGE: "$SNUBA_IMAGE"
-    command: '"*/5 * * * * gosu snuba snuba cleanup --storage errors --dry-run False"'
+    command: '"*/5 * * * * /usr/sbin/gosu snuba snuba cleanup --storage errors --dry-run False"'
```
aminvakil, line 267: This is not related to this PR and we're not still sure if there is a problem at all, as no one else has reported this issue until now. Let's keep the discussion in https://github.com/getsentry/self-hosted/issues/1556.

install/build-docker-images.sh
```diff
@@ -3,8 +3,8 @@ echo "${_group}Building and tagging Docker images ..."
 echo ""
 # Build any service that provides the image sentry-self-hosted-local first,
 # as it is used as the base image for sentry-cleanup-self-hosted-local.
-$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm web
-$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm
```
aminvakil, line 7: Doesn't this break current users' workflow using `http_proxy` env variable?

install/install-wal2json.sh
```diff
@@ -5,9 +5,7 @@ ARCH=$(uname -m)
 FILE_NAME="wal2json-Linux-$ARCH-glibc.so"
 
 docker_curl() {
-    # The environment variables can be specified in lower case or upper case.
-    # The lower case version has precedence. http_proxy is an exception as it is only available in lower case.
-    docker run --rm -e http_proxy -e https_proxy -e HTTPS_PROXY -e no_proxy -e NO_PROXY curlimages/curl:7.77.0 "$@"
+    docker run --rm curlimages/curl:7.77.0 "$@"
```
aminvakil, line 8: Also here.

aminvakil (changes requested):

docker-compose.yml
goganchic: My mistake, yep. Removed this commit from PR.

install/build-docker-images.sh
goganchic: Yes you are right. But I think it's very specific scenario: such users can access the Dockerhub without any proxy, but need http proxy for some resources during Sentry installation, like https://github.com/getsentry/wal2json.

For users who access any Internet resources (including Dockerhub) during installation it's required to set HTTP Proxy several times:
1. Docker systemd file for `docker pull`
2. `~/.docker/config.json` for any commands inside docker containers
3. `/etc/environment` for settings shell proxy variables

I think we can set only do 1. and 2. But in this case `docker build` will fail because in the commands above we explicitly overwrite proxy variables with empty values, that is why I suggest to modify these lines.

Actually we can keep these lines unchanged but from our discussion https://github.com/getsentry/develop/pull/622#discussion_r912366290 I'd found that this behavior is a bit unclear, that is why I'd created this PR with such changes.

goganchic: Let me check it one more time. I'd tried to install Sentry with my changes again and there are some issues.

goganchic: I'd rechecked this one more time. Proxies from `~/.docker/config.json` work only for `docker run` command and do not work for `docker build`. It is written in the [docker docs](https://docs.docker.com/network/proxy/):

> When you create or start new containers, the environment variables are set automatically within the container.

So we have to pass proxy env variables explicitly for the build command, that is why this PR shouldn't be merged.

aminvakil:
@goganchic Thank you very much for your effort and getting to the bottom of this!

PR #1564, opened 2022-07-05
do not use gosu for snuba-transactions-cleanup and snuba-cleanup

goganchic:
Fixes https://github.com/getsentry/self-hosted/issues/1556

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -264,7 +264,7 @@ services:
       context: ./cron
       args:
         BASE_IMAGE: "$SNUBA_IMAGE"
-    command: '"*/5 * * * * gosu snuba snuba cleanup --storage errors --dry-run False"'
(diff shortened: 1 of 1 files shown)


aminvakil:
`gosu` has been removed from Dockerfile because sunba Dockerfile is using `USER snuba` at the end of Dockerfile, so there isn't a need to install `gosu` again.

I would say remove `gosu` from commands in `docker-compose.yml` instead of adding `gosu` again.

aminvakil:
I could reproduce the issue (#1556) and also could verified removing `gosu` fixed the problem.

@goganchic Do you want to change your PR to the solution I've said?

goganchic:
Ok, I'll try.

goganchic:
I've checked you suggestion locally: everything works fine. Changed PR as you said.

aminvakil (approved):
Thanks for fixing this!

BYK (approved):

Issue #1532, opened 2022-06-20
Installation failed: Configuration file does not exist


gander:
### Version

22.6.0

### Steps to Reproduce

1. Download https://github.com/getsentry/self-hosted/archive/refs/tags/22.6.0.zip
2. Unpack
3. Run `./install.sh`. Error `Failed to create relay credentials in ../relay/credentials.json.` occurs
4. Run `./install.sh` again
5. Processing (fetching images, starting containers)

### Expected Result

To make it work

### Actual Result

```
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
Error: Configuration file does not exist. Use 'sentry init' to initialize the file.
An error occurred, caught SIGERR on line 12
Cleaning up...
```

gander:
It's crashing when running:

```bash
$dcr web upgrade
# docker compose --ansi never --env-file ${PWD}/.env run --rm web upgrade
```

```bash
docker compose --ansi never --env-file ${PWD}/.env run --rm web bash
# and then
sentry upgrade
```
It looks like the process is going on. But only this fragment.

gander:
Some errors:
```
/usr/local/lib/python3.8/site-packages/memcache.py:1303: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if key is '':
/usr/local/lib/python3.8/site-packages/memcache.py:1304: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if key_extra_len is 0:
```

gander:
I used this workaround:
```bash
$dcr web bash -c 'sentry upgrade --noinput'
```
The process seems to be completed, but the web does not get up:
```
container for service "web" is unhealthy
```

gander:
Infinite restart loop:
```
Error: Configuration file does not exist. Use 'sentry init' to initialize the file.
```

aminvakil:
There is no need to perform any command manually, please remove and clean anything and execute `./install.sh` again, this time with debug mode. Something like this:

```
docker-compose down -v --remove-orphans
git clean -fdx
docker volume prune -a **### This will erase all volumes which no container is using now, notice if you have other volumes on this machine. ###**
```
You can also use this command to make sure you're only removing volumes related to sentry:
```
for volume in $(docker volume list --format '{{ .Name }}' | grep '^sentry-'); do
  docker volume remove $volume > /dev/null \
    && echo "Removed volume: $volume" \
[...]

PR #1589, opened 2022-07-20
Add check requiring git checkout

ethanhs:
This adds a requirement that the install script is run from a git checkout, not from a zip. This is to avoid issues like https://github.com/getsentry/self-hosted/issues/1532

### DIFF
--- a/install/check-latest-commit.sh
+++ b/install/check-latest-commit.sh
@@ -9,7 +9,9 @@ if [[ -d "../.git" && "${SKIP_COMMIT_CHECK:-0}" != 1 ]]; then
     fi
   fi
 else
-  echo "skipped"
+  # Require a git checkout to avoid permissions issues ref https://github.com/getsentry/self-hosted/issues/1532
+  echo "Detected sources not from git checkout. Please clone the repository instead of downloading a ZIP."
+  exit 1
 fi
 
 echo "${_endgroup}"


aminvakil (changes requested):
`check-latest-commit` only returns error and does something if it's `.git` directory exists.

If self-hosted has been downloaded from an archive and `.git` directory is absent, it would be just skipped.

So I don't know why should we enforce users to use git clone instead of downloading an archive.

Do we face another problem somewhere else, because I can't think of a scenario that `check-latest-commit` should `exit 1` if `.git` directory is absent.

I would try a manual clean install with zip folder and report back.

Wild guess, but maybe `echo "skipped"` manipulates input of another command and we should change that.

BYK:
I think forcing a git checkout is a heavy-handed solution. If I remember correctly, we ourselves, refer people to using the zip archive. Even more, we use the `.tar.gz` archive for testing self-hosted on other repos: https://github.com/getsentry/sentry/blob/88892a6c8a29f12f6c678901cca8f8455ca7ef78/docker/cloudbuild.yaml#L44

So we should make sure the archive path works.

BYK:
>Even more, we use the .tar.gz archive for testing self-hosted on other repos: [getsentry/sentry@88892a6/docker/cloudbuild.yaml#L44](https://github.com/getsentry/sentry/blob/88892a6c8a29f12f6c678901cca8f8455ca7ef78/docker/cloudbuild.yaml#L44)

Pretty sure this will cause all repo-integration tests to fail once merged to master

aminvakil:
Downloading https://github.com/getsentry/self-hosted/archive/refs/tags/22.7.0.zip and using on a clean server works.

aminvakil:
https://github.com/getsentry/self-hosted/archive/refs/heads/master.zip worked fine too.

ethanhs:
Ok, sounds like I should close this then.

Issue #740, opened 2020-11-17
Add Sentry error reporting to install.sh


BYK:
## Summary

`sentry-cli` has a Docker image and Bash signal trap support: https://docs.sentry.io/product/cli/send-event/#bash-hook

The two didn't play well together but now this can be overcome: https://github.com/getsentry/sentry-cli/pull/852

Hence, we should try this out and see if it is useful to catch unknown issues in the wild.

## Motivation

We rely on people to report issues with the install script and provide relevant logs along with that. This has worked okay so far but the process is very manual, cumbersome, and relies on the users reporting issues. Sentry was created to solve these issues to why not dog food?

chadwhitacre:
@armenzg has a blog post coming out about how the Developer Productivity team did something very similar with our environment bootstrapping script. Inspo!

chadwhitacre:
I'm going to reopen this because there are many more details we need to pay attention to in order to not piss off users with this change.

chadwhitacre:
Sorry ... we've got https://github.com/getsentry/team-ospo/issues/36 as our meta ticket.

PR #1671, opened 2022-08-29
Fix fresh install of self-hosted for ARM by removing the --platform parameter

hubertdeng123:
Fresh install of self-hosted was throwing an exception here on my M1 macbook pro:

```console
▶ Setting up / migrating database …
...
Container sentry-self-hosted-clickhouse-1  Healthy
Container sentry-self-hosted-kafka-1  Healthy
Container sentry-self-hosted-kafka-1  Healthy
Container sentry-self-hosted-kafka-1  Healthy
Container sentry-self-hosted-redis-1  Healthy
Container sentry-self-hosted-clickhouse-1  Healthy
Container sentry-self-hosted-clickhouse-1  Healthy
Container sentry-self-hosted-clickhouse-1  Healthy
Container sentry-self-hosted-snuba-sessions-consumer-1  Starting
Container sentry-self-hosted-snuba-subscription-consumer-events-1  Starting
Container sentry-self-hosted-snuba-outcomes-consumer-1  Starting
Container sentry-self-hosted-snuba-transactions-consumer-1  Starting
Container sentry-self-hosted-snuba-api-1  Starting
Container sentry-self-hosted-snuba-consumer-1  Starting
Container sentry-self-hosted-snuba-subscription-consumer-transactions-1  Starting
Container sentry-self-hosted-snuba-replacer-1  Starting
Container sentry-self-hosted-snuba-outcomes-consumer-1  Started
Container sentry-self-hosted-snuba-api-1  Started
Container sentry-self-hosted-snuba-subscription-consumer-events-1  Started
Container sentry-self-hosted-snuba-transactions-consumer-1  Started
Container sentry-self-hosted-snuba-subscription-consumer-transactions-1  Started
Container sentry-self-hosted-snuba-sessions-consumer-1  Started
Container sentry-self-hosted-snuba-replacer-1  Started
Container sentry-self-hosted-snuba-consumer-1  Started
Error response from daemon: image with reference sentry-self-hosted-local was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64
```

This is happening in the installation process because certain images are only available for `linux/amd64` and the `platform` parameter here forces us to look for a `linux/arm64` image, eventually erroring out. For M1 macs, I don't think we can confidently assume the platform to be `linux/arm64` as long as we are running `linux/amd64` single arch images. For multi-arch images, docker should automatically pull in the image that matches the current OS and architecture.

Github thread here for more context on a similar scenario:
[...]

PR #1679, opened 2022-09-01
Error monitoring of the self-hosted installer

ethanhs:
In this PR we:

- Add the ability to send events to sentry (currently just an event on error/exit of non-zero status)
- Print a nice stack trace when there is an error (I have some tricks to make it even nicer, such as moving things into functions)
- Prompt the user on (and save!) their preference of reporting errors to us or not (saved e.g. for future upgrading etc)

This seems to work if I sprinkle a "false" into one of the files in `install/` and the grouping seems to work correctly too.

Fixes https://github.com/getsentry/self-hosted/issues/740


install/_lib.sh
```diff
@@ -8,6 +8,47 @@ umask 002
...
+    export CLICKHOUSE_IMAGE="altinity/clickhouse-server:21.6.1.6734-testing-arm"
+else
+    echo "FAIL: Unsupported docker architecture $DOCKER_ARCH."
+    exit 1
+fi
+echo "Detected Docker platform is $DOCKER_PLATFORM"
+
+function send_event {
+  if [[ $DOCKER_PLATFORM == "linux/amd64" ]]; then
+    local sentry_cli="docker run --rm -v \$(pwd):/work -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
```
BYK, line 39: This volume mount is very likely to cause issues on WSL and potentially in macOS as they use VMs and network drives. Just beware :)

install/error-handling.sh
```diff
@@ -1,10 +1,34 @@
 echo "${_group}Setting up error handling ..."
 
+export SENTRY_DSN='https://afa81b4b7c634e318b0ec138abb2645f@self-hosted.getsentry.net/2'
+export SENTRY_ORG=sentry
+export SENTRY_PROJECT=dogfooding
```
BYK, line 5: Really? :D Why not `self-hosted`?

BYK (approved):
Nice!

aminvakil (approved):
This will be great! Thanks!

Some questions though:
- Can non-sentry collaborator (myself :) ) view these logs and if yes, how?
- Do we have a sample log of what will be sent?

I'd also suggest adding some uuid (some hash of IP?) which gets printed and gets sent. Then we can ask users their uuid to view their logs on our instance too. I'm not sure about privacy/policy issues this may brought, but as we're already getting users' IP, I don't think a hash of that would be a problem. This can be pasted in github in public too without a problem.

install/_lib.sh
ethanhs: Well, I was testing on a Mac, so at least we should be good there! :)

Could you explain why would it cause issues with the WSL?

install/error-handling.sh
ethanhs: It's already at `self-hosted.sentry.net`! But I can always change this if you think I should.

ethanhs:
> Can non-sentry collaborator (myself :) ) view these logs and if yes, how?

The sentry instance is not SaaS, so I'll talk to @chadwhitacre about if this would be possible.

> Do we have a sample log of what will be sent?

We don't, but I could generate one. It is the same as the default log we store locally after installation.

> I'd also suggest adding some uuid (some hash of IP?) which gets printed and gets sent.

This is a good idea. I am worried that someone could just enumerate all hashes of IPs (there aren't that many IPv4s) and so it wouldn't be secret. I'll have to think about a better source for the hash but I think this is a good idea overall.

install/_lib.sh
```diff
@@ -8,6 +8,47 @@ umask 002
 log_file="sentry_install_log-`date +'%Y-%m-%d_%H-%M-%S'`.txt"
 exec &> >(tee -a "$log_file")
 
+# Sentry SaaS uses stock Yandex ClickHouse, but they don't provide images that
```
hubertdeng123, line 11: Is there a specific reason to move the logic from `detect-platform.sh` here? Thought it might be potentially easier to read if the logic existed in another file

install/error-handling.sh
```diff
@@ -1,10 +1,34 @@
...
+if [[ -f .reporterrors ]]; then
+  if [[ "$(cat .reporterrors)" == "yes" ]]; then
+    export REPORT_ERRORS=1
+  else
+    export REPORT_ERRORS=0
+  fi
+else
+  echo "Would you like to opt-in to error monitoring for the Sentry installer?"
+  echo "This helps us catch and fix errors when installing Sentry."
+  echo "We may retain your OS username, IP address, and installer log, for 30 days."
```
hubertdeng123, line 16: tiny: `We may retain your OS username, IP address, and installer log for 30 days.`

install/error-handling.sh
```diff
@@ -1,10 +1,34 @@
...
+          No ) export REPORT_ERRORS=0; echo "no" > .reporterrors; break;;
+      esac
+  done
+fi
+
 # Courtesy of https://stackoverflow.com/a/2183063/90297
 trap_with_arg() {
   func="$1" ; shift
   for sig ; do
-    trap "$func $sig "'$LINENO' "$sig"
```
hubertdeng123, line 7: is this because our error monitoring will be more detailed so this is not needed?

install/error-handling.sh
BYK: Ah, didn't notice that :D I'd still prefer something like `installer` or something but of course, up to you :)

install/_lib.sh
BYK: If it works on Mac, I think it will work on other places. This might have been my scar from a different mapping for my drives.

install/error-handling.sh
chadwhitacre: I like `installer`, fits if we ever get around to adding run-time monitoring in a separate project.

install/error-handling.sh
chadwhitacre: Good use case for [GitHub suggestions](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request), @hubertdeng123. ;)

```suggestion
  echo "We may retain your OS username, IP address, and installer log for 30 days."
```

chadwhitacre:
> Can non-sentry collaborator (myself :) ) view these logs and if yes, how?

I am checking with Legal to see what we can offer here. At the very least Sentry has a feature for sharing public links to specific events, I don't have a ton of experience with it but once we start collecting data we can experiment with this to see if it's anonymized enough. Here's a contrived example (I'm not finding mention of this in the docs):

https://self-hosted.getsentry.net/share/issue/81cdf4322c86473e8fd7662aac0f8d51/

install/error-handling.sh
ethanhs: Yes, the stack trace includes the lineno for each part of the stack up to where the error occured.

install/_lib.sh
ethanhs: I know these mounts can be quite finicky so understandable 😅

install/_lib.sh
ethanhs: It is needed below for the `send_event` function, so it had to be moved here.

install/error-handling.sh
ethanhs: Done!

install/error-handling.sh
```diff
@@ -1,10 +1,36 @@
 echo "${_group}Setting up error handling ..."
 
+export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentry.net/3'
+export SENTRY_ORG=sentry
```
chadwhitacre, line 4: Can we use `self-hosted`? Makes it easier to distinguish between ours and SaaS.

Screen Shot 2022-09-07 at 2 48 51 PM

hubertdeng123 (approved):

install/_lib.sh
chadwhitacre: Why do we need to define `send_event` in `_lib.sh` instead of in `error-handling.sh`?

install/error-handling.sh
```diff
@@ -15,9 +41,30 @@ cleanup () {
...
+          local src=${BASH_SOURCE[$i]}
+          local lineno=${BASH_LINENO[$i-1]}
+          local funcname=${FUNCNAME[$i]}
+          printf -v traceback '%s\n' "$traceback${indent//a/-}>$src:$funcname:$lineno"
+      done
+    fi
+    echo "$traceback"
+
+    if [ "$REPORT_ERRORS" == 1 ]; then
+      local traceback_hash=$(echo -n $traceback | docker run --rm busybox md5sum | cut -d' ' -f1)
```
chadwhitacre, line 65: Will this hash the same if I hit the error from an installation at `/home/sentry` and a second at `/usr/local/sentry`?

install/error-handling.sh
```diff
@@ -15,9 +41,30 @@ cleanup () {
     return 0;
   fi
   DID_CLEAN_UP=1
-
+  local retcode=$?
+  local cmd="${BASH_COMMAND}"
   if [[ "$1" != "EXIT" ]]; then
-    echo "An error occurred, caught SIG$1 on line $2";
+    set +o xtrace
```
chadwhitacre, line 47: Any need to unset after we're done?

install/error-handling.sh
```diff
@@ -15,9 +41,30 @@ cleanup () {
...
+    printf -v err '%s' "Error in ${BASH_SOURCE[1]}:${BASH_LINENO[1]}."
```
chadwhitacre, line 48: What's `-v`? Is it portable?

install.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -eE
```
chadwhitacre, line 2: What is capital E again?

chadwhitacre: Ethan says it's for subshells and sourced files or something.

install.sh
```diff
@@ -11,9 +11,10 @@ source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other
 
 # Pre-flight. No impact yet.
 source parse-cli.sh
-source detect-platform.sh
 source dc-detect-version.sh
 source error-handling.sh
+# We set the trap at the top level so that we get better tracebacks.
+trap_with_arg cleanup ERR INT TERM EXIT
```
chadwhitacre, line 17: What if there's an error in parse cli or dc-detect?

chadwhitacre: Our error handling depends on Docker specifics, so dc detect should come before.

Parse CLI is pretty thin, fine to leave it where it is.

install/error-handling.sh
chadwhitacre: Sets a variable, should be fine.

install/_lib.sh
```diff
@@ -8,6 +8,47 @@ umask 002
...
+    local sentry_cli="docker run --rm -v \$(pwd):/work -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
+  else
+    if ! command -v sentry-cli &> /dev/null; then
+        echo "sentry-cli could not be found, please install it"
+        exit 1
+    fi
+    local sentry_cli=sentry-cli
+  fi
+  command pushd .. > /dev/null
+  $sentry_cli send-event --no-environ -f "$1" -m "$2" --logfile $log_file
```
chadwhitacre, line 48: Can we send full logfile as an attachment?

install/error-handling.sh
```diff
@@ -15,9 +41,30 @@ cleanup () {
...
+    printf '%s\n%s\n' "$err" "$cmd_exit"
+    local stack_depth=${#FUNCNAME[@]}
+    local traceback=""
+    if [ $stack_depth -gt 2 ]; then
+      for ((i=$(($stack_depth - 1)),j=1;i>0;i--,j++)); do
+          local indent="$(yes a | head -$j | tr -d '\n')"
+          local src=${BASH_SOURCE[$i]}
+          local lineno=${BASH_LINENO[$i-1]}
+          local funcname=${FUNCNAME[$i]}
+          printf -v traceback '%s\n' "$traceback${indent//a/-}>$src:$funcname:$lineno"
```
chadwhitacre, line 59:
```suggestion
          printf -v traceback '%s\n' "$traceback${indent//a/ }→ $src:$funcname:$lineno"
```

install/error-handling.sh
chadwhitacre: xtrace enhances the debugging info we get from bash, it's fine to run dc stop (the last thing we do) with this still on

install/_lib.sh
chadwhitacre: No, not trivially given current sentry-cli. Will have to be a follow-on. :-/

chadwhitacre (changes requested):
A few open items. Looking good so far!

install/error-handling.sh
ethanhs: It seems that I would have to re-create the sentry instance to change this. I can do that if you want.

install/error-handling.sh
ethanhs: It will, I confirmed this is just the file name not the full path.

install/error-handling.sh
ethanhs: Done, changed on both sides, forgot there was a setting for this 😓

install/_lib.sh
ethanhs: Moved to `error-handling.sh`

install.sh
```diff
@@ -11,9 +11,11 @@ source "$(dirname $0)/install/_lib.sh"  # does a `cd .../install/`, among other
 
 # Pre-flight. No impact yet.
 source parse-cli.sh
-source detect-platform.sh
 source dc-detect-version.sh
+source detect-platform.sh
```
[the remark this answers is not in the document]
ethanhs: I misremembered where it was, and there was no simple revert, oops! Now corrected.

install/error-handling.sh
```diff
@@ -1,10 +1,51 @@
...
+    export REPORT_ERRORS=1
+  else
+    export REPORT_ERRORS=0
+  fi
+else
+  echo "Would you like to opt-in to error monitoring for the Sentry installer?"
+  echo "This helps us catch and fix errors when installing Sentry."
+  echo "We may collect and retain your OS username, IP address, and installer log, for 30 days."
+  echo "Your information is solely used to improve the installer and is subject to our privacy policy[1]."
+  echo "You may change your preference at any time by deleting the '.reporterrors' file."
```
chadwhitacre, line 33: This is a lie-ish. The `.reporterrors` file ends up under `install/`. I would expect to be in the distribution root. No?

ethanhs: Yeah, this is a good point, since the installer mostly runs things in the `install` folder. I'll change it so it places it in the distribution root.

install/error-handling.sh
```diff
@@ -1,10 +1,51 @@
...
+        exit 1
+    fi
+    local sentry_cli=sentry-cli
+  fi
+  command pushd .. > /dev/null
+  $sentry_cli send-event --no-environ -f "$1" -m "$2" --logfile $log_file
+  command popd > /dev/null
+}
+
+if [[ -f ../.reporterrors ]]; then
```
chadwhitacre, line 22: This is insufficient, because we aren't guaranteed to be in the `install` directory.

https://github.com/getsentry/self-hosted/blob/013519a46c696438016a1577615d47f1ec677212/install/_lib.sh#L11-L16

I made a PR at #1686 to propose a significant copy update, and it includes a fix using `$basedir`. What do you think?

chadwhitacre:
Trying out the error reporter (with #1686) ...

```diff
diff --git a/install/check-latest-commit.sh b/install/check-latest-commit.sh
index 7f35660..334b942 100644
--- a/install/check-latest-commit.sh
+++ b/install/check-latest-commit.sh
@@ -1,5 +1,7 @@
 echo "${_group}Checking for latest commit ... "
 
+echo "Oh bother!"; exit 1
+
 # Checks if we are on latest commit from github if it is running from master branch
 if [[ -d "../.git" && "${SKIP_COMMIT_CHECK:-0}" != 1 ]]; then
   if [[ $(git branch | sed -n '/\* /s///p') == "master" ]]; then

```

```
$ ./install.sh 
▶ Parsing command line ...

▶ Detecting Docker platform
Detected Docker platform is linux/arm64

▶ Initializing Docker Compose ...

Found a .reporterrors file. What does it say? yes

▶ Checking for latest commit ... 
Oh bother!
$
```

Will it [show up](https://self-hosted.getsentry.net/organizations/self-hosted/issues/)? 🧐

chadwhitacre:
```diff
diff --git a/install/check-latest-commit.sh b/install/check-latest-commit.sh
index 7f35660..e9248c3 100644
--- a/install/check-latest-commit.sh
+++ b/install/check-latest-commit.sh
@@ -1,5 +1,7 @@
 echo "${_group}Checking for latest commit ... "
 
+false "Oh bother!"
+
 # Checks if we are on latest commit from github if it is running from master branch
 if [[ -d "../.git" && "${SKIP_COMMIT_CHECK:-0}" != 1 ]]; then
   if [[ $(git branch | sed -n '/\* /s///p') == "master" ]]; then

```

```
$ ./install.sh 
▶ Parsing command line ...

▶ Detecting Docker platform
Detected Docker platform is linux/arm64

▶ Initializing Docker Compose ...

Found a .reporterrors file. What does it say? yes

▶ Checking for latest commit ... 
Error in check-latest-commit.sh:19.
'[[ "$DID_CLEAN_UP" -eq 1 ]]' exited with status 0
-> ./install.sh:main:19
--> check-latest-commit.sh:source:3

Event dispatched: 6ee22387-6644-4fcd-873b-638d6a08c6f6
Cleaning up...
$
```

chadwhitacre:
Hrm, a couple issues there:

1. `Error in check-latest-commit.sh:19.` does not exist, there is no line 19 in that file (even with the patch). From the traceback it looks like lineno is coming from the `main.sh` frame?
2. `check-latest-commit.sh:source:3` is the correct line, but the message is wrong, I expect to see `false "Oh bother\!"` as the message, not `[[ "$DID_CLEAN_UP" -eq 1 ]]`

Looks like the traceback code needs some work.

chadwhitacre (changes requested):
Seeing unexpected behavior in traceback code. Comments in main thread.

install/error-handling.sh
```diff
@@ -1,23 +1,121 @@
...
   if [[ "$DID_CLEAN_UP" -eq 1 ]]; then
     return 0;
   fi
   DID_CLEAN_UP=1
-
+  local retcode=$?
   if [[ "$1" != "EXIT" ]]; then
-    echo "An error occurred, caught SIG$1 on line $2";
+    set +o xtrace
+    printf -v err '%s' "Error in ${BASH_SOURCE[1]}:${BASH_LINENO[0]}."
```
hubertdeng123, line 99: > 1. `Error in check-latest-commit.sh:19.` does not exist, there is no line 19 in that file (even with the patch). From the traceback it looks like lineno is coming from the `main.sh` frame?

The `${BASH_LINENO[$i]}` is the line number in the source file `${BASH_SOURCE[$i+1]}` as seen [here](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html)

install/error-handling.sh
```diff
@@ -1,23 +1,121 @@
...
   for sig ; do
-    trap "$func $sig "'$LINENO' "$sig"
+    trap "$func $sig" "$sig"
   done
 }
 
 DID_CLEAN_UP=0
 # the cleanup function will be the exit point
 cleanup () {
+  local cmd="${BASH_COMMAND}"
```
hubertdeng123, line 91: > 2. `check-latest-commit.sh:source:3` is the correct line, but the message is wrong, I expect to see `false "Oh bother\!"` as the message, not `[[ "$DID_CLEAN_UP" -eq 1 ]]`

cmd here was being overwritten by line 92 beforehand
```
if [[ "$DID_CLEAN_UP" -eq 1 ]]; then
```

hubertdeng123:
This is the result I get now:
```
./install.sh
▶ Parsing command line ...

▶ Detecting Docker platform
Detected Docker platform is linux/arm64

▶ Initializing Docker Compose ...

Found a .reporterrors file. What does it say? yes

▶ Checking for latest commit ... 
Error in check-latest-commit.sh:3.
'false "Oh bother!"' exited with status 0
-> ./install.sh:main:19
--> check-latest-commit.sh:source:3

Event dispatched: ccb89503-3b67-4f65-b492-c1b56c2c7832
Cleaning up...
```

chadwhitacre:
Why is it saying the exit status is 0, though? That indicates success. 🤔

```
$ false
$ echo $?
1
$
```

install/error-handling.sh
chadwhitacre: Good find! 👍

chadwhitacre:
Fixed in fee3d5105c168c995321512eeec3216082fcdcd4. ;)

[Here it is in Sentry](https://self-hosted.getsentry.net/organizations/self-hosted/issues/3/?project=3):

Screen Shot 2022-09-09 at 10 54 09 AM

Screen Shot 2022-09-09 at 10 53 43 AM

chadwhitacre:
Not sure the IP is going to be that valuable if it tends to be from an internal block.

hubertdeng123:
> Fixed in [fee3d51](https://github.com/getsentry/self-hosted/commit/fee3d5105c168c995321512eeec3216082fcdcd4). ;)
> 
> [Here it is in Sentry](https://self-hosted.getsentry.net/organizations/self-hosted/issues/3/?project=3):
> 

ah nice catch

install/error-handling.sh
```diff
@@ -1,23 +1,111 @@
...
+    if ! command -v sentry-cli &> /dev/null; then
+        echo "sentry-cli could not be found, please install it"
+        exit 1
+    fi
+    local sentry_cli=sentry-cli
+  fi
+  command pushd .. > /dev/null
+  $sentry_cli send-event --no-environ -f "$1" -m "$2" --logfile $log_file
+  command popd > /dev/null
+}
```
chadwhitacre, line 20: This is confusing to me. Why do we pass `SENTRY_DSN` in the dockerized case but not `SENTRY_ORG` or `SENTRY_PROJECT`?

This also seems like not a great experience on non-linux/amd64, to hard exit during error handling vs. warning early during setup-error-handling and silently passing here. Of course better would be to make it work, how hard would it be to build sentry-cli for ARM? Are we close? Link?

chadwhitacre: I'm also not sure of the advantage of hashing the traceback ourselves vs. sending the traceback as the fingerprint and letting Sentry hash.

Do we want separate status codes to hash together as they do now? It seems a little weird to have variable messages in events that are grouped together.

variation

hubertdeng123: I think we should have sentry-cli for ARM

https://docs.sentry.io/product/cli/installation/#docker-image

hubertdeng123: a bit confused, is there a reason why we're using the dockerized case for `linux/amd64`?

hubertdeng123: It's unclear to me why we're not passing the rest of the env variables too.

hubertdeng123: actually we only have a pre built image for amd64 😢 

https://hub.docker.com/search?q=sentry-cli

hubertdeng123: I'll answer my own question, dockerized case is preferred or else sentry-cli will need to be installed manually and what you mentioned above chad

install/error-handling.sh
```diff
@@ -1,23 +1,111 @@
 echo "${_group}Setting up error handling ..."
 
+export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentry.net/3'
+export SENTRY_ORG=self-hosted
+export SENTRY_PROJECT=installer
+
+function send_event {
+  if [[ $DOCKER_PLATFORM == "linux/amd64" ]]; then
+    local sentry_cli="docker run --rm -v \$(pwd):/work -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
```
hubertdeng123, line 9: notes here after running on ubuntu linux/amd64

This line errors out
```
docker: Error response from daemon: create $(pwd): "$(pwd)" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
```

install/error-handling.sh
```diff
@@ -1,23 +1,111 @@
 echo "${_group}Setting up error handling ..."
 
+export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentry.net/3'
+export SENTRY_ORG=self-hosted
+export SENTRY_PROJECT=installer
+
+function send_event {
+  if [[ $DOCKER_PLATFORM == "linux/amd64" ]]; then
+    local sentry_cli="docker run --platform linux/amd64 --rm -v $basedir:/work -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
```
hubertdeng123, line 9: Needed to change `$(pwd)` to `$basedir` here to find the logfile

install/error-handling.sh
hubertdeng123: Resolved now that \$(pwd) -> $basedir

hubertdeng123:
This is the sentry event sent from my ubuntu setup:

https://self-hosted.getsentry.net/organizations/self-hosted/issues/3/?query=is%3Aunresolved

hubertdeng123:
Behavior on `amd` and `arm` platform is tested and looks good to me now.

We can put something on the backlog to try and dockerize non `amd` platforms as a future item.

chadwhitacre:
I've updated the non-amd64 case in de504d8c9fb696c2df7a33cf0c454a3477488a0f. Here are the cases we need to check when an error is present (`false 'Oh bother!'` as above):

| platform | sentry-cli | .reporterrors | early warning (n/y) | prompt (n/y) | late warning (n/y) | attempt (n/y/f = failed) |
|:---|:---|:---|:---|:---|:---|:---|
| amd | no | ∅ |  |  |  |  |
| amd | no | no |  |  |  |  |
| amd | no | yes |  |  |  |  |
| amd | yes | ∅ |  |  |  |  |
| amd | yes | no |  |  |  |  |
| amd | yes | yes |  |  |  |  |
| arm | no | ∅ |  |  |  |  |
| arm | no | no |  |  |  |  |
| arm | no | yes |  |  |  |  |
| arm | yes | ∅ |  |  |  |  |
| arm | yes | no |  |  |  |  |
| arm | yes | yes |  |  |  |  |

My plan for testing this (short of writing a test script) would be to fiddle with the `check_for_sentry_cli` and `check_for_linux_amd64` to simulate the alternate cases. Like this:

```diff
diff --git a/install/error-handling.sh b/install/error-handling.sh
index 13902cf..17c1c0f 100644
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -7,10 +7,10 @@ export SENTRY_PROJECT=installer
 function check_for_linux_amd64 {
   # This is the only platform for which we have a build of sentry-cli.
   # I guess this could theoretically be dynamic (try to fetch a build).
-  test $DOCKER_PLATFORM == "linux/amd64"
+  test $DOCKER_PLATFORM == "linux/arm64"
 }
 function check_for_sentry_cli {
-  command -v sentry-cli &> /dev/null
+  command -v entry-cli &> /dev/null
 }
 function check_for_reportability {
   # I'm sure this isn't the most terse syntax, but it's the one I could get to work.
```

chadwhitacre:
I hate to say it but I think we should check for whether there is a `sentry-cli` Docker image dynamically rather than hard-coding it. That way we can ship this and then build a Docker image for arm64 at a later date decoupled from a code change and it will "just work."

hubertdeng123:
lol, just tried running the `docker run --rm -v $basedir:/work -e SENTRY_ORG=$SENTRY_ORG -e SENTRY_PROJECT=$SENTRY_PROJECT -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli` on my mac M1 and it is working...

Perhaps it was failing in the first place due to syntax issues and we never got around to verifying that. I don't even think we need to include separate logic for arm64

hubertdeng123:
well, pushed up changes! could use some assistance just to confirm that this works

chadwhitacre:
Here's what I get: 

```
$ ./install.sh 
▶ Parsing command line ...

▶ Detecting Docker platform
Detected Docker platform is linux/arm64

▶ Initializing Docker Compose ...

Found a .reporterrors file. What does it say? yes

▶ Checking for latest commit ... 
Error in check-latest-commit.sh:3.
'false 'Oh bother!'' exited with status 1
-> ./install.sh:main:19
--> check-latest-commit.sh:source:3

Unable to find image 'getsentry/sentry-cli:latest' locally
latest: Pulling from getsentry/sentry-cli
c7ed990a2339: Pulling fs layer
18d3cc651798: Pulling fs layer
3e6a010ae3c1: Pulling fs layer
4e81a4699a2d: Pulling fs layer
0c93296acec2: Pulling fs layer
4e81a4699a2d: Waiting
0c93296acec2: Waiting
18d3cc651798: Verifying Checksum
3e6a010ae3c1: Verifying Checksum
3e6a010ae3c1: Download complete
4e81a4699a2d: Verifying Checksum
4e81a4699a2d: Download complete
c7ed990a2339: Pull complete
18d3cc651798: Pull complete
3e6a010ae3c1: Pull complete
4e81a4699a2d: Pull complete
0c93296acec2: Verifying Checksum
0c93296acec2: Download complete
0c93296acec2: Pull complete
Digest: sha256:9c214c9b56a9292c2b1dc727e096bab185aa23bafaddb290a52377e7f5022157
Status: Downloaded newer image for getsentry/sentry-cli:latest
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Event dispatched: 13f97186-4ffa-4203-8ef6-ce5abdb7128d
Cleaning up...
$
```

What do you think @ethanhs? Is this okay to ship?

install/error-handling.sh
ethanhs: Yeah, we should pass SENTRY_ORG and SENTRY_PROJECT to docker as well, this was an oversight (added them after writing this I think 😢 )

Agreed testing up front for sentry-cli missing is better. I haven't gotten much information about sentry-cli's build system but that they are working on it, I can ask if there is a link to follow along on the changes.

ethanhs:
> What do you think @ethanhs? Is this okay to ship?

I think with one or two minor changes!

> Do we want separate status codes to hash together as they do now?

I think this is still an issue? But I'm not sure why it is happening tbh. I wanted to hash the traceback as a means of making the issues grouped based on traceback, but it seems that isn't happening...

chadwhitacre:
I do see [an event showing up](https://self-hosted.getsentry.net/organizations/self-hosted/issues/3/events/13f971864ffa42038ef6ce5abdb7128d/?project=3), so it appears to work despite the warning.

chadwhitacre:
> I think this is still an issue?

I'm not going to block on this. Generally the exit code will likely be stable, it was a bit of a fluke that it varied for us.

> But I'm not sure why it is happening tbh.

I think it's happening because the exit code is not included in the traceback, it only appears in `$cmd_exit`.

hubertdeng123:
To add more context for the warning 
```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

I get that for other images we use in `self-hosted` since we use other `linux/amd64` images, so I think this is fine

chadwhitacre:
We set `DOCKER_PLATFORM` though ...

https://github.com/getsentry/self-hosted/blob/0e1239c78f87f2b810576f6fc7327b6e8c908a85/install/detect-platform.sh#L18-L23

So why does it say, "and no specific platform was requested"?

chadwhitacre:
lolsob `DOCKER_PLATFORM` is unique to us, it's not a Docker envvar. 🙄

hubertdeng123:
What's also interesting is that I don't get that warning when pulling sentry-cli:latest

```
Unable to find image 'getsentry/sentry-cli:latest' locally
latest: Pulling from getsentry/sentry-cli
c7ed990a2339: Pulling fs layer
18d3cc651798: Pulling fs layer
3e6a010ae3c1: Pulling fs layer
4e81a4699a2d: Pulling fs layer
0c93296acec2: Pulling fs layer
4e81a4699a2d: Waiting
0c93296acec2: Waiting
18d3cc651798: Download complete
3e6a010ae3c1: Verifying Checksum
3e6a010ae3c1: Download complete
c7ed990a2339: Verifying Checksum
c7ed990a2339: Download complete
c7ed990a2339: Pull complete
18d3cc651798: Pull complete
3e6a010ae3c1: Pull complete
4e81a4699a2d: Verifying Checksum
4e81a4699a2d: Download complete
4e81a4699a2d: Pull complete
0c93296acec2: Verifying Checksum
0c93296acec2: Download complete
0c93296acec2: Pull complete
Digest: sha256:9c214c9b56a9292c2b1dc727e096bab185aa23bafaddb290a52377e7f5022157
Status: Downloaded newer image for getsentry/sentry-cli:latest
Event dispatched: 3cf243d5-19be-44c8-8c21-c486a5c1d6f3
Cleaning up...
```

ethanhs:
Ok then I think this is ready, @chadwhitacre want to give it the green check-mark?

chadwhitacre:
Oh wow @hubertdeng123 I wonder if the delta between your env and mine is related to what we're seeing on #1671?

I think we're good here though. We use `DOCKER_PLATFORM` to set `platform` in `docker-compose.yml`, but that does not influence the `sentry-cli` image.

ethanhs:
> Oh wow @hubertdeng123 I wonder if the delta between your env and mine is related to what we're seeing on https://github.com/getsentry/self-hosted/pull/1671?

This is what I was thinking as well.

chadwhitacre:
Okay, pushed one more commit 5df2ab645f186044e4af3b809b6ee7b334be990d to fail gracefully, if the `sentry-cli` image doesn't work for some reason let's find out early so we don't muddy the waters with a double failure during error handling.

I'm ready to merge if yinz are!

chadwhitacre (approved):

PR #1685, opened 2022-09-08
[WIP] Test upgrading sentry

ethanhs:
This turned out to be a bit simpler than I expected! (Famous last words)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      
+      - name: Install previous major version for testing upgrade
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -46,6 +46,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      
+      - name: Install previous major version for testing upgrade
```
BYK, line 50: I think you should `Get Compose` first.

Also, if we are doing this some of this should become obsolete:
https://github.com/getsentry/self-hosted/blob/013519a46c696438016a1577615d47f1ec677212/integration-test.sh#L23-L26

PR #1686, opened 2022-09-08
Update copy and fix .reporterrors location

chadwhitacre:
PR-ception. This PR updates the copy on #1679 (and fixes a bug in file pathing).

### Cache Hit

```
Found a .reporterrors file. What does it say? no
```

### Cache Miss

```

Hey, so ... we would love to find out when you hit an issue with this here
installer you are running. Turns out there is an app for that, called Sentry.
Are you okay with us sending info to Sentry when you run this installer?

  y / yes / 1
  n / no / 0

(Btw, we send this to our own self-hosted Sentry instance, not to Sentry SaaS,
so that we can be in this together.)

Here's the info we may collect:

  - OS username
  - IP address
  - install log
  - performance data

Thirty (30) day retention. No marketing. Privacy policy at sentry.io/privacy.

y or n? n

Understood. Your answer is cached in '.reporterrors', remove it to see this
prompt again.
```

or ...

```
y or n? y

Thank you. Your answer is cached in '.reporterrors', remove it to see this
prompt again.
```

### DIFF
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -19,26 +19,62 @@ function send_event {
   command popd > /dev/null
 }
 
-if [[ -f ../.reporterrors ]]; then
-  if [[ "$(cat ../.reporterrors)" == "yes" ]]; then
+reporterrors="$basedir/.reporterrors"
+if [[ -f $reporterrors ]]; then
+  echo -n "Found a .reporterrors file. What does it say? "
+  cat $reporterrors
+  if [[ "$(cat $reporterrors)" == "yes" ]]; then
     export REPORT_ERRORS=1
   else
     export REPORT_ERRORS=0
   fi
 else
-  echo "Would you like to opt-in to error monitoring for the Sentry installer?"
-  echo "This helps us catch and fix errors when installing Sentry."
-  echo "We may collect and retain your OS username, IP address, and installer log, for 30 days."
-  echo "Your information is solely used to improve the installer and is subject to our privacy policy[1]."
-  echo "You may change your preference at any time by deleting the '.reporterrors' file."
   echo
-  echo "[1] https://sentry.io/privacy/"
-  select yn in "Yes" "No"; do
-      case $yn in
-          Yes ) export REPORT_ERRORS=1; echo "yes" > ../.reporterrors; break;;
-          No ) export REPORT_ERRORS=0; echo "no" > ../.reporterrors; break;;
-      esac
+  echo "Hey, so ... we would love to find out when you hit an issue with this here"
+  echo "installer you are running. Turns out there is an app for that, called Sentry."
+  echo "Are you okay with us sending info to Sentry when you run this installer?"
+  echo
+  echo "  y / yes / 1"
+  echo "  n / no / 0"
+  echo
+  echo "(Btw, we send this to our own self-hosted Sentry instance, not to Sentry SaaS,"
+  echo "so that we can be in this together.)"
+  echo
+  echo "Here's the info we may collect:"
+  echo
+  echo "  - OS username"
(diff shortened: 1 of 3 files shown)


chadwhitacre:
CI is failing here because we're blocked on the new prompt. Why is #1679 _not_ failing? 🤔

install/error-handling.sh
```diff
@@ -19,26 +19,62 @@ function send_event {
   command popd > /dev/null
 }
 
-if [[ -f ../.reporterrors ]]; then
-  if [[ "$(cat ../.reporterrors)" == "yes" ]]; then
+reporterrors="$basedir/.reporterrors"
```
ethanhs, line 22: ah yeah, I forgot we defined basedir. It might be helpful to change variables we expect used elsewhere to follow UPPERCASE, which I think is the common convention (but that can be another PR).

install/error-handling.sh
```diff
@@ -19,26 +19,62 @@ function send_event {
...
+          echo -n "Understood."
+          ;;
+        *) yn="";;
+    esac
   done
+
+  echo " Your answer is cached in '.reporterrors', remove it to see this"
+  echo "prompt again."
+  echo
+  sleep 5
```
ethanhs, line 77: Why the sleep here? To give people time to read the above?

ethanhs:
> CI is failing here because we're blocked on the new prompt. Why is https://github.com/getsentry/self-hosted/pull/1679 not failing? 🤔

@chadwhitacre CI is not failing on the other PR because select will choose a default case if there is not input. Here you require input and don't handle running read without a tty without stdin. I intentionally thought about this failure mode ahead of time :)

install/error-handling.sh
chadwhitacre: Yup. Otherwise it would fly by and be easy to miss. This seemed the simplest solution vs. "Press any key to continue" or something more clever.

chadwhitacre:
> select will choose a default case if there is not input

TIL re: select. Which does it choose?

> Here you require input and don't handle running read without a tty without stdin.

Right, I figured the options we would offer would be:

- `echo yes > .reporterrors`
- `yes | ./install.sh`
- `./install.sh --report-errors=yes/no`

I think we want to force everyone to choose when we ship this, no? I don't think we want to default (one way or another), since a lot of people (probably?) are running `./install.sh` under automation.

install/error-handling.sh
chadwhitacre: Could do, not sure how consistent we are. I think of `UPPERCASE` as standard especially for user-facing envvars, `basedir` is internal to the installer.

chadwhitacre:
Potential fourth option:

- `REPORT_ERRORS=1`

hubertdeng123:
Would also prefer to force everyone to choose here or else if people are running the install under automation, they may never try to opt into reporting errors. Is it reasonable to edit the `./integration-test` to run `./install.sh` with the --report-errors flag or something of that sort?

ethanhs:
> TIL re: select. Which does it choose?

Ah actually it falls through and sets neither.

> Right, I figured the options we would offer would be:
>
>    echo yes > .reporterrors
>    yes | ./install.sh
>    ./install.sh --report-errors=yes/no
>
> I think we want to force everyone to choose when we ship this, no? I don't think we want to default (one way or another), since a lot of people (probably?) are running ./install.sh under automation.

That makes sense. Do you want to add that here?

chadwhitacre:
Let's see if 73577b134cae2a986ba5d28fb9ce4fdf3b5f6d08 works. I'd like to land this PR asap and pick up on #1679. We might decide we _want_ errors from CI in Sentry, if we can segment envs.

PR #1695, opened 2022-09-14
Default to not sending data to Sentry for now

### LINKED ISSUE #1693: Improve error handling handling in non-interactive case
With #1679 we provide a sub-par experience in the non-interactive case. We _do_ want to interrupt people and force them to choose yes or no for error handling, but we should notice when they're non-interactive and give them a more helpful error message vs. blocking forever (until reaped) on the `y or n` prompt.

chadwhitacre:
Follow-up to #1679 to relax the non-interactive experience, closes #1693.

The proposal here is to default to not sending data for the non-interactive case, but with a notice that we will be forcing a decision in an upcoming version. This gives folks three months to notice and adjust, and it gives us some time to clean up configuration a bit.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,7 +7,6 @@ steps:
       - "-c"
       - |
         git clone -b $BRANCH_NAME https://github.com/getsentry/self-hosted.git
-        echo 'no' > self-hosted/.reporterrors
         echo '{"version": "3.4", "networks":{"default":{"external":{"name":"cloudbuild"}}}}' > self-hosted/docker-compose.override.yml
     timeout: 60s
   - name: "gcr.io/$PROJECT_ID/docker-compose"
--- a/install/_lib.sh
+++ b/install/_lib.sh
@@ -8,6 +8,13 @@ umask 002
 log_file="sentry_install_log-`date +'%Y-%m-%d_%H-%M-%S'`.txt"
 exec &> >(tee -a "$log_file")
 
+# Thanks to https://tldp.org/LDP/abs/html/intandnonint.html
+if [[ "$(tty)" == "not a tty" ]]; then
+  PROMPTABLE=0
+else
+  PROMPTABLE=1
+fi
+
 # Work from /install/ for install.sh, project root otherwise
 if [[ "$(basename $0)" = "install.sh"  ]]; then
   cd "$(dirname $0)/install/"
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -43,9 +43,10 @@ else
     echo
     echo "Thirty (30) day retention. No marketing. Privacy policy at sentry.io/privacy."
     echo
-    echo "For now we are defaulting to opt-out, but our plan is to hard-require a choice"
-    echo "from you starting in version 22.12.0, because let's be honest, none of you will"
-    echo "act on this otherwise. To avoid disruption you can:"
+    echo "For now we are defaulting to not sending data, but our plan is to hard-require"
+    echo "a choice from you starting in version 22.10.0, because let's be honest, none of"
+    echo "you will act on this otherwise. To avoid disruption you can use one of these"
+    echo "flags:"
     echo
     echo "  echo no > /path/to/sentry/.reporterrors"
     echo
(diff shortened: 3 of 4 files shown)


chadwhitacre:
What do you think of this approach @Dherlou?

Dherlou:
@chadwhitacre 
My personal opinion is that data privacy matters should always be opt-in and not opt-out. I am from Germany, so I am used to work according to the EU GDPR, which seems to be in contrast to e.g. CCPA where it is the other way around and opt-out is typically the default.
That said, if you want to force users to make an explicit choice to a later point of time anyway, I would recommend just showing the message in non-interactive-mode and exit with status code 1. People can then respond to the resulting error in CI/CD pipelines by explicitly making a decision and implementing it in their CI/CD pipeline (e.g. by echoing 'yes' or 'no' in the '.reporterrors' as a build step or use the environment variable once implemented) instead of (probably) unknowingly sending their data by default without an explicit consent. I doubt that people will read the output of their pipelines as long as it 'still works'. :D

chadwhitacre:
> I doubt that people will read the output of their pipelines as long as it 'still works'. :D

I know I don't! 😜

> showing the message in non-interactive-mode and exit with status code 1

I think for tomorrow it's best to show the message but continue with the installation. When we add the exit in 22.12.0 we will be able to say "We gave you three months to respond." Better than abrupt change.

ethanhs:
> I think for tomorrow it's best to show the message but continue with the installation (defaulting to opt-out, of course). When we add the exit in 22.12.0 we will be able to say "We gave you three months to respond." Better than abrupt change.

I don't think there is any point in waiting 3 months, the main way people are going to notice this change is reading the changelog, so I think doing one release with it default opt-out, then making choosing opt-in or out mandatory in the next one is a good idea.

chadwhitacre:
> I don't think there is any point in waiting 3 months, the main way people are going to notice this change is reading the changelog, so I think doing one release with it default opt-out, then making choosing opt-in or out mandatory in the next release is a good idea.

Fair enough. changed to 22.10.0 in 109036d365de954de938980d8feea9ba7b9bf56e, and also stopped mis-using "opt-out." It's not "default to opt-out," it's that opt-in is default to not send. Next month it'll be neither opt-in nor opt-out, we'll force a choice.

hubertdeng123:
I'm guessing this PR will be going in before the others? Also agree that 1 month should be enough since it's not a change most people will even notice

chadwhitacre:
> this PR will be going in before the others

Yes, let's get this landed ASAP as it is the most vital to not making everyone mad tomorrow. :)

I've pushed 75c713e207dae3b55f5c8c971fbbbf19529feb88 so we can see this working in our two CI environments. I was seeing different behavior between the two with an earlier implementation of `PROMPTABLE` (GitHub provides stdin whereas GCP does not).

ethanhs (approved):
Looks like this is working! Just need to revert the test commit and this should be good I think?

hubertdeng123:
Should we revert the changes we made yesterday to the other cloudbuild.yaml files once this goes in? I think ideally those should also be changed

chadwhitacre:
Yup, looks good in both GH and GCP. 👍

https://github.com/getsentry/self-hosted/actions/runs/3055058465/jobs/4927684866#step:4:43

Screen Shot 2022-09-14 at 2 41 28 PM

https://console.cloud.google.com/cloud-build/builds/5b965c3c-980f-4eaf-8e52-401c7ab28c9b?project=sentryio

Screen Shot 2022-09-14 at 2 42 06 PM

chadwhitacre:
Reporting turned off again in CI in ea809d82e853da57c08c9121100452e70104b6ac. I think we can do something similar in `sentry` and `snuba`, sound right?

hubertdeng123:
Ah yeah, that's what I meant. Once we change completely from using `.reporterrors` we can probably go ahead in `sentry` and `snuba` to remove references to the `.reporterrors` file

chadwhitacre:
Turned on automerge. Better withdraw your ✅ if you don't want this merged, @ethanhs. 😄

ethanhs:
> Turned on automerge. Better withdraw your ✅ if you don't want this merged

Oh, but I do! :)

PR #1697, opened 2022-09-14
Switch from .reporterrors file to flag + envvar

chadwhitacre:
The `.reporterrors` file is a somewhat idiosyncratic way to configure the sentry installer. We use environment variables and CLI flags for other configuration, we don't have a proper configuration file for the installer itself (vs. the runtime config files in `sentry/`). This PR switches to using envvars and CLI flags to configure whether we report to Sentry.

This also expands the scope of the ask to allow for future expansion into runtime data collection. We're starting with just the installer but we don't want to have to go through the consent exercise again in the future if we don't have to, better to ask once up front here.

### DIFF
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -47,7 +47,7 @@ if [[ -z "${REPORT_SELF_HOSTED_ISSUES:-}" ]]; then
     echo "  --report-self-hosted-issues"
     echo "  --no-report-self-hosted-issues"
     echo
-    echo "or the REPORT_SELF_HOSTED_ISSUES environment variable:"
+    echo "or set the REPORT_SELF_HOSTED_ISSUES environment variable:"
     echo
     echo "  REPORT_SELF_HOSTED_ISSUES=1 to send data"
     echo "  REPORT_SELF_HOSTED_ISSUES=0 to not send data"
@@ -95,7 +95,7 @@ if [[ -z "${REPORT_SELF_HOSTED_ISSUES:-}" ]]; then
     echo "  --report-self-hosted-issues"
     echo "  --no-report-self-hosted-issues"
     echo
-    echo "or the REPORT_SELF_HOSTED_ISSUES environment variable:"
+    echo "or set the REPORT_SELF_HOSTED_ISSUES environment variable:"
     echo
     echo "  REPORT_SELF_HOSTED_ISSUES=1 to send data"
     echo "  REPORT_SELF_HOSTED_ISSUES=0 to not send data"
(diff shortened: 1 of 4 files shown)


ethanhs:
> The .reporterrors file is a somewhat idiosyncratic way to configure the sentry installer.

I wanted some way to persist the user's choice since in the non-interactive case, I imagined a user coming along, running the updated installer, making a choice, then completely forgetting the choice was made in the future. The persisted file would avoid users having to remember every time that they need to opt in or out of error collection. I expect if we don't prompt them every month users are more likely to stay opted-in to error collection.

Perhaps a more robust solution is a config file which is similar to a .env file (each option is just the environment variable setting?). e.g. they can put `REPORT_TO_SENTRY=1` in a file somewhere?

hubertdeng123:
I think we should also update the readme for this to include information here

hubertdeng123:
> Perhaps a more robust solution is a config file which is similar to a .env file (each option is just the environment variable setting?). e.g. they can put REPORT_TO_SENTRY=1 in a file somewhere?

What's the harm in using the .env file?

ethanhs:
> What's the harm in using the .env file?

The .env files is more meant for runtime, not install time. Also if a user does the following things could break:

- set option in .env
- override that option when running the installer e.g. via CLI
- docker will read from .env not CLI

Also, ideally we'd be able to persist the error collection choice of users, but updating the .env file would be harder to do.

hubertdeng123:
Ah yeah, I guess it's better to leave installation configs separate from runtime

install/error-handling.sh
```diff
@@ -48,31 +44,36 @@ else
     echo "you will act on this otherwise. To avoid disruption you can use one of these"
     echo "flags:"
     echo
-    echo "  echo no > /path/to/sentry/.reporterrors"
+    echo "  --report-to-sentry"
+    echo "  --hide-from-sentry"
     echo
-    echo "We'll probably also add an env var and/or CLI flag before then, too. See:"
+    echo "or the REPORT_TO_SENTRY environment variable:"
```
ethanhs, line 50:
```suggestion
    echo "or set the REPORT_TO_SENTRY environment variable:"
```

ethanhs (approved):

chadwhitacre:
Automerge enabled! Here we gooooo!

chadwhitacre:
I did some bikeshedding on the naming of the flags/vars by looking at parallels with `SENTRY_BEACON`, `SENTRY_SKIP_UPSTREAM_REPORTING` and https://docs.sentry.io/product/sentry-basics/key-terms/. I tried to come up with something that avoided the confusion of "Like, is Sentry forwarding my events upstream to their server?" No, we're sending issues that arise with your self-hosted instance itself. I also tried to be generic enough to cover future expansion as mentioned. "Issues" is generic and encompasses all sorts of things, from errors to performance to whatever else we come up with.

PR #1700, opened 2022-09-14
Fix traceback hash for error monitoring

hubertdeng123:
add `-i` flag to keep STDIN open even if not attached
We were hashing an empty string before 😬


ethanhs (approved):
Thanks!

PR #1703, opened 2022-09-16
Add postgres upgrade script from 9.6 to 14

hubertdeng123:
This PR adds an upgrade script for postgres from 9.6 to 14. 

https://github.com/getsentry/self-hosted/issues/1610

### DIFF
--- a/.env
+++ b/.env
@@ -13,3 +13,5 @@ WAL2JSON_VERSION=latest
 HEALTHCHECK_INTERVAL=30s
 HEALTHCHECK_TIMEOUT=60s
 HEALTHCHECK_RETRIES=5
+# Change the variable based on the Postgres version you are using
+POSTGRES_VERSION=9.6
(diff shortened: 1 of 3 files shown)


BYK:
Related: #1705. I think instead of introducing a postgres version variable, it would be better to introduce a [hard-stop](https://develop.sentry.dev/self-hosted/releases/#hard-stops) upgrade point, add the script there, and then remove it afterwards.

hubertdeng123:
Going to put this on the backlog for now to work on dogfooding, ran this script and it seems to work just fine. Introducing a hard-stop upgrade will take more time to investigate

aminvakil:
> Related: #1705. I think instead of introducing a postgres version variable, it would be better to introduce a [hard-stop](https://develop.sentry.dev/self-hosted/releases/#hard-stops) upgrade point, add the script there, and then remove it afterwards.

What's wrong with users being able to upgrade from any self-hosted version using 9.6 to any self-hosted version post merging this which uses 14? So we won't need a hard-stop right?

PR #1710, opened 2022-09-19
ref: Remove unused scripts and code

### LINKED ISSUE #1705: Chore: Remove old and obsolete migration scripts
### Problem Statement

About 2+ years ago, when we have not settled on [hard-stops](https://develop.sentry.dev/self-hosted/releases/#hard-stops) for upgrading we had to account for any possible old version to be able to upgrade them. Most of these scripts are now obsolete so should be removed.

### Solution Brainstorm

List of scripts that are obsolete:

1. https://github.com/getsentry/self-hosted/blob/master/install/upgrade-postgres.sh
  If you need this script, that means you are still on < 21.5.0 which you should upgrade to first and that version of the repo will have this script.
1. https://github.com/getsentry/self-hosted/blob/master/install/migrate-file-storage.sh
  Just like the Postgres upgrade above, this was also needed when coming from very old versions. It should not be needed after 21.5.0 and the script hasn't been modified since "the great refactor of #906" which happened on March 2021. The contents should be even older.
1. https://github.com/getsentry/self-hosted/blob/master/install/replace-tsdb.sh
  Same story above.
1. https://github.com/getsentry/self-hosted/blob/master/install/set-up-zookeeper.sh
1. These specific lines: https://github.com/getsentry/self-hosted/blob/bacecf4fedde3635d27433d00c833890af086fc7/install/turn-things-off.sh#L10-L12

BYK:
Fixes #1705


BYK:
Ping @chadwhitacre? :)

chadwhitacre:
@BYK From [here](https://github.com/getsentry/self-hosted/issues/1705#issuecomment-1251322065) ... 

> For (4) do we have an associated hard-stop defined?

The hard stop for all of these is 21.5.0, yes?

BYK:
>The hard stop for all of these is 21.5.0, yes?

Yup (just read my former self). The main theme here is mostly me trying to cover up my mistakes during the "continuous release" era 😅

chadwhitacre (approved):

chadwhitacre:
Approved and set to automerge.

Do we reference these scripts anywhere in documentation? If so we should remove there as well.

BYK:
>Do we reference these scripts anywhere in documentation? If so we should remove there as well.

Don't think so as these were "invisible" migrations.

PR #1715, opened 2022-09-21
Send full logs as an attachment to our dogfood instance

hubertdeng123:
This change allows us to support sending envelopes to our sentry dogfood instance. 

1. An envelope file is created with envelope headers
2. Log file contents are dumped into the file as the attachment payload
3. File is deleted

https://github.com/getsentry/team-ospo/issues/46


install/error-handling.sh
```diff
@@ -4,11 +4,36 @@ export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentr
...
+  # Add header to specify the event type of envelope to be sent
+  echo '{"type":"event"}' >> $envelope_file_path
+  # Add traceback message to event
+  echo '{"message":"'$traceback_escaped'","level":"error"}' >> $envelope_file_path
+  # Add attachment to the event
+  echo '{"type":"attachment","length":'$file_length',"content_type":"text/plain","filename":"install_log.txt"}' >> $envelope_file_path
+  cat "$basedir/$log_file" >> $envelope_file_path
+  # Send envelope and cleanup temporary file
   local sentry_cli="docker run --rm -v $basedir:/work -e SENTRY_ORG=$SENTRY_ORG -e SENTRY_PROJECT=$SENTRY_PROJECT -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
-  command pushd .. > /dev/null
```
hubertdeng123, line 9: I don't really see a need to do this especially since we're using $basedir in our `sentry_cli`. Maybe I'm missing something though

ethanhs: Yeah this is not needed anymore.

ethanhs:
I'm not sure I agree we should delete the envelope, if someone runs into an error, tries to fix it, runs again and hits the same error, we don't want 2 envelopes I don't think. I originally put the envelope file in /tmp so that they will get cleaned up periodically but while someone is working on an issue we don't send multiple.

hubertdeng123:
I see your point there, my thought for deleting the file was that it's possible to have the same traceback but a different error and in that case I think we should send both. What we could do is determine the event id using both the traceback hash and error. It actually is probably better that way since we have to scroll through all the events to see what kind of errors people are seeing on a certain line. Then we don't have to delete the file

ethanhs:
> it's possible to have the same traceback but a different error

Perhaps we should hash both together? If someone has the same traceback and a different error, I don't think those should be grouped together, they aren't always related.

ethanhs (approved):
Looks great!

PR #1718, opened 2022-09-22
Tag releases with sentry-cli

hubertdeng123:
uses sentry-cli to tag releases for us on our self-hosted instance

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,3 +31,17 @@ jobs:
           version: ${{ github.event.inputs.version }}
           force: ${{ github.event.inputs.force }}
           calver: true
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -31,3 +31,21 @@ jobs:
           version: ${{ github.event.inputs.version }}
           force: ${{ github.event.inputs.force }}
           calver: true
+      - name: Create new release in Self-hosted Sentry
```
ethanhs, line 34:
```suggestion
      - name: Create new release in self-hosted Sentry
```

ethanhs (approved):
One minor nit but otherwise looks good!

PR #1301, opened 2022-02-04
Pull relay explicitly to avoid garbage in creds

chadwhitacre:
Fixes #1297 🤞

PR #1722, opened 2022-09-26
Error handling unit tests

hubertdeng123:
https://github.com/getsentry/team-ospo/issues/47


_integration-test/error-handling.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "${_group}Testing send_event duplicate"
+SEND_EVENT_RESPONSE=$(send_event "12345123451234512345123451234512" Test exited with status 1)
+if [ "$SEND_EVENT_RESPONSE" != 'Looks like you'\''ve already sent this error to us, we'\''re on it :)' ]; then
+  exit -1
+fi
+echo "Pass."
+rm "$basedir/$log_file"
+echo "${_endgroup}"
+
+echo "${_group} Testing cleanup without minimizing downtime"
```
hubertdeng123, line 42: I'm not a huge fan of these test cases because technically I'm testing cleanup when there isn't an error code being thrown. It seems to be impossible to unit test by setting bash internal variables by overwriting `BASH_COMMAND`, `BASH_SOURCE`, `BASH_LINENO`, etc so don't think there's a better way to do this

_unit-test/error-handling-test.sh
```diff
@@ -0,0 +1,67 @@
...
+export MINIMIZE_DOWNTIME=1
+CLEANUP_RESPONSE=$(cleanup ERROR)
+if [ "$CLEANUP_RESPONSE" != 'Error in ./_integration-test/error-handling.sh:59.
+'\''local cmd="${BASH_COMMAND}"'\'' exited with status 0
+
+*NOT* cleaning up, to clean your environment run "docker compose stop".' ]; then
+  exit -1
+fi
+echo "Pass."
+echo "${_endgroup}"
```
ethanhs, line 67:
```suggestion
echo "${_endgroup}"

```
Missing newline at the end of the file.

hubertdeng123:
ah noticing our unit tests pipeline isn't doing anything actually

_unit-test/error-handling-test.sh
```diff
@@ -0,0 +1,67 @@
...
+export -f send_envelope
+echo "${_group}Testing initial send_event"
+export log_file="test_log.txt"
+echo "Test Logs" > "$basedir/$log_file"
+SEND_EVENT_RESPONSE=$(send_event "12345123451234512345123451234512" Test exited with status 1)
+if [ "$SEND_EVENT_RESPONSE" != 'Test Sending sentry-envelope-12345123451234512345123451234512' ]; then
+  exit -1
+fi
+ENVELOPE_CONTENTS=$(cat /tmp/sentry-envelope-12345123451234512345123451234512)
+if [ "$ENVELOPE_CONTENTS" !='{"event_id":"12345123451234512345123451234512","dsn":"https://19555c489ded4769978daae92f2346ca@self-hosted.getsentry.net/3"}
```
ethanhs, line 23: I think it might be better to have expected envelopes as files we can cat.

_unit-test/error-handling-test.sh
```diff
@@ -0,0 +1,67 @@
...
+{"type":"attachment","length":10,"content_type":"text/plain","filename":"install_log.txt"}
+Test Logs' ]; then
+  exit -1
+fi
+echo "Pass."
+echo "${_endgroup}"
+
+echo "${_group}Testing send_event duplicate"
+SEND_EVENT_RESPONSE=$(send_event "12345123451234512345123451234512" Test exited with status 1)
+if [ "$SEND_EVENT_RESPONSE" != 'Looks like you'\''ve already sent this error to us, we'\''re on it :)' ]; then
```
ethanhs, line 35: Are you able to use double quotes here so we don't end up with `'\''`?

ethanhs (commented):
I do agree I think a more robust testing framework might be better, but I think there are some changes that will make these tests easier to understand.

I think this will do for now, especially since mocking BASH_LINENO and such is difficult.

_unit-test/error-handling-test.sh
hubertdeng123: done!

install/error-handling.sh
```diff
@@ -20,7 +26,7 @@ function send_envelope() {
   # If we haven't sent the envelope file, make it and send to Sentry
   # The format is documented at https://develop.sentry.dev/sdk/envelopes/
   # Grab length of log file, needed for the envelope header to send an attachment
-  local file_length=$(stat -f %z < "$basedir/$log_file")
+  local file_length=$(wc -c < "$basedir/$log_file" | awk '{print $1}')
```
hubertdeng123, line 29: turns out we do need `wc -c`

https://www.cyberciti.biz/faq/howto-bash-check-file-size-in-linux-unix-scripting/

The syntax for stat is different for different platforms and was failing during our unit tests. I checked to make sure that wc -c works

hubertdeng123:
sorry for all the messy commits, will squash all of those messages LOL

_unit-test/create-docker-volumes-test.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 source "$(dirname $0)/_test_setup.sh"
+set -ex
 
-expected=7
+expected=14
```
hubertdeng123, line 5: test has been failing for some time now:

Our docker volumes should be
```
sentry-clickhouse
sentry-data
sentry-kafka
sentry-postgres
sentry-redis
sentry-self-hosted_sentry-clickhouse-log
sentry-self-hosted_sentry-kafka-log
sentry-self-hosted_sentry-nginx-cache
sentry-self-hosted_sentry-secrets
sentry-self-hosted_sentry-smtp
sentry-self-hosted_sentry-smtp-log
sentry-self-hosted_sentry-zookeeper-log
sentry-symbolicator
sentry-zookeeper
```

_unit-test/ensure-relay-credentials-test.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+source "$(dirname $0)/_test_setup.sh"
+source dc-detect-version.sh
```
hubertdeng123, line 3: git compare here isn't working properly, but this line was added and is needed since this test was failing due to changes months ago with the addition of `$dc` and `$dcr`. 

https://github.com/getsentry/self-hosted/issues/1289
https://github.com/getsentry/self-hosted/issues/1301

install/error-handling.sh
ethanhs: Did you test with logs that have the ▶️ symbol?

hubertdeng123: yeah I did, works locally for me

_unit-test/create-docker-volumes-test.sh
hubertdeng123: Ok I take this back, expected number of volumes is 7. This test will fail locally though due to these containers that are brought up with docker compose, which isn't included in `create-docker-volumes.sh`
```
sentry-self-hosted_sentry-clickhouse-log
sentry-self-hosted_sentry-kafka-log
sentry-self-hosted_sentry-nginx-cache
sentry-self-hosted_sentry-secrets
sentry-self-hosted_sentry-smtp
sentry-self-hosted_sentry-smtp-log
sentry-self-hosted_sentry-zookeeper-log
```

hubertdeng123: Test was failing due to `docker volume ls --quiet | grep '^sentry-.*' | wc -l` failing when there is are no docker volumes listed throwing an exception.

hubertdeng123: I just changed the file to be more flexible so that it uses regex to determine if a list of expected volumes are in the actual docker volumes so that it passes both locally and in cli.

chadwhitacre (commented):
- I'd rather have seen three separate PRs - one to bring the bad tests back online, one to add the test we want, one to refactor the filesystem layout of the tests
- I'm fine with the filesystem layout changes if you and @ethanhs prefer that style. In that case I think `snapshots` should be nested under `_unit-tests`, also could you not have to name things `-test`? Can you only run executable files e.g.?
- The test for the error handling looks good.

hubertdeng123:
Fair point, I can separate this PR out

hubertdeng123:
I actually prefer naming things `-test` as they were done before as it gives an easy way to differentiate the files when they're being searched up. Is there a different naming convention you were thinking about? Right now only executable `.sh` files are being run.

chadwhitacre:
>  Is there a different naming convention you were thinking about?

I was thinking you could do:

```
_unit-test/create-docker-volumes.sh
```

since the `-test` is redundant with the directory name (it was useful when the test files were next to the files they were testing, Go-style). Then you could use [executability](https://stackoverflow.com/questions/7812324/finding-executable-files-using-ls-and-grep) as the condition for selecting test scripts to run. Not a big deal either way.

hubertdeng123:
Gotcha, not a big deal for me either but I do like having the `-test` there just because I'm not a fan when files are named the same thing. Especially because I've opened the wrong file in the past before in that scenario during a search so I'll keep that there. I'll add the executability part though!

hubertdeng123:
For the separated PR's, they will be merged in this order. 
([Fix failing unit tests](https://github.com/getsentry/self-hosted/pull/1728)) ->
([Reorganize unit test layout](https://github.com/getsentry/self-hosted/pull/1729)) ->
This PR

_unit-test/error-handling-test.sh
```diff
@@ -0,0 +1,51 @@
...
+# mock send_envelope
+send_envelope() {
+  echo "Test Sending $1"
+}
+
+export -f send_envelope
+echo "Testing initial send_event"
+export log_file="test_log.txt"
+echo "Test Logs" > "$basedir/$log_file"
+SEND_EVENT_RESPONSE=$(send_event "12345123451234512345123451234512" Test exited with status 1)
```
chadwhitacre, line 17: Quotes around err msg to preserve across calls

install/ensure-relay-credentials-test.sh
```diff
@@ -1,24 +1,25 @@
 #!/usr/bin/env bash
 source "$(dirname $0)/_test_setup.sh"
+source dc-detect-version.sh
 
-cfg="../relay/config.yml"
-creds="../relay/credentials.json"
+cfg_file="../relay/config.yml"
+creds_file="../relay/credentials.json"
```
chadwhitacre, line 6: For the record there is a `creds` defined in `dc-detect-version.sh` that we don't want to clobber. `cfg` changed to match for style.

snapshots/sentry-envelope-12345123451234512345123451234512
```diff
@@ -0,0 +1,5 @@
+{"event_id":"12345123451234512345123451234512","dsn":"https://19555c489ded4769978daae92f2346ca@self-hosted.getsentry.net/3"}
+{"type":"event"}
+{"message":"Test exited with status 1","level":"error"}
+{"type":"attachment","length":10,"content_type":"text/plain","filename":"install_log.txt"}
+Test Logs
```
chadwhitacre, line 5: Can we move this under `_unit-test`?

hubertdeng123: Yep, will do that but will have to merge my other PR into this branch

chadwhitacre (approved):

PR #1728, opened 2022-09-29
Fix failing unit tests

hubertdeng123:
This PR both fixes the failing unit tests and fixes our pipeline to correctly fail unit tests that are not passing.

### DIFF
--- a/install/create-docker-volumes-test.sh
+++ b/install/create-docker-volumes-test.sh
@@ -18,7 +18,7 @@ sentry-zookeeper"
 
 before=$(get_volumes)
 
-test "$before" == "" || test "$before" == $expected_volumes
+test "$before" == "" || test "$before" == "$expected_volumes"
 
 source create-docker-volumes.sh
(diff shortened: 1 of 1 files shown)


install/create-docker-volumes-test.sh
```diff
@@ -1,20 +1,23 @@
...
+"
 
 source create-docker-volumes.sh
 source create-docker-volumes.sh
 source create-docker-volumes.sh
 
-test $(count) -eq $expected
+docker_volumes=$(docker volume ls --quiet | grep '^sentry-.*')
+for expected_volume in $expected_volumes; do
+  [[ "$docker_volumes" =~ "$expected_volume" ]] || exit 1
```
chadwhitacre, line 20: Per video chat let's preserve semantics here while now checking full strings vs. count, i.e., not just contains but:

- exact match of the set of volumes (consider sorting for determinacy)
- or no created volumes (to account for first run)

Issue #1706, opened 2022-09-16
Turn things off script erroring out


hubertdeng123:
### Problem Statement

Seeing an exception being thrown during the `./turn-things-off.sh` script with this error:

`$dc down -t $STOP_TIMEOUT --rmi local --remove-orphans`

aminvakil:
Can the error be shared too?

hubertdeng123:
```
▶ Turning things off ...

Container sentry-self-hosted-snuba-subscription-consumer-events-1  Stopping
Container sentry-self-hosted-snuba-replacer-1  Stopping
Container sentry-self-hosted-snuba-api-1  Stopping
...
Container sentry-self-hosted-zookeeper-1  Removing
Container sentry-self-hosted-zookeeper-1  Removed


Error response from daemon: removal of container 21136ea8655aba78b4a259809b256fafd22dbd8cb3c9f94694605a9add7af70a is already in progress
Error in turn-things-off.sh:8.
'$dc down -t $STOP_TIMEOUT --rmi local --remove-orphans' exited with status 1

-> ./install.sh:main:23
--> turn-things-off.sh:source:8

'$dc down -t $STOP_TIMEOUT --rmi local --remove-orphans' exited with status 1
```

hubertdeng123:
Seeing this one a good amount!

ethanhs:
We do run docker compose down twice: https://github.com/getsentry/self-hosted/blob/bacecf4fedde3635d27433d00c833890af086fc7/install/turn-things-off.sh#L8-L12 so my guess is one of the containers is in the removal state when we run the down command again.

aminvakil:
I don't think `docker-compose down` gets executed twice, as if the `COMPOSE_PROJECT_NAME` you want to do stuff with does not exist, it won't do anything,

A simple reproducer with any `docker-compose.yml` you want:
```
$ docker-compose up -d
[+] Running 1/0
 ⠿ Container httpd  Running                                                                                                                                                                                                            0.0s
$ COMPOSE_PROJECT_NAME=nonexistent docker-compose down -v --remove-orphans
$
$ docker-compose down -v --remove-orphans
[+] Running 2/2
 ⠿ Container httpd      Removed                                                                                                                                                                                                        1.3s
 ⠿ Network nfs_default  Removed                                                                                                                                                                                                        0.1s
$ COMPOSE_PROJECT_NAME=nonexistent docker-compose down -v --remove-orphans
$
```
[...]

PR #1732, opened 2022-09-30
Fix minimum version requirements for docker and docker compose

hubertdeng123:
The function vergte here doesn't correctly use the return code in order to be used in version comparison. As a result, users are running older versions of docker compose. The issue with the largest number of events in our dogfood instance seems to be a result of folks using an older version of docker compose, then erroring out later during the turn things off script.

This fixes the issue here:
https://github.com/getsentry/self-hosted/issues/1706

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -13,14 +13,14 @@ if [[ -z "$DOCKER_VERSION" ]]; then
   exit 1
 fi
 
-if [[ "$(vergte ${DOCKER_VERSION//v} $MIN_DOCKER_VERSION)" ]]; then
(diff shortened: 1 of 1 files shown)


BYK (approved):

install/check-minimum-requirements.sh
```diff
@@ -13,14 +13,14 @@ if [[ -z "$DOCKER_VERSION" ]]; then
   exit 1
 fi
 
-if [[ "$(vergte ${DOCKER_VERSION//v} $MIN_DOCKER_VERSION)" ]]; then
+if [[ "$(vergte ${DOCKER_VERSION//v} $MIN_DOCKER_VERSION)" -eq 1 ]]; then
```
hubertdeng123, line 16: Equal comparison is needed for this to succeed

ethanhs (approved):
Thanks!

PR #1733, opened 2022-10-03
Check to see if docker compose exists, else error out

hubertdeng123:
One of our most common errors that we're seeing on self-hosted occurs when docker-compose doesn't exist. Our version check then throws an exception.

https://self-hosted.getsentry.net/organizations/self-hosted/issues/16/?query=is%3Aunresolved

This PR adds logic to ensure that we're able to find the version of docker compose and output an error msg to the user when docker compose does not exist.

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -25,7 +25,7 @@ if [[ -z "$COMPOSE_VERSION" ]]; then
   exit 1
 fi
 
-if [[ "$(vergte ${COMPOSE_VERSION//v} $MIN_COMPOSE_VERSION)" ]]; then
+if [[ "$(vergte ${COMPOSE_VERSION//v} $MIN_COMPOSE_VERSION)" -eq 1 ]]; then
   echo "FAIL: Expected minimum $dc_base version to be $MIN_COMPOSE_VERSION but found $COMPOSE_VERSION"
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


ethanhs:
Can you also add a check that the docker command works as well? I noticed that is a common issue for users (e.g. when the daemon isn't running etc)

install/check-minimum-requirements.sh
```diff
@@ -7,8 +7,9 @@ function vergte () { printf "%s\n%s" $1 $2 | sort --version-sort --check=quiet -
 
 
 
-DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
-if [[ -z "$DOCKER_VERSION" ]]; then
+if [[ $(docker version --format '{{.Server.Version}}') ]]; then
+  DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
+else
```
BYK, line 12: How about this to avoid repeating?

```suggestion
DOCKER_VERSION=$(docker version --format '{{.Server.Version}}' || echo '')
if [[ -z "$DOCKER_VERSION" ]]; then
```

install/check-minimum-requirements.sh
```diff
@@ -19,7 +20,13 @@ if [[ "$(vergte ${DOCKER_VERSION//v} $MIN_DOCKER_VERSION)" ]]; then
 fi
 echo "Found Docker version $DOCKER_VERSION"
 
-COMPOSE_VERSION=$($dc_base version --short)
+if [[ $($dc_base version --short) ]]; then
+  COMPOSE_VERSION=$($dc_base version --short)
+else
```
ethanhs, line 25:
```suggestion
COMPOSE_VERSION=$($dc_base version --short || echo '')
if [[ -z "$COMPOSE_VERSION" ]]; then
```
Same as above, I think this should be changed to  be consistent if you change above to follow BYK's suggestion.

hubertdeng123:
Changes implemented!

ethanhs (approved):

PR #1737, opened 2022-10-04
Refactor sub-scripts to use functions and consistent indentation

ethanhs:
In this PR the sub-scripts are refactored to use functions which will improve the quality of tracebacks we get.

I'm opening this as a draft to get a better view of the diff.

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -3,19 +3,19 @@ echo "${_group}Checking minimum requirements ..."
 source "$(dirname $0)/_min-requirements.sh"
 
 # Check the version of $1 is greater than or equal to $2 using sort. Note: versions must be stripped of "v"
-function vergte () {
-  printf "%s\n%s" $1 $2 | sort --version-sort --check=quiet --reverse; echo $?;
+function vergte() {
+  printf "%s\n%s" $1 $2 | sort --version-sort --check=quiet --reverse
+  echo $?
 }
-
 
 function check_docker_version {
   DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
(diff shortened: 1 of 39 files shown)


install/check-minimum-requirements.sh
```diff
@@ -3,54 +3,68 @@ echo "${_group}Checking minimum requirements ..."
 source "$(dirname $0)/_min-requirements.sh"
 
 # Check the version of $1 is greater than or equal to $2 using sort. Note: versions must be stripped of "v"
-function vergte () { printf "%s\n%s" $1 $2 | sort --version-sort --check=quiet --reverse; echo $?; }
-
-
-
-DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')
```
ethanhs, line 10: I decided to split these into multiple functions, I thought that would be clearer.

install/create-docker-volumes.sh
```diff
@@ -1,11 +1,16 @@
 echo "${_group}Creating volumes for persistent storage ..."
 
-echo "Created $(docker volume create --name=sentry-clickhouse)."
-echo "Created $(docker volume create --name=sentry-data)."
-echo "Created $(docker volume create --name=sentry-kafka)."
-echo "Created $(docker volume create --name=sentry-postgres)."
-echo "Created $(docker volume create --name=sentry-redis)."
-echo "Created $(docker volume create --name=sentry-symbolicator)."
-echo "Created $(docker volume create --name=sentry-zookeeper)."
+function create_docker_volume {
```
ethanhs, line 3: I also refactored this into multiple calls to the same function.

ethanhs:
I think this is too big to review, so I will instead:
1. make a PR to add pre-commit
2. make a PR with the function-izing changes

PR #1738, opened 2022-10-04
Add pre-commit config

ethanhs:
This adds pre-commit to run mvdan's sh formatter on our scripts to form a consistent style. We should at some point add shell lint to check for code quality.

We also run this in CI to maintain the quality.

I split this into 3 commits for easier review:
1. add the config
2. run the config
3. add CI config


install/_lib.sh
```diff
@@ -45,7 +48,7 @@ fi
 # A couple of the config files are referenced from other subscripts, so they
 # get vars, while multiple subscripts call ensure_file_from_example.
 function ensure_file_from_example {
-  if [[ -f "$1" ]]; then
+  if [[ -f $1 ]]; then
```
hubertdeng123, line 51: actually a little worried about pre-commits modifying this, could be a problem if $1 is a string with spaces in it. I noticed this at least when sending envelopes

hubertdeng123: Sometimes we actually want to wrap that in quotes

install/error-handling.sh
```diff
@@ -138,40 +145,41 @@ if [ "$REPORT_SELF_HOSTED_ISSUES" == 1 ]; then
...
-  fi;
-fi;
+  fi
+fi
 
 # Courtesy of https://stackoverflow.com/a/2183063/90297
 trap_with_arg() {
-  func="$1" ; shift
-  for sig ; do
+  func="$1"
```
hubertdeng123, line 153: hmmmm, unless it only removes the quotes in the if statements?

install/_lib.sh
hubertdeng123: https://unix.stackexchange.com/questions/86800/difference-between-quoting-variables-in-shell-script-if-statements

Apparently if $1 is null we'll have an issue if it's not in quotes too

ethanhs: This was due to passing `--simplify`, which I have now removed. Let me know if you notice anything else amiss.

_integration-test/run.sh
```diff
@@ -111,10 +115,9 @@ declare -a EVENT_TEST_STRINGS=(
   '"title":"a failure"'
   '"object":"42"'
 )
-for i in "${EVENT_TEST_STRINGS[@]}"
-do
+for i in "${EVENT_TEST_STRINGS[@]}"; do
   echo "Testing '$i'..."
-  echo "$EVENT_RESPONSE" | grep "$i[,}]" >& /dev/null
+  echo "$EVENT_RESPONSE" | grep "$i[,}]" >&/dev/null
```
hubertdeng123, line 120: everything looks good to me now

Just one minor thing, is there a way to disable removing the space here? I actually prefer that, if not it's fine too

hubertdeng123 (approved):

_integration-test/run.sh
ethanhs: I don't think so :/

aminvakil (approved):

PR #1739, opened 2022-10-05
Improve output of tracebacks

ethanhs:
This PR improves the quality of the error messages we get, so that it is clearer where they occure (instead of `source:lineno` we now get `function:lineno`, which is hopefully clearer).

Note that I chose to not functionalize things that get run before we catch errors, there's no point.

### DIFF
--- a/install/bootstrap-snuba.sh
+++ b/install/bootstrap-snuba.sh
@@ -1,6 +1,10 @@
 echo "${_group}Bootstrapping and migrating Snuba ..."
 
-$dcr snuba-api bootstrap --no-migrate --force
-$dcr snuba-api migrations migrate --force
+function bootstrap_snuba {
+  $dcr snuba-api bootstrap --no-migrate --force
+  $dcr snuba-api migrations migrate --force
+}
+
+bootstrap_snuba
 
 echo "${_endgroup}"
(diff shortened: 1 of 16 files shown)


chadwhitacre (approved):
Approving, but I'm not sure of the value. The function names are the same as the filenames, and we do get the filename in the traceback. :shrug:

ethanhs:
> Approving, but I'm not sure of the value. The function names are the same as the filenames, and we do get the filename in the traceback

@chadwhitacre 

Yeahh I am not sure either. As memory serves when we talked about this originally, you were unhappy with the errors being `source:32` and not `function:32` at the end of the traceback. If you don't mind the current state then I'd suggest we just close this because I'd rather not make the installer more complicated than we need (but I will leave that decision up to you). At least we got https://github.com/getsentry/self-hosted/pull/1738 out of this work :)

chadwhitacre:
> At least we got https://github.com/getsentry/self-hosted/pull/1738 out of this work :)

Yeah that's a win. :)

hubertdeng123:
I'm also fine with how it is right now

ethanhs:
Ok! I'll go ahead and close this now.

PR #1753, opened 2022-10-17
Enforce error reporting for self-hosted

hubertdeng123:
For 22.10.0, we must enforce error reporting


ethanhs:
Can you double check that (sentry, relay, snuba, maybe others) all set this in their e2e tests?

hubertdeng123:
relay, sentry, and snuba all have conditions to create a `.reporterrors` files with no in it 👍

ethanhs (approved):

PR #1768, opened 2022-10-18
Fix sentry release for dogfood instance

hubertdeng123:
uses the action-release/v1 to handle releases for dogfood since that's much easier!!
https://github.com/getsentry/action-release

Apparently, the version of the release is set earlier in the prepare release action and set as an environment variable `env.RELEASE_VERSION` so I'm using that. The `ignore_missing` and `ignore_empty` flags are set to avoid erroring out of release script.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,10 @@ jobs:
     name: Create release on self-hosted dogfood instance
     needs: release
     steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GH_RELEASE_PAT }}
+          fetch-depth: 0
       - uses: getsentry/action-release@v1
         env:
           SENTRY_ORG: self-hosted
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,3 +42,4 @@ jobs:
           environment: production
           version: ${{ env.RELEASE_VERSION }}
           ignore_empty: true
+          ignore_missing: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,16 +42,16 @@ jobs:
         run: |
           if [[ -n SENTRY_RELEASE ]]; then
             # Our releases are on 15th of each month so go back 14 days (a fortnight)
-            DATE_PART=$(date -d ''-1 fortnight'' +''%y.%-m'')
+            DATE_PART=$(date -d '-1 fortnight' '+%y.%-m')
             declare -i PATCH_VERSION=0
             while gh api --silent "repos/:owner/:repo/git/ref/tags/$DATE_PART.$PATCH_VERSION" 2>/dev/null; do
               PATCH_VERSION+=1
             done
             echo "RELEASE_VERSION=${DATE_PART}.${PATCH_VERSION}" >> $GITHUB_ENV;
-            curl -sL https://sentry.io/get-cli/ | bash
           else
             echo "RELEASE_VERSION=SENTRY_RELEASE" >> $GITHUB_ENV
           fi
+          curl -sL https://sentry.io/get-cli/ | bash
 
           # Create new Sentry release
           sentry-cli --auth-token $SENTRY_AUTH_TOKEN releases new -p $SENTRY_PROJECT $RELEASE_VERSION
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,31 +32,13 @@ jobs:
       #     force: ${{ github.event.inputs.force }}
       #     calver: true
       - name: Create new release in self-hosted Sentry
+        uses: getsentry/action-release@v1
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -32,20 +32,14 @@ jobs:
...
-          # Create new Sentry release
-          sentry-cli --auth-token $SENTRY_AUTH_TOKEN releases new -p $SENTRY_PROJECT $SENTRY_RELEASE
-          sentry-cli --auth-token $SENTRY_AUTH_TOKEN releases set-commits --auto $SENTRY_RELEASE
-          sentry-cli --auth-token $SENTRY_AUTH_TOKEN releases finalize $SENTRY_RELEASE
-
-          # Create new deploy for this Sentry release
-          sentry-cli --auth-token $SENTRY_AUTH_TOKEN releases deploys $SENTRY_RELEASE new -e production
+        with:
+          environment: production
+          version: ${{ env.RELEASE_VERSION }}
```
hubertdeng123, line 43: set in the prepare-release action, see [here](https://github.com/getsentry/self-hosted/actions/runs/3256633323/jobs/5347332615)

.github/workflows/release.yml
```diff
@@ -32,20 +32,14 @@ jobs:
           force: ${{ github.event.inputs.force }}
           calver: true
       - name: Create new release in self-hosted Sentry
+        uses: getsentry/action-release@v1
         env:
-          SENTRY_DSN: 'https://19555c489ded4769978daae92f2346ca@self-hosted.getsentry.net/3'
```
chadwhitacre, line 36: Is this set in `environment.production` as well?

hubertdeng123: We actually don't need this to set up releases since that's used for monitoring events for a specific project.

hubertdeng123: Just for reference, ran the entire workflow for this and it succeeded
https://self-hosted.getsentry.net/organizations/self-hosted/releases/22.10.1/?project=3

.github/workflows/release.yml
ethanhs: Using the prepare-release action is definitely the right way to go. However, the `RELEASE_VERSION` is set in the environment but if we are going to rely on it I think it might be better to change the `prepare-release` action to set it as an output. You can see an example of what the changes to the action would be here: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file

hubertdeng123: I'd rather not touch the `prepare-release` action since I'm not sure exactly where else it's used. Instead, I split up the  release job into two parts

1. Release a new version (`action-prepare-release`)
2. Create release on self-hosted dogfood instance (`action-release`)

I added an output for the 1st job that uses the release version from the `env.RELEASE_VERSION` that is set in `action-prepare-release`. The 2nd job uses the `needs` keyword to ensure the 1st job is run, and another `needs` keyword to ensure that the `release-version` output from the 1st job is available.

hubertdeng123: let me know if that resolves your concerns!

hubertdeng123: tested the run of the github action here:
https://github.com/getsentry/self-hosted/actions/runs/3292567306

.github/workflows/release.yml
```diff
@@ -24,28 +24,34 @@ jobs:
...
         uses: getsentry/action-prepare-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GH_RELEASE_PAT }}
         with:
           version: ${{ github.event.inputs.version }}
           force: ${{ github.event.inputs.force }}
           calver: true
-      - name: Create new release in self-hosted Sentry
+    outputs:
+      release-version: ${{ env.RELEASE_VERSION }}
```
ethanhs, line 36: Ah, I meant the actual action itself `getsentry/action-prepare-release` should be changed to set this as an output. Right now if the action is changed and `RELEASE_VERSION` is no longer set then our CI breaks.

hubertdeng123: Gotcha, I had a comment above that saying I'm not sure if we should change the github action itself if it's suited for just the self hosted's release use case. Is there a benefit to doing that as opposed to how I structured it to include an output for the prepare-release job? I feel like the behavior is pretty similar and I'd want to avoid changing `action-prepare-release` action if no other job uses the new output. 

If the action is changed and `RELEASE_VERSION` is no longer set (which many other steps in the action use so it should never happen), the dogfood-release will fail because it needs the release-version to be set in the outputs of prepare-release. If we set the output in the `action-prepare-release` the behavior will be similar here.

ethanhs: > I feel like the behavior is pretty similar and I'd want to avoid changing action-prepare-release action if no other job uses the new output.

The behavior is similar but implicitly relies on the internals of the action. I'd rather have a clear separation in case of a refactor or what-have you. It's better to be explicit rather than implicit on module boundaries especially, and actions are similar to code modules. I guess it won't matter since we pin to v1, so I think I am ok with this.

ethanhs (approved):

PR #1773, opened 2022-10-20
Do not send event on INT signal

hubertdeng123:
Right now, if someone presses ctrl+c, the event is sent to our dogfood instance. This removes that so we hopefully will see reduced noise

### DIFF
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -175,7 +175,7 @@ cleanup () {
       done
     fi
     echo "$traceback"
-    echo "$1"
+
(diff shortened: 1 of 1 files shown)


chadwhitacre:
Nice. Leave a comment for future us to know what this is doing? 😅

hubertdeng123:
> Nice. Leave a comment for future us to know what this is doing? 😅

done!

chadwhitacre (approved):

ethanhs (approved):
Great! Thanks for this fix.

PR #1775, opened 2022-10-22
Upload end of log as breadcrumbs, use exceptions and stacktrace

ethanhs:
In this PR, we change the send-event function to write out more information into the envelope we send via `sentry-cli`, particularly adding breadcrumbs with the end of the log and giving the stack trace in a format Sentry self-hosted can understand.

Fixes https://github.com/getsentry/team-ospo/issues/54 and https://github.com/getsentry/team-ospo/issues/53
Example issue with this formatting: https://self-hosted.getsentry.net/organizations/self-hosted/issues/227/?query=is%3Aunresolved

Also a cheat sheet for `jq` arguments:
 - `-c` collapses to 1 line
 - `-n` means null input (we're just constructing JSON, not mangling it)
 - `-s` is "slurp" JSON lines into an array.
 - `--arg` can be used to write object arguments
 -  `--argjson` is like `--arg`, but takes JSON expressions as the argument!

### DIFF
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -79,7 +79,7 @@ send_event() {
       --arg length $file_length \
       --arg content_type "text/plain" \
       --arg filename install_log.txt \
-      '$ARGS.named'
+      '{"type": $type,"length": $length|tonumber,"content_type": $content_type,"filename": $filename}'
   )
   echo "$attachment" >>$envelope_file_path
   cat $log_path >>$envelope_file_path
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -14,13 +14,12 @@ send_envelope() {
 
 generate_breadcrumb_json() {
   # Based on https://stackoverflow.com/a/1521498
-  while IFS="" read -r line || [ -n "$line" ]
-  do
+  while IFS="" read -r line || [ -n "$line" ]; do
     jq -n -c --arg message "$line" \
-             --arg category log \
-             --arg level info \
-             '$ARGS.named'
-  done < $log_path
+      --arg category log \
+      --arg level info \
+      '$ARGS.named'
+  done <$log_path
 }
 
 send_event() {
@@ -43,11 +42,11 @@ send_event() {
 
   # Add header for initial envelope information
   jq -n -c --arg event_id "$event_hash" \
-             --arg dsn "$SENTRY_DSN" \
-             '$ARGS.named' > $envelope_file_path
+    --arg dsn "$SENTRY_DSN" \
+    '$ARGS.named' >$envelope_file_path
   # Add header to specify the event type of envelope to be sent
-  echo '{"type":"event"}' >> $envelope_file_path
-  
+  echo '{"type":"event"}' >>$envelope_file_path
+
   # Next we construct the meat of the event payload, which we build up
   # inside out using jq
   # See https://develop.sentry.dev/sdk/event-payloads/
@@ -61,19 +60,21 @@ send_event() {
   # but first we need to make the stacktrace which goes in the exception payload
   frames=$(echo "$traceback_json" | jq -s -c)
   stacktrace=$(jq -n -c --argjson frames $frames '$ARGS.named')
-  exception=$(jq -n -c --arg "type" Error \
-                       --arg value "$traceback" \
-                       --argjson stacktrace $stacktrace \
-                       '$ARGS.named'
+  exception=$(
+    jq -n -c --arg "type" Error \
+      --arg value "$traceback" \
+      --argjson stacktrace $stacktrace \
+      '$ARGS.named'
   )
-  event_body=$(jq -n -c --arg level error \
-                        --argjson exception "{\"values\":[$exception]}" \
-                        --argjson breadcrumbs "{\"values\": $breadcrumbs}" \
-                        '$ARGS.named'
+  event_body=$(
+    jq -n -c --arg level error \
+      --argjson exception "{\"values\":[$exception]}" \
+      --argjson breadcrumbs "{\"values\": $breadcrumbs}" \
+      '$ARGS.named'
   )
-  echo "$event_body" >> $envelope_file_path
+  echo "$event_body" >>$envelope_file_path
   # Add attachment to the event
-  echo '{"type":"attachment","length":'$file_length',"content_type":"text/plain","filename":"install_log.txt"}' >> $envelope_file_path
+  echo '{"type":"attachment","length":'$file_length',"content_type":"text/plain","filename":"install_log.txt"}' >>$envelope_file_path
   cat $log_path >>$envelope_file_path
   # Send envelope
   send_envelope $envelope_file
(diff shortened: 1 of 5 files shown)


install/error-handling.sh
```diff
@@ -4,20 +4,32 @@ export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentr
...
+sentry_cli="docker run --rm -v /tmp:/work -e SENTRY_ORG=$SENTRY_ORG -e SENTRY_PROJECT=$SENTRY_PROJECT -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
+
 send_envelope() {
   # Send envelope
-  local sentry_cli="docker run --rm -v /tmp:/work -e SENTRY_ORG=$SENTRY_ORG -e SENTRY_PROJECT=$SENTRY_PROJECT -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
-  $sentry_cli send-envelope $envelope_file
+  $sentry_cli send-envelope "$envelope_file"
+}
+
+generate_breadcrumb_json() {
```
ethanhs, line 15: `sentry-cli` sends the logs as one line at a time of JSON data, so I am doing this to replicate that.

install/error-handling.sh
```diff
@@ -4,20 +4,32 @@ export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentr
...
+      --arg category log \
+      --arg level info \
+      '$ARGS.named'
+  done <$log_path
 }
 
 send_event() {
   # Use traceback hash as the UUID since it is 32 characters long
   local event_hash=$1
   local traceback=$2
```
ethanhs, line 28: I'm going to change this before merge as I'd like to rename this to "error" or something, because it isn't actually the traceback.

install/error-handling.sh
```diff
@@ -26,16 +38,44 @@ send_event() {
...
+  jq -n -c --arg event_id "$event_hash" \
+    --arg dsn "$SENTRY_DSN" \
+    '$ARGS.named' >$envelope_file_path
   # Add header to specify the event type of envelope to be sent
   echo '{"type":"event"}' >>$envelope_file_path
-  # Add traceback message to event
-  echo '{"message":"'$traceback_escaped'","level":"error"}' >>$envelope_file_path
+
+  # Next we construct the meat of the event payload, which we build up
+  # inside out using jq
```
ethanhs, line 51: It's a little unfortunate that this needs to be constructed inside out, but I wasn't sure of a better way to do this. Maybe could break some of this out into functions I guess, but not sure if it is worth.

ethanhs:
For your reading pleasure, here are the relevant docs pages:
https://develop.sentry.dev/sdk/envelopes/#data-model
https://develop.sentry.dev/sdk/event-payloads/
https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
https://develop.sentry.dev/sdk/event-payloads/exception/
https://develop.sentry.dev/sdk/event-payloads/stacktrace/

install/error-handling.sh
```diff
@@ -26,16 +38,44 @@ send_event() {
...
+  # First, create the breadcrumb payload
+  # https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
+  breadcrumbs=$(generate_breadcrumb_json | jq -s -c)
+  # Next we need the exception payload
+  # https://develop.sentry.dev/sdk/event-payloads/exception/
+  # but first we need to make the stacktrace which goes in the exception payload
+  frames=$(echo "$traceback_json" | jq -s -c)
+  stacktrace=$(jq -n -c --argjson frames $frames '$ARGS.named')
+  exception=$(
+    jq -n -c --arg "type" Error \
```
hubertdeng123, line 64: guessing here is where we could use the trap code? not sure if that's super helpful though because it would be either ERR or TERM

install/error-handling.sh
```diff
@@ -4,20 +4,32 @@ export SENTRY_DSN='https://19555c489ded4769978daae92f2346ca@self-hosted.getsentr
...
 send_envelope() {
   # Send envelope
-  local sentry_cli="docker run --rm -v /tmp:/work -e SENTRY_ORG=$SENTRY_ORG -e SENTRY_PROJECT=$SENTRY_PROJECT -e SENTRY_DSN=$SENTRY_DSN getsentry/sentry-cli"
-  $sentry_cli send-envelope $envelope_file
+  $sentry_cli send-envelope "$envelope_file"
+}
+
+generate_breadcrumb_json() {
+  # Based on https://stackoverflow.com/a/1521498
+  while IFS="" read -r line || [ -n "$line" ]; do
```
hubertdeng123, line 17: what does IFS stand for?

install/error-handling.sh
```diff
@@ -26,16 +38,44 @@ send_event() {
...
+  )
+  event_body=$(
+    jq -n -c --arg level error \
+      --argjson exception "{\"values\":[$exception]}" \
+      --argjson breadcrumbs "{\"values\": $breadcrumbs}" \
+      '$ARGS.named'
+  )
+  echo "$event_body" >>$envelope_file_path
   # Add attachment to the event
   echo '{"type":"attachment","length":'$file_length',"content_type":"text/plain","filename":"install_log.txt"}' >>$envelope_file_path
```
hubertdeng123, line 77: should we change this to use jq as well? just so everything has the same format

install/error-handling.sh
ethanhs: Yeah I don't think using the trap code would be useful, maybe I should just leave as is?

install/error-handling.sh
ethanhs: IFS is a special bash variable used for splitting strings.

install/error-handling.sh
ethanhs: Sure! Good point

install/error-handling.sh
hubertdeng123: I'm fine with that

install/error-handling.sh
hubertdeng123: just wanted to make sure this was done! since everything lgtm

hubertdeng123 (approved):

install/error-handling.sh
ethanhs: Good catch, pushed a change for this.

jq/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM debian:bullseye-slim
+
+LABEL MAINTAINER="oss@sentry.io"
+
+RUN set -x \
```
hubertdeng123, line 5: wasn't sure what this did until reading this:
https://opensource.com/article/20/5/optimize-container-builds

Can we link that as a comment?

hubertdeng123: or another link if you used another site

hubertdeng123: or is this common for dockerfiles and I just didn't know?
```
DEBIAN_FRONTEND=noninteractive
```
skips interactive prompt during installation
```
rm -rf /var/lib/apt/lists/*
```
Clear package list so it isn't stale

ethanhs: Yeah I do this whenever I use apt in Docker, to reduce size as much as possible, and I've seen it in other containers as well, e.g. https://github.com/docker-library/python/blob/230bb8c07e599a1e230b63895647e677d109c5c0/3.10/slim-bullseye/Dockerfile#L17

hubertdeng123:
lgtm

PR #1784, opened 2022-10-27
Add context line, error msg to envelope

hubertdeng123:
This PR let's us more easily filter our issues by the error msg
1. Adds the actual error message as the error msg instead of the line we fail on
2. Adds context line to display the line we fail on

Before:
Screen Shot 2022-10-27 at 3 41 27 PM
After:
Screen Shot 2022-10-27 at 3 41 43 PM

### DIFF
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -59,11 +59,11 @@ send_event() {
   # https://develop.sentry.dev/sdk/event-payloads/exception/
   # but first we need to make the stacktrace which goes in the exception payload
   frames=$(echo "$traceback_json" | jq -s -c)
-  stacktrace=$(jq -n -c --argjson frames $frames '$ARGS.named')
+  stacktrace=$(jq -n -c --argjson frames "$frames" '$ARGS.named')
   exception=$(
     jq -n -c --arg "type" Error \
       --arg value "$error_message" \
-      --argjson stacktrace $stacktrace \
+      --argjson stacktrace "$stacktrace" \
       '$ARGS.named'
   )
   event_body=$(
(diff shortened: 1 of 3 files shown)


install/error-handling.sh
```diff
@@ -188,6 +189,10 @@ cleanup() {
             --arg lineno "$lineno" \
             '{"filename": $filename, "function": $function, "lineno": $lineno|tonumber}'
         )
+        if [[ $i -eq 1 ]]; then
```
hubertdeng123, line 192: if we're in the stacktrace of the file we failed on, we can add a context line with the command run that failed

hubertdeng123:
On this note, wondering if we should remove the stacktrace in the breadcrumb since it's already displayed in the `Stack Trace` section above

install/error-handling.sh
```diff
@@ -188,6 +190,10 @@ cleanup() {
             --arg lineno "$lineno" \
             '{"filename": $filename, "function": $function, "lineno": $lineno|tonumber}'
         )
+        # If we're in the stacktrace of the file we failed on, we can add a context line with the command run that failed
+        if [[ $i -eq 1 ]]; then
+          JSON="{\"context_line\": \"$cmd\", ${JSON:1}"
```
ethanhs, line 195: I'd rather this be done with jq as well since jq will handle any escaping that is needed if `$cmd` has characters that would be a problem.

ethanhs:
> On this note, wondering if we should remove the stacktrace in the breadcrumb since it's already displayed in the Stack Trace section above

If you can do that in a way where the stacktrace still persists in the log that's fine by me!

hubertdeng123:
Removed stacktrace from breadcrumbs here:
https://self-hosted.getsentry.net/organizations/self-hosted/issues/232/?project=3&query=is%3Aignored&referrer=issue-stream

Also made it so we don't display the error msg as a `log info` as a duplicate

install/error-handling.sh
```diff
@@ -47,22 +48,19 @@ send_event() {
   # Add header to specify the event type of envelope to be sent
   echo '{"type":"event"}' >>$envelope_file_path
 
-  # Next we construct the meat of the event payload, which we build up
+  # First we construct the meat of the event payload, which we build up
```
ethanhs, line 51: I think this was correct as is, this isn't the first part of the payload we create.

hubertdeng123: fair enough, I was just bothered by two `Next` in a row, changed the other one to `Then` haha

ethanhs (approved):
Great!

PR #1789, opened 2022-10-31
ref: upgrade actions/setup-python to avoid set-output deprecation

asottile-sentry:
Committed via https://github.com/asottile/all-repos


chadwhitacre (approved):
I'll take your word for it.

Issue #1782, opened 2022-10-27
Deployment in kvm prompts me with an error:The CPU your machine is running on does not support the SSE 4.2 instruction set


Winspain:
### Self-Hosted Version

22.10.0

### CPU Architecture

x86_64

### Docker Version

19.03.9, build 9d988398e7

### Docker Compose Version

1.29.2

### Steps to Reproduce

1  ./install.sh
2  report an error

### Expected Result

correctly deploy

### Actual Result

The CPU your machine is running on does not support the SSE 4.2 instruction set

ethanhs:
Yeah, I suppose if you pass through your CPU that it won't show up as a KVM processor? Does the machine you are running on actually support SSE 4.2?

Winspain:
@ethanhs my host machine supports SSE 4.2,and i deployed successfully by skipping the IS_KVM codes
the machine can be correctly identified whether the kvm by the following command

```bash
[root@localhost ~]# dmesg |grep -i hypervisor
[    0.000000] Hypervisor detected: KVM

ethanhs:
This should remain open until we merge https://github.com/getsentry/self-hosted/pull/1790 or a similar PR.

PR #1790, opened 2022-10-31
Add --skip-requirements to install.sh and enable SUPPORTS_SSE42 override

erinaceous:
I've discovered that the check for SSE4.2 added in #361 fails on virtualization solutions other than KVM - possibly what the author of #1782 is having trouble with too.

Trying to install Sentry on a virtual machine running on a VMWare ESXi host ("ESXi 6.5 and later (VM version 13)"), it reports running on processors that _definitely_ [support SSE4.2](https://ark.intel.com/content/www/us/en/ark/products/120495/intel-xeon-gold-6154-processor-24-75m-cache-3-00-ghz.html):

```cat /proc/cpuinfo
model name	: Intel(R) Xeon(R) Gold 6154 CPU @ 3.00GHz
```

But this is not reported in their `flags`.

In this case, this is likely to do with our VM cluster having old servers which really don't have SSE4.2 and so [VWMare targets the lowest possible instruction set](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vcenterhost.doc/GUID-8B226625-4923-410C-B7AF-51BCD2806A3B.html) -- or, there's just a bug or a misconfigured option which means SSE4.2 isn't reported for CPUs which really do support it.

Obviously the right solution is "fix that underlying problem", but that's outside of our control (virtual machines managed by a central IT department, long lead time on updating the VMWare environment, etc).  

Luckily, I was able to run the executable mentioned in [this comment](https://github.com/ClickHouse/ClickHouse/issues/20#issuecomment-226853025) on a ClickHouse issue, which verified that yes, our virtual machine _does_ support SSE4.2, it just doesn't report it.

Bypassing the check in `check-minimum-requirements.sh`, I was able to install Sentry just fine on the VM.

Decided to clean it up and make it part of the install script - in two ways - being able to override `SUPPORTS_SSE42`, so you could run `SUPPORTS_SSE42=1 ./install.sh`, but also more generally skip requirements checking with an "I know, my set-up is weird, carry on anyway" flag to the script, `--skip-requirements`

### Legal Boilerplate

### DIFF
--- a/install/parse-cli.sh
+++ b/install/parse-cli.sh
@@ -18,6 +18,9 @@ Options:
                           branch of a \`self-hosted\` Git working copy.
  --skip-user-creation   Skip the initial user creation prompt (ideal for non-
                           interactive installs).
+ --skip-requirements    Skip checking that your environment meets the minimum
+                          requirements to run Sentry. Only do this if you know
+                          what you are doing.
  --report-self-hosted-issues
                         Report error and performance data about your self-hosted
                           instance upstream to Sentry. See sentry.io/privacy for
@@ -41,6 +44,7 @@ SKIP_USER_CREATION="${SKIP_USER_CREATION:-}"
 MINIMIZE_DOWNTIME="${MINIMIZE_DOWNTIME:-}"
 SKIP_COMMIT_CHECK="${SKIP_COMMIT_CHECK:-}"
 REPORT_SELF_HOSTED_ISSUES="${REPORT_SELF_HOSTED_ISSUES:-}"
+SKIP_REQUIREMENTS="${SKIP_REQUIREMENTS:-}"
 
 while (($#)); do
   case "$1" in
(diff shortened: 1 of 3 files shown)


aminvakil:
What do you think about changing this to `--skip-sse42-requirements` ?

ethanhs:
@erinaceous are you still interested in working on this PR? Totally fine if not but just wanted to check in where we're at here :)

erinaceous:
> What do you think about changing this to `--skip-sse42-requirements` ?

Yep, noted the thumbs up on that comment from all you maintainers, changed it to be more specific :+1:

install/parse-cli.sh
```diff
@@ -17,7 +17,11 @@ Options:
  --skip-commit-check    Skip the check for the latest commit when on the master
                           branch of a \`self-hosted\` Git working copy.
  --skip-user-creation   Skip the initial user creation prompt (ideal for non-
-                          interactive installs).
+                          	interactive installs).
+ --skip-sse42-requirements
+			Skip checking that your environment meets the
+			  requirements to run Sentry. Only do this if you know
```
ethanhs, line 23:
```suggestion
			  requirements to run Clickhouse on x86. Only do this if you know
```
The real need of this is clickhouse uses SSE 4.2 on x86 (not on ARM)

ethanhs (approved):
Thank you for updating this! Just one suggestion.

PR #1798, opened 2022-11-04
Disable cache when rebuilding Docker images

spawnia:
With the cache enabled, the steps in `enhance-image.sh` were not being run again on a new base image. This resulted in a new image being built without the proper additional libraries installed.

### Legal Boilerplate

### DIFF
--- a/install/build-docker-images.sh
+++ b/install/build-docker-images.sh
@@ -3,8 +3,8 @@ echo "${_group}Building and tagging Docker images ..."
 echo ""
 # Build any service that provides the image sentry-self-hosted-local first,
 # as it is used as the base image for sentry-cleanup-self-hosted-local.
-$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm web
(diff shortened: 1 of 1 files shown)


aminvakil:
https://github.com/getsentry/self-hosted/blob/97a4e5cce4fd74685ff0694ab2f67dd75cf5723f/sentry/Dockerfile#L4

This happens before executing `enhance-image.sh` and I guess if `enhance-image.sh` gets changed, this will executed again, right?

If the script in `enhance-image.sh` acts differently on different times of running it should be changed IMO.

In some cases this is not possible, then I would think of a putting a version or something in `enhance-image.sh` so I can bump it from time to time, so it will gets changed and docker cache gets invalidated and gets run again.

Using docker cache is pretty beneficial to many users, if above solutions does not work, I would say we should implement a switch which we can pass to `install.sh`, so it won't use docker cache if the argument gets passed to it.

spawnia:
I have investigated this issue further and found that the issue is that `enhance-image.sh` was not even copied to the container. Adding `--no-cache` did not actually resolve the problem, it has to do with `build-docker-images.sh` being run in the current working directory `./install`, thus not setting up the build context correctly. I am working on a solution.

PR #1800, opened 2022-11-07
Reference paths relative to the current script or project root

spawnia:
Before this PR:
- some scripts change the current working directory and use relative paths
- different approaches are taken to know which directory a script is running in
- paths are sometimes relative, sometimes absolute, sometimes traversing directories

After this PR:
- scripts do neither change nor care much about the current working directory
- a unified approach determines the directory of the current script
- paths are always relative to the project root

This should resolve an issue I already tried to fix with https://github.com/getsentry/self-hosted/pull/1798, where the contents of `./sentry` were not copied
into the built container image,
thus `enhance-image.sh` did not apply.

### Legal Boilerplate

### DIFF
--- a/_integration-test/ensure-customizations-not-present.sh
+++ b/_integration-test/ensure-customizations-not-present.sh
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 set -ex
 
-source "$(dirname $0)/../install/_lib.sh"
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source "$SCRIPT_DIR/../install/_lib.sh"
 
-source ../install/dc-detect-version.sh
+source "$PROJECT_ROOT/install/dc-detect-version.sh"
 
 # Negated version of ensure-customizations-work.sh, make changes in sync
 echo "${_group}Ensure customizations not present"
--- a/_unit-test/error-handling-test.sh
+++ b/_unit-test/error-handling-test.sh
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
-source "$(dirname $0)/_test_setup.sh"
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source "$SCRIPT_DIR/_test_setup.sh"
 
 export REPORT_SELF_HOSTED_ISSUES=1
 
-source error-handling.sh
+source "$PROJECT_ROOT/install/error-handling.sh"
 
 # mock send_envelope
 send_envelope() {
@@ -12,7 +14,7 @@ send_envelope() {
 
 export -f send_envelope
 echo "Testing initial send_event"
-export log_path="$basedir/test_log.txt"
+export log_path="$PROJECT_ROOT/test_log.txt"
 echo "Test Logs" >"$log_path"
 echo "Error Msg" >>"$log_path"
 breadcrumbs=$(generate_breadcrumb_json | sed '$d' | jq -s -c)
@@ -20,7 +22,7 @@ SEND_EVENT_RESPONSE=$(send_event "12345123451234512345123451234512" "Test exited
 rm "$log_path"
 test "$SEND_EVENT_RESPONSE" == 'Test Sending sentry-envelope-12345123451234512345123451234512'
 ENVELOPE_CONTENTS=$(cat /tmp/sentry-envelope-12345123451234512345123451234512)
-test "$ENVELOPE_CONTENTS" == "$(cat "$basedir/_unit-test/snapshots/sentry-envelope-12345123451234512345123451234512")"
+test "$ENVELOPE_CONTENTS" == "$(cat "$PROJECT_ROOT/_unit-test/snapshots/sentry-envelope-12345123451234512345123451234512")"
 echo "Pass."
 
 echo "Testing send_event duplicate"
--- a/install.sh
+++ b/install.sh
@@ -7,29 +7,30 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
-source "$(dirname $0)/install/_lib.sh" # does a `cd .../install/`, among other things
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source "$SCRIPT_DIR/install/_lib.sh"
 
 # Pre-flight. No impact yet.
-source parse-cli.sh
-source detect-platform.sh
-source dc-detect-version.sh
-source error-handling.sh
+source "$PROJECT_ROOT/install/parse-cli.sh"
+source "$PROJECT_ROOT/install/detect-platform.sh"
+source "$PROJECT_ROOT/install/dc-detect-version.sh"
+source "$PROJECT_ROOT/install/error-handling.sh"
 # We set the trap at the top level so that we get better tracebacks.
 trap_with_arg cleanup ERR INT TERM EXIT
-source check-latest-commit.sh
-source check-minimum-requirements.sh
+source "$PROJECT_ROOT/install/check-latest-commit.sh"
+source "$PROJECT_ROOT/install/check-minimum-requirements.sh"
 
 # Let's go! Start impacting things.
-source turn-things-off.sh
-source create-docker-volumes.sh
-source ensure-files-from-examples.sh
-source ensure-relay-credentials.sh
-source generate-secret-key.sh
-source update-docker-images.sh
-source build-docker-images.sh
-source install-wal2json.sh
-source bootstrap-snuba.sh
-source create-kafka-topics.sh
-source set-up-and-migrate-database.sh
-source geoip.sh
-source wrap-up.sh
+source "$PROJECT_ROOT/install/turn-things-off.sh"
+source "$PROJECT_ROOT/install/create-docker-volumes.sh"
+source "$PROJECT_ROOT/install/ensure-files-from-examples.sh"
+source "$PROJECT_ROOT/install/ensure-relay-credentials.sh"
+source "$PROJECT_ROOT/install/generate-secret-key.sh"
+source "$PROJECT_ROOT/install/update-docker-images.sh"
+source "$PROJECT_ROOT/install/build-docker-images.sh"
+source "$PROJECT_ROOT/install/install-wal2json.sh"
+source "$PROJECT_ROOT/install/bootstrap-snuba.sh"
+source "$PROJECT_ROOT/install/create-kafka-topics.sh"
+source "$PROJECT_ROOT/install/set-up-and-migrate-database.sh"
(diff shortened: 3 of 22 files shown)


aminvakil (approved):
I haven't checked each change, but this seems like a good improvement to me, thanks!

_unit-test/error-handling-test.sh
```diff
@@ -33,9 +35,9 @@ echo "Testing cleanup without minimizing downtime"
 export REPORT_SELF_HOSTED_ISSUES=0
 export MINIMIZE_DOWNTIME=''
 export dc=':'
-echo "Test Logs" >"$log_path"
+echo "Test Logs" >"$log_file"
 CLEANUP_RESPONSE=$(cleanup ERROR)
-rm "$log_path"
+rm "$log_file"
 test "$CLEANUP_RESPONSE" == 'Error in ./_unit-test/error-handling-test.sh:37.
```
aminvakil, line 41:
```suggestion
test "$CLEANUP_RESPONSE" == 'Error in ./_unit-test/error-handling-test.sh:24.
```

aminvakil (commented):
I tried reproducing your branch on a clean VM and this change fixed the issue.

_unit-test/error-handling-test.sh
spawnia: Thank you!

_integration-test/ensure-customizations-not-present.sh
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 set -ex
 
-source "$(dirname $0)/../install/_lib.sh"
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
```
ethanhs, line 4: I don't think this should need to be in each integration test, we have a test runner script where this might fit better. Also perhaps `INSTALL_SCRIPTS_DIR` is better, `SCRIPT_DIR` is a bit ambiguous.

_unit-test/error-handling-test.sh
```diff
@@ -33,10 +35,10 @@ echo "Testing cleanup without minimizing downtime"
 export REPORT_SELF_HOSTED_ISSUES=0
 export MINIMIZE_DOWNTIME=''
 export dc=':'
-echo "Test Logs" >"$log_path"
+echo "Test Logs" >"$log_file"
 CLEANUP_RESPONSE=$(cleanup ERROR)
-rm "$log_path"
-test "$CLEANUP_RESPONSE" == 'Error in ./_unit-test/error-handling-test.sh:37.
+rm "$log_file"
+test "$CLEANUP_RESPONSE" == 'Error in ./_unit-test/error-handling-test.sh:24.
```
ethanhs, line 41: My guess is the output here is what is causing the test failure?

hubertdeng123: +1

_integration-test/ensure-customizations-not-present.sh
spawnia: This expression evaluates to the path of the file it is written in. The whole point of it is not having to rely on some convention of how the scripts are arranged or from where they are called. I only ever use it once per script to refer to another script which then provides other constant values such as `PROJECT_ROOT`.

How else could we make references to other scripts more uniform and clear? Perhaps by always assuming we run from project root and define all paths as relative from there? That should solve the issue with the Docker build context too.

Alternatively, could we rename this variable to make its purpose clearer? `DIR_THIS_FILE_IS_IN`? Or maybe inline it, since it is only used once per file anyway?

install.sh
```diff
@@ -7,29 +7,30 @@ if [[ -n "$MSYSTEM" ]]; then
...
-source "$(dirname $0)/install/_lib.sh" # does a `cd .../install/`, among other things
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+source "$SCRIPT_DIR/install/_lib.sh"
 
 # Pre-flight. No impact yet.
-source parse-cli.sh
-source detect-platform.sh
-source dc-detect-version.sh
-source error-handling.sh
+source "$PROJECT_ROOT/install/parse-cli.sh"
```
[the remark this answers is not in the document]
hubertdeng123: Can we perform a cd into the install directory here without it interfering with the functionality of this PR? That should resolve that issue of reporting the entire path

.pre-commit-config.yaml
```diff
@@ -16,7 +16,6 @@
   hooks:
   - id: check-case-conflict
   - id: check-executables-have-shebangs
-    exclude: ^(install/|_unit-test/_test_setup.sh)
```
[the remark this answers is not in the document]
spawnia: Those files are not ever run as executables, they are only ever sourced - given they did not have a shebang, running them as executable would not have worked anyway. Instead of ignoring the check, I fixed it by removing the executable permission (chmod -x).

install.sh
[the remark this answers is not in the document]
spawnia: > Can we perform a cd into the install directory here without it interfering with the functionality of this PR?

Avoiding this `cd` is the entire point of this PR, it messes with the Docker build context. The integration tests did not catch this problem, because the `cd` only happens when actually running `install.sh`.

If you prefer relative paths, we could try an alternative solution I proposed in https://github.com/getsentry/self-hosted/pull/1800#discussion_r1017148876: always run from the project root and define all paths as relative from there. Should I give that a go?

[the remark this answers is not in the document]
spawnia: For reference, this was the offending code I removed that causes the behaviour change between tests and an actuall install: https://github.com/getsentry/self-hosted/blob/97a4e5cce4fd74685ff0694ab2f67dd75cf5723f/install/_lib.sh#L11-L16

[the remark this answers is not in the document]
hubertdeng123: It seems reasonable to run things from the project root and define all paths as relative from there. The primary concern for me is that the absolute path may include details for our error reporting that we don't want to be sending. Otherwise, this seems like a change that's fine with me, so thanks for putting this together.

_integration-test/ensure-customizations-not-present.sh
spawnia: I removed this from every script, now everything always runs from project root and uses relative paths.

install.sh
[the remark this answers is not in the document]
spawnia: Alright, I changed it so everything is always run from and relative to project root.

_unit-test/error-handling-test.sh
spawnia: I tried changing the line numbers again, but I am having trouble to debug this failure. When running `_unit-test/error-handling-test.sh` locally, I get the following error:

```
Cloning into '/tmp/sentry-self-hosted-test-sandbox.i0T'...
latest: Pulling from getsentry/sentry-cli
Digest: sha256:7b961ed50cffe9b40e110e7b87a8bf718a2dbfa55c413b0915af3d08c57d0101
Status: Image is up to date for getsentry/sentry-cli:latest
docker.io/getsentry/sentry-cli:latest

Testing initial send_event
jq - commandline JSON processor [version 1.5-1-a5b5cbe]
Usage: jq [options] <jq filter> [file...]

        jq is a tool for processing JSON inputs, applying the
        given filter to its JSON text inputs and producing the
        filter's results as JSON on standard output.
        The simplest filter is ., which is the identity filter,
        copying jq's input to its output unmodified (except for
        formatting).
        For more advanced filters see the jq(1) manpage ("man jq")
        and/or https://stedolan.github.io/jq

        Some of the options include:
         -c             compact instead of pretty-printed output;
         -n             use `null` as the single input value;
         -e             set the exit status code based on the output;
         -s             read (slurp) all inputs into an array; apply filter to it;
         -r             output raw strings, not JSON texts;
         -R             read raw strings, not JSON texts;
         -C             colorize JSON;
         -M             monochrome (don't colorize JSON);
         -S             sort keys of objects on output;
         --tab  use tabs for indentation;
         --arg a v      set variable $a to value <v>;
         --argjson a v  set variable $a to JSON value <v>;
         --slurpfile a f        set variable $a to an array of JSON texts read from <f>;
        See the manpage for more options.
jq: error: ARGS/0 is not defined at <top-level>, line 1:
$ARGS.named
jq: 1 compile error
jq: error: ARGS/0 is not defined at <top-level>, line 1:
$ARGS.named
jq: 1 compile error
```

spawnia: I suspect that the issue I am seeing locally is due to me using an older version of `jq` - I am currently on `jq-1.5.1-a5b5cbe`.

ethanhs: jq should be run in docker, is it picking up your local `jq` for some reason?

spawnia: Not sure what is happening locally - I was able to debug and fix the failing test through CI.

ethanhs:
Hm, not sure why Google Cloud build is failing, I'll look into it.

spawnia:
@ethanhs @hubertdeng123 I guess CI is only failing due to https://github.com/getsentry/self-hosted/pull/1803#issuecomment-1319079709? Given everything else passes, can you merge regardless so this fix can move forward?

ethanhs:
I'll take another look at this in a little bit, and if it looks good merge things.

spawnia:
I still think this is relevant and necessary and hope to see it merged some time soon.

spawnia:
@ethanhs @hubertdeng123 I have now updated this branch with the latest changes from `master` 4 times already, please review and consider merging 🙏

hubertdeng123 (changes requested):
Sorry for the late response @spawnia , did a re-review here. I'm pretty neutral on these changes here but seems generally fine

It also looks like the permissions of `detect-platform.sh` has been changed.

spawnia:
> It also looks like the permissions of detect-platform.sh has been changed.

Yup, this is intentional. The file is only able to be sourced from other scripts. Due to lack of a hashbang and the nature of its contents, it would not be able to be executed by itself.

install/build-docker-images.sh
```diff
@@ -4,9 +4,7 @@ echo ""
 # Build any service that provides the image sentry-self-hosted-local first,
 # as it is used as the base image for sentry-cleanup-self-hosted-local.
 $dcb --force-rm web
-for service in "$($dc config --services)"; do
-  $dcb --force-rm $service
-done
+$dcb --force-rm $($dc config --services)
```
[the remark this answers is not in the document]
spawnia: I made this change because my IDE warned me about https://github.com/koalaman/shellcheck/wiki/SC2066. I figured since `docker-compose build` accepts multiple arguments, I could simplify. For a quick validation of this, try the following two commands:

```sh
for file in "$(ls)"; do printf "File: $file\n"; done
for file in $(ls); do printf "File: $file\n"; done
```

Given the intention, I will now modify this example to actually build each service separately by removing quoting.

hubertdeng123 (approved):

spawnia:
Thank you @hubertdeng123

hubertdeng123:
Sorry for the long delay here @spawnia, thanks for submitting this PR 😁

PR #1803, opened 2022-11-10
Add .idea to .gitignore

spawnia:
Just like `.vscode/tags`, it is unnecessary IDE stuff.

### Legal Boilerplate


BYK (approved):

hubertdeng123 (approved):

hubertdeng123:
Seems like our pr check is failing on builds from outside contributors, I'll need to take a look into this

hubertdeng123:
@spawnia could you merge master in here? That should fix the failing test

hubertdeng123:
oh, interesting how since your root directory is `sentry-self-hosted` the test breaks, since it isn't `self-hosted`

spawnia:
> oh, interesting how since your root directory is `sentry-self-hosted` the test breaks, since it isn't `self-hosted`

Good discovery, I would not know how to fix it though.

hubertdeng123:
going to merge this because CI here should succeed, root directory is of a different name

PR #1759, opened 2022-10-17
Split post process forwarders

chadwhitacre:
For https://github.com/getsentry/sentry/issues/40120.

Ref:

- https://github.com/getsentry/snuba/pull/3250
- https://github.com/getsentry/sentry/pull/39988

PR #1817, opened 2022-11-16
Fix the problem that the commit-log-topic of the transactions consume…

klboke:
https://github.com/getsentry/self-hosted/pull/1759 After this pr,
post-process-forwarders split errors and transactions, this PR aims to Fix the problem that the commit-log-topic of the transactions consumer group is incorrect after post-process-forwarders split errors and transactions

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -254,7 +254,7 @@ services:
   # Kafka consumer responsible for feeding transactions data into Clickhouse
   snuba-transactions-consumer:
     <<: *snuba_defaults
-    command: consumer --storage transactions --consumer-group transactions_group --auto-offset-reset=latest --max-batch-time-ms 750 --commit-log-topic=snuba-commit-log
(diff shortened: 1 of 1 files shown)


klboke:
This problem will cause the following tasks to not work properly
```shell
run post-process-forwarder --commit-batch-size 1 --entity transactions --commit-log-topic=snuba-transactions-commit-log --synchronize-commit-group transactions_group
```

klboke:
@ethanhs @BYK @aminvakil  Maybe someone can take a look at this question

hubertdeng123 (approved):

lynnagara (approved):
Thanks for this!

Fyi - if you like you can actually omit the --commit-log-topic parameter altogether. The default will be the correct one.

hubertdeng123:
I like removing the `commit-log-topic` param better, making that change

PR #1820, opened 2022-11-16
(fix): Fix contributor PR e2e tests

hubertdeng123:
Our contributor e2e tests are failing since we're trying to clone a branch that doesn't exist on `https://github.com/getsentry/self-hosted.git`. Their branches only exist on their forked repos, so we need to reflect that in our e2e tests.

Also tested this out with contributor PR's [here](https://github.com/getsentry/self-hosted/pull/1821)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -6,7 +6,7 @@ steps:
       - "-e"
       - "-c"
       - |
-        git clone -b $BRANCH_NAME https://github.com/getsentry/self-hosted.git
+        git clone -b $BRANCH_NAME $_HEAD_REPO_URL
         echo '{"version": "3.4", "networks":{"default":{"external":{"name":"cloudbuild"}}}}' > self-hosted/docker-compose.override.yml
     timeout: 60s
   - name: "gcr.io/$PROJECT_ID/docker-compose"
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -6,7 +6,7 @@ steps:
       - "-e"
       - "-c"
       - |
-        git clone -b $BRANCH_NAME https://github.com/getsentry/self-hosted.git
+        git clone -b $BRANCH_NAME $_HEAD_REPO_URL
```
ethanhs, line 9: Hm, is this documented as supported? I am concerned about the starting underscore 😅

hubertdeng123: It has an underscore since it's only for pull request triggers, but our cloud build only runs on that so it should be fine

ethanhs: Ok great!

ethanhs (approved):

cloudbuild.yaml
```diff
@@ -3,10 +3,12 @@ steps:
     id: clone-and-configure
     entrypoint: "bash"
     args:
+      # $BRANCH_NAME and $_HEAD_REPO_URL usage is documented here
+      # https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values
```
chadwhitacre, line 7: Yay for docs.

chadwhitacre: And linking them in comments. :)

PR #1829, opened 2022-11-23
use postgres 14

sergei-iurchenko:
Let's upgrade to Postgres14
If is faster and more modern. 
Postgres 9.6 is not supported anymore https://www.postgresql.org/support/versioning/

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -124,7 +124,7 @@ services:
         hard: 10032
   postgres:
     <<: *restart_policy
-    image: "postgres:9.6"
+    image: "postgres:14.6"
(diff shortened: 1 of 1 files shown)


ethanhs:
Hi, I think @hubertdeng123 looked at this but we decided to put it on the backlog for now. It's definitely something we'd like to do soon, but we have some other priorities at the moment which are taking up our time. We also need to think through the upgrade path for existing installs.

aminvakil:
It's not simple as changing the version, let's continue discussion here: https://github.com/getsentry/self-hosted/issues/1610

Issue #1848, opened 2022-12-05
Add debug logging around bringing up clickhouse and snuba


ethanhs:
Clickhouse will sometimes get reported as unhealthy, potentially due to taking too long to come up. We should

1. [ ] Increase the timeout value for clickhouse
2. [ ] Log more information about the state of clickhouse if this error does happen

PR #1855, opened 2022-12-08
Set higher kafka helathcheck timeout

ethanhs:
Ref https://github.com/getsentry/self-hosted/issues/1848

This increases the timeout of kafka based on failures such as https://self-hosted.getsentry.net/organizations/self-hosted/issues/18/events/d9ae718413c94659b20b785ab0bb8aa8/?project=3&query=is%3Aunresolved&referrer=previous-event&sort=freq&statsPeriod=14d

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -190,6 +190,9 @@ services:
     healthcheck:
       <<: *healthcheck_defaults
       test: ["CMD-SHELL", "nc -z localhost 9092"]
+      interval: 3s
+      timeout: 600s
+      retries: 200
   clickhouse:
     <<: *restart_policy
     image: clickhouse-self-hosted-local


docker-compose.yml
```diff
@@ -190,6 +190,9 @@ services:
     healthcheck:
       <<: *healthcheck_defaults
       test: ["CMD-SHELL", "nc -z localhost 9092"]
+      interval: 3s
+      timeout: 600s
+      retries: 200
```
chadwhitacre, line 195: Is this the best way to structure this? From the [docs](https://docs.docker.com/engine/reference/builder/#healthcheck): 

> The health check will first run **interval** seconds after the container is started, and then again **interval** seconds after each previous check completes.
>
> If a single run of the check takes longer than **timeout** seconds then the check is considered to have failed.
>
> It takes **retries** consecutive failures of the health check for the container to be considered `unhealthy`.

The defaults are:

```
--interval=DURATION (default: 30s)
--timeout=DURATION (default: 30s)
--retries=N (default: 3)
```

This means currently the check is running every 30 seconds, each check is timing out after 30 seconds, and we're checking three times before giving up and getting an `unhealthy`. With this PR we would run the check every three seconds, each check out would timeout after 5 minutes, and we would try 200 times before giving up. This feels weird, 603 s × 200 = 120,600 s / 60 / 60 = 33.5 h. Is that what's intended here?

ethanhs: Good point!

I copied the settings from our [existing ones fro clickhouse](https://github.com/getsentry/self-hosted/blob/fcf0d3599b839535f19aa9bc96bc4b7af942a992/docker-compose.yml#L226), so I guess both should be updated.

ethanhs:
> not sure the semantics are as intended tho?

Fixed!

docker-compose.yml
chadwhitacre: Ah, good call. Yeah let's update both. I think the intention is more like:

```suggestion
      interval: 3s
      timeout: 3s
      retries: 100
```

(3 s + 3 s) * 100 = 600 s overall. What do you think?

chadwhitacre (approved):

docker-compose.yml
chadwhitacre: 10 / 10 / 30 also works. :o)

hubertdeng123: I like 10/10/30 a bit better!

hubertdeng123 (approved):

Issue #1847, opened 2022-12-05
Add more logging to image build script to detect failed services build


ethanhs:
We currently have seen a number of issues from [this line of build-docker-images.sh](https://github.com/getsentry/self-hosted/blob/144fd37153e6dad9f7aadf6f86931b4eced4a0fe/install/build-docker-images.sh#L7). To understand which images are failing better, we should split this line out into each service that needs to be built so we can identify which service(s) are failing to build.

PR #1858, opened 2022-12-08
Build each service image individually

ethanhs:
Fixes https://github.com/getsentry/self-hosted/issues/1847

I generated this list from things that have a `build` key, and those that use `x-sentry-defaults` (which builds an image as well).

### DIFF
--- a/install/build-docker-images.sh
+++ b/install/build-docker-images.sh
@@ -1,12 +1,10 @@
 echo "${_group}Building and tagging Docker images ..."
 
-
 echo ""
 # Build any service that provides the image sentry-self-hosted-local first,
 # as it is used as the base image for sentry-cleanup-self-hosted-local.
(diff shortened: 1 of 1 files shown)


install/build-docker-images.sh
```diff
@@ -1,10 +1,26 @@
...
+                     "worker"
+                     "ingest-consumer"
+                     "post-process-forwarder-errors"
+                     "post-process-forwarder-transactions"
+                     "subscription-consumer-events"
+                     "subscription-consumer-transactions"
+                     "clickhouse"
+                     "snuba-cleanup"
+                     "snuba-transactions-cleanup"
+                     "symbolicator-cleanup"
```
BYK, line 13: Why not `docker compose config --services`?

ethanhs: Because I didn't know that was a thing, neat! I'll change this to use that :)

hubertdeng123 (approved):

Issue #1835, opened 2022-11-28
Transaction stats not appearing in web UI


pavels:
### Self-Hosted Version

22.11.0

### CPU Architecture

x86_64

### Docker Version

containerd

### Docker Compose Version

Kubernetes - Helm chart

### Steps to Reproduce

Transaction stats stopped working for us. Most likely it some problem with out setup, but without any clue as where to look, it is hard to know.  As i understand, this is connected with outcomes consumer. It completely stopped working after 22.11.0 upgrade, the performance metrics and everything works just fine, so transactions are being processed just fine. Also, error stats is just fine. (The gaps in the past are due to problems with our nodestore implementation, that is solved now). It would really help to at least have some clues where to look, what to check.

### Expected Result

Stats should work.

### Actual Result

Stats for transactions don't show up, no error from any of the snuba services are logged. Outcomes consumer consumes messages fine according to it's log

![image](https://user-images.githubusercontent.com/244128/204309979-14494f08-281f-4ecc-a3a3-31d2d8c5336d.png)

![image](https://user-images.githubusercontent.com/244128/204309888-b99c9368-1335-4dd1-a868-72c0668b7b97.png)

(outcomes consumer)
```
2022-11-28 15:00:42,711 Completed processing <Batch: 48 messages, open for 2.49 seconds>.
2022-11-28 15:00:44,812 Completed processing <Batch: 72 messages, open for 2.07 seconds>.
2022-11-28 15:00:47,476 Completed processing <Batch: 82 messages, open for 2.73 seconds>.
2022-11-28 15:00:50,319 Completed processing <Batch: 78 messages, open for 2.86 seconds>.
2022-11-28 15:00:53,215 Completed processing <Batch: 94 messages, open for 2.93 seconds>.
2022-11-28 15:00:55,275 Completed processing <Batch: 111 messages, open for 2.06 seconds>.
2022-11-28 15:00:57,849 Completed processing <Batch: 61 messages, open for 2.63 seconds>.
2022-11-28 15:00:59,886 Completed processing <Batch: 119 messages, open for 2.11 seconds>.
2022-11-28 15:01:02,274 Completed processing <Batch: 50 messages, open for 2.48 seconds>.
2022-11-28 15:01:04,808 Completed processing <Batch: 57 messages, open for 2.56 seconds>.
2022-11-28 15:01:06,894 Completed processing <Batch: 103 messages, open for 2.08 seconds>.
[...]

PR #1862, opened 2022-12-12
Enable dynamic sampling

ethanhs:
I'm unsure if we should merge this so close to a new release, but it seems to be relatively low-risk (famous last words haha).

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -240,6 +240,8 @@ SENTRY_OPTIONS["mail.from"] = f"sentry@{SENTRY_OPTIONS['mail.list-namespace']}"
 # Features #
 ############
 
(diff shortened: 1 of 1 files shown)


chadwhitacre:
Have we had any conversation with DS stakeholders about this? Did they think through self-hosted when they built this? What are the implications?

hubertdeng123:
is there a reason why this isn't turned on by default in `getsentry/sentry` in `src/sentry/conf/server.py`?

ethanhs:
I will check with DS people and get answers to both of those.

ethanhs:
I think this may be more complicated, c.f. [this user who enabled it in kubernetes](https://github.com/getsentry/self-hosted/issues/1835#issuecomment-1349928636). I'll close this for now and open an issue to track this.

ethanhs:
Filed https://github.com/getsentry/self-hosted/issues/1866

PR #1881, opened 2022-12-22
ref: Retry wal2json download in installer

ethanhs:
This has been causing a lot of failures in CI in Sentry and elsewhere recently. Hopefully the failures are flakes and this will work around them.

### DIFF
--- a/install/install-wal2json.sh
+++ b/install/install-wal2json.sh
@@ -11,12 +11,16 @@ docker_curl() {
 }
 
 if [[ $WAL2JSON_VERSION == "latest" ]]; then
-  VERSION=$(
-    docker_curl https://api.github.com/repos/getsentry/wal2json/releases/latest |
-      grep '"tag_name":' |
-      sed -E 's/.*"([^"]+)".*/\1/'
-  )
-
+  # Retry up to 3 times, with 10 seconds between each attempt.
+  for i in {0..30..10}; do
+    sleep $i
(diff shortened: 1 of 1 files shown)


install/install-wal2json.sh
```diff
@@ -11,12 +11,16 @@ docker_curl() {
...
 
 if [[ $WAL2JSON_VERSION == "latest" ]]; then
-  VERSION=$(
-    docker_curl https://api.github.com/repos/getsentry/wal2json/releases/latest |
-      grep '"tag_name":' |
-      sed -E 's/.*"([^"]+)".*/\1/'
-  )
-
+  # Retry up to 3 times, with 10 seconds between each attempt.
+  for i in {0..30..10}; do
```
BYK, line 15: So it's just the API look up that flakes? (wondering if there should be a retry mechanism below where we actually download the file)

BYK: Also, looks like `curl` has built-in retry mechanisms: https://stackoverflow.com/a/42873372

Might be worthwhile to bake this into `docker_curl` for all operations?

BYK (approved):

hubertdeng123 (approved):

install/install-wal2json.sh
ethanhs: Ah good point! I'll switch to doing that.

chadwhitacre (approved):
Seems reasonable. 👍

Issue #1762, opened 2022-10-17
Check health of containers in e2e tests


ethanhs:
We want to check that all the containers are healthy after we run the event through the test instance - this catches any containers that crash due to an issue at runtime. 

Currently we only check container health in the installer.

BYK:
Did this actually happen causing a bug getting through? If so I'm willing to submit a PR :)

ethanhs:
Yes, just before the 22.10.0 release, the error and transaction pipelines were split, causing https://github.com/getsentry/sentry/issues/40120

If you would like to submit a PR to add this that would be great, but I think if we are going to centralize our e2e tests (#1756 for anyone reading along) we should probably hold off until that is done first.

PR #1897, opened 2023-01-06
ci: Check health of services after running integration tests

ethanhs:
Fixes https://github.com/getsentry/self-hosted/issues/1762

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -136,18 +136,10 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
 
-service_failures=false
-SERVICES_NOT_RUNNING=$(docker compose ps --format json | jq -r '.[]| select(.State != "running")')
-for service in $(echo "$SERVICES_NOT_RUNNING" | jq -r '.Service'); do
-  # geoipupdate runs once at startup, and should always exit.
-  if [[ $service != "geoipupdate" ]]; then
-    echo "ERROR: Service $service has failed after the integration tests!"
-    service_failures=true
-  fi
-done
+docker compose ps --format json | jq -r \
+  '.[] | \
+  # we only care about running services, geoipupdate always exits, so we ignore it
+  select(.State != "running" and .Service != "geoipupdate") | \
+  # Filter to only show the service name and state
+  with_entries(select(.key | in({"Service":1, "State":1})))'
 
-if [[ $service_failures ]]; then
-  echo "$SERVICES_NOT_RUNNING"
-  echo "One or more services failed, exiting..."
-  exit 1
-fi
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -135,3 +135,10 @@ source ./custom-ca-roots/setup.sh
 $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
+
+for service in $(docker compose ps --format json | jq -r '.[]| select(.State != "running") | .Service'); do
+  # geoipupdate is not set up, so we can safely ignore it, it will fail if not configured.
+  if [[ $service != "geoipupdate" ]]; then
+    echo "ERROR: Service $service has failed after the integration tests!"
+  fi
+done
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -136,12 +136,17 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
 
+# Table formatting based on https://stackoverflow.com/a/39144364
 COMPOSE_PS_OUTPUT=$(docker compose ps --format json | jq -r \
-  '.[] |
-  # we only care about running services. geoipupdate always exits, so we ignore it
-  select(.State != "running" and .Service != "geoipupdate") |
-  # Filter to only show the service name and state
-  with_entries(select(.key | in({"Service":1, "State":1})))')
+  '["Service","State"], ["--","------"], (
+      .[] |
+      # we only care about running services. geoipupdate always exits, so we ignore it
+      select(.State != "running" and .Service != "geoipupdate") |
+      # Filter to only show the service name and state
+      with_entries(select(.key | in({"Service":1, "State":1})))
+    ) |
+    @tsv
+ ')
 
 if [[ "$COMPOSE_PS_OUTPUT" ]]; then
   echo "Services failed, oh no!"
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -136,7 +136,7 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
 
-COMPOSE_PS_OUTPUT=$(docker compose ps --format json | jq -r
+COMPOSE_PS_OUTPUT=$(docker compose ps --format json | jq -r \
   '.[] |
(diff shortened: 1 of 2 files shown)


_integration-test/run.sh
```diff
@@ -135,3 +135,10 @@ source ./custom-ca-roots/setup.sh
 $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
+
+for service in $(docker compose ps --format json | jq -r '.[]| select(.State != "running") | .Service'); do
+  # geoipupdate is not set up, so we can safely ignore it, it will fail if not configured.
```
chadwhitacre, line 140: The `geoipupdate` container is designed to always exit, it runs once on startup.

_integration-test/run.sh
```diff
@@ -135,3 +135,10 @@ source ./custom-ca-roots/setup.sh
 $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source ./custom-ca-roots/teardown.sh
 echo "${_endgroup}"
+
+for service in $(docker compose ps --format json | jq -r '.[]| select(.State != "running") | .Service'); do
```
chadwhitacre, line 139: We're guaranteed to have `jq` installed?

_integration-test/run.sh
```diff
@@ -135,3 +135,10 @@ source ./custom-ca-roots/setup.sh
...
+  # geoipupdate is not set up, so we can safely ignore it, it will fail if not configured.
+  if [[ $service != "geoipupdate" ]]; then
+    echo "ERROR: Service $service has failed after the integration tests!"
```
chadwhitacre, line 142: I'd recommend outputting the full `docker compose ps` so we have better debugging helps, maybe even logs of the failing service.

Also do we want to exit with a non-zero status so that CI goes red?

chadwhitacre (changes requested):
Some suggested improvements.

_integration-test/run.sh
ethanhs: Yes, the unit tests require it already

_integration-test/run.sh
ethanhs: Done! I figured the logs may be too verbose if multiple services are failing.

hubertdeng123 (approved):

chadwhitacre:
Whoa! Is it ... working? 🤔

```
self-hosted-1-snuba-replacer-1                             "./docker_entrypoint…"   snuba-replacer                             restarting          
```

https://github.com/getsentry/self-hosted/actions/runs/3877996309/jobs/6613667226#step:4:6875

Is the replacer _supposed_ to restart?

chadwhitacre:
Seeing `snuba-replacer` restarting [again](https://github.com/getsentry/self-hosted/actions/runs/3877996309/jobs/6628177660#step:4:6890) but [not consistently](https://github.com/getsentry/self-hosted/actions/runs/3877996309/jobs/6628177231#step:4:6702). What fails on that second? 🤔

ethanhs:
@chadwhitacre 

> Seeing snuba-replacer restarting [again](https://github.com/getsentry/self-hosted/actions/runs/3877996309/jobs/6628177660#step:4:6890) but [not consistently](https://github.com/getsentry/self-hosted/actions/runs/3877996309/jobs/6628177231#step:4:6702). What fails on that second? 🤔

My guess is that it restarted but switched to a running state after we queried docker ps, so the second docker ps (the one we print) doesn't show the issue.

BYK:
Not sure of this helps but Compose V2 has a `--wait` option that waits until all the containers are healthy:  https://docs.docker.com/engine/reference/commandline/compose_up/

chadwhitacre:
> My guess is that it restarted but switched to a running state after we queried docker ps, so the second docker ps (the one we print) doesn't show the issue.

Seems plausible. Should we capture the output of the first call so we can output exactly what we're testing against? Related: I did try to find a `$dc ps --format` that would reduce the columns of output for easier reading (we only really care about name and state/status), but afaict `docker-compose ps --format` does not support templates the way `docker ps` does (of course). Maybe dump that json and work with it twice?

ethanhs:
> Maybe dump that json and work with it twice?

Yeah I'm leaning towards that right now as well.

ethanhs:
Of course, I'm still not sure why snuba-replacer is restarting...

For using `--wait`, I think since geoipupdate always exit we can't use that?

ethanhs:
OK, now printing the JSON of the failing/exited services.

_integration-test/run.sh
```diff
@@ -146,7 +147,7 @@ for service in $(docker compose ps --format json | jq -r '.[]| select(.State !=
 done
 
 if [[ $service_failures ]]; then
-  $dc ps
+  echo "$SERVICES_NOT_RUNNING"
```
chadwhitacre, line 150: Maybe a little more jq on this bad boy? We really on care about name and status, also don't care about geoipupdate.

chadwhitacre:
> I'm still not sure why snuba-replacer is restarting

Aye, there's the rub!

ethanhs:
Found it! [Batching was removed from replacer](https://github.com/getsentry/snuba/commit/6e9807459631efdd737bd918fe85907ec257948a), added a fix for that.

_integration-test/run.sh
ethanhs: Fair point! Will do.

chadwhitacre:
So does that mean self-hosted has been broken for three weeks? We should add some tests to catch this sort of thing. Oh wait ... ;-)

chadwhitacre:
I pushed 668ae2a1f867509556eee72f097e96dd6ec1cfce in order to evaluate the failure output formatting.

chadwhitacre:
Here's what I'm [seeing](https://github.com/getsentry/self-hosted/actions/runs/3894165549/jobs/6647771080#step:4:6727):

```
{
  "Service": "snuba-replacer",
  "State": "restarting"
}
```

How about using the `@tsv` filter?

https://stackoverflow.com/questions/39139107/how-to-format-a-json-string-as-a-table-using-jq

chadwhitacre:
(I force-pushed 668ae2a1f867509556eee72f097e96dd6ec1cfce off again.)

chadwhitacre (changes requested):
Could we use the `@tsv` filter for better readability?

chadwhitacre:
Heh, oops! With `@tsv` there is always output (at least the headers) so it always fails.

https://github.com/getsentry/self-hosted/actions/runs/3895765109/jobs/6651476439#step:4:6707

chadwhitacre:
Sob and actually broken run won't blend.

```
jq: error (at <stdin>:1): object ({"Service":...) cannot be tsv-formatted, only array
```

https://github.com/getsentry/self-hosted/actions/runs/3895778334/jobs/6651506996#step:4:2661

chadwhitacre:
Can you push a commit to show the error case?

chadwhitacre:
b9f949cb69b41a027462da1559a1c9dda22cab5c is expected to pass CI (right?).

ethanhs:
> [b9f949c](https://github.com/getsentry/self-hosted/commit/b9f949cb69b41a027462da1559a1c9dda22cab5c) is expected to pass CI (right?).

Yes

ethanhs:
I will push something that breaks but I wanted to make sure the expected success case works first :)

ethanhs:
[64d51a2](https://github.com/getsentry/self-hosted/pull/1897/commits/64d51a27d902154b98019ba7fe6e101c5ba25748) adds back the snuba breakage to test things

ethanhs:
Sample run with TSV: https://github.com/getsentry/self-hosted/actions/runs/3896522967/jobs/6653150931

hubertdeng123:
Can you add this [comment](https://github.com/getsentry/self-hosted/pull/1897#issuecomment-1378016913) to the PR description? For future reference so it's easier to see why it was deleted. Otherwise lgtm

chadwhitacre (approved):

Issue #1899, opened 2023-01-06
Wal2JSON version download is flaking


ethanhs:
Wal2JSON keeps failing in CI. I initially thought this was due to Github's API being flaky, but it seems that it is something to do with the reading end of the pipes hanging up early. We should replace the pipes into grep and sed with a call to `jq` now that we use that in error handling.

Note I've only seen this in Github Actions... if someone sees this outside of that, please do comment below with your OS!

chadwhitacre:
Boom.

```
{"message":"API rate limit exceeded for 13.64.10.224. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

https://github.com/getsentry/self-hosted/actions/runs/3877670089/jobs/6612957076#step:4:957

chadwhitacre:
We haven't cut a new wal2json release in almost two years. Let's hard-code it.

PR #1906, opened 2023-01-09
Fix wal2json flakiness

chadwhitacre:
Kludge to fix https://github.com/getsentry/self-hosted/issues/1899.

### DIFF
--- a/install/install-wal2json.sh
+++ b/install/install-wal2json.sh
@@ -17,8 +17,16 @@ docker_curl() {
 
 if [[ $WAL2JSON_VERSION == "latest" ]]; then
   # Hard-code this. Super-hacky. We were curling the GitHub API here but
-  # hitting rate limits in CI. This library hasn't seen a new release for
-  # a year and a half at time of writing.
+  # hitting rate limits in CI. This library hasn't seen a new release for a
+  # year and a half at time of writing.
+  #
+  # If you're reading this do us a favor and go check:
+  #
+  #   https://github.com/getsentry/wal2json/releases
+  #
+  # If there's a new release can you update this please? If not maybe subscribe
+  # for notifications on the repo with "Watch > Custom > Releases". Together we
+  # can make a difference.
   VERSION=0.0.2
 else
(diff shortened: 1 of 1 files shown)


install/install-wal2json.sh
```diff
@@ -16,11 +16,11 @@ docker_curl() {
 }
 
 if [[ $WAL2JSON_VERSION == "latest" ]]; then
-  VERSION=$(
-    docker_curl https://api.github.com/repos/getsentry/wal2json/releases/latest |
-      grep '"tag_name":' |
-      sed -E 's/.*"([^"]+)".*/\1/'
-  )
+  docker_curl https://api.github.com/repos/getsentry/wal2json/releases/latest >wal2json.json
+  cat "====================================================="
```
hubertdeng123, line 20: seems wrong here?

hubertdeng123: also failing in e2e-tests

chadwhitacre: 😬

hubertdeng123 (approved):

hubertdeng123:
oh wow, were you able to find the issue on a build failure on the PR? :O

chadwhitacre:
Yeah, posted [up here](https://github.com/getsentry/self-hosted/issues/1899#issuecomment-1376305584).

chadwhitacre:
Checked with SNS.

>> How likely is it that [wal2json will ever see another release](https://github.com/getsentry/wal2json/releases)?
>> Trying to decide how to handle updates in self-hosted (current approach is leading to API rate limits in CI, wondering if we can cheat).
>
> Our fork or the origin one ?
> Our fork is unlikely.
> The origin does.
> Now the question is whether we will keep doing change data capture at all.
> The two use cases are
> - Data pipelines. No self hosted. The data team wants to get rid of it, but the solution currently investigated does not work. So it is an open question.
> - Issues platform. The implementation of issue search based on CDC data was never rolled out and the team was repurposed, though there are possible ways issue platform may need it again. So this one is an open question as well

PR #1907, opened 2023-01-10
[WIP] Fix snuba replacer

ethanhs:
This is mostly used for testing


ethanhs:
```
self-hosted-1-snuba-replacer-1  | Error: No such option: --max-batch-size
```

Well that would do it!

Issue #1888, opened 2022-12-28
Add test for symbolification pipeline, based on test in Sentry repo


ethanhs:
Arpad suggested for testing symbolicator workflows:

>    That seems like a good idea, yes. To cover this particular usecase, you would need a test that uses all the builtin symbol servers. A Windows minidump perhaps. We should have one already in the main Sentry tests, though those tests are always testing the very latest symbolicator master.

The tests live here: https://github.com/getsentry/sentry/tree/master/tests/symbolicator

PR #1916, opened 2023-01-17
Add test for symbolicator pipeline

ethanhs:
Fixes https://github.com/getsentry/self-hosted/issues/1888

This integration test:

uploads a symbol file, which indicates where to find the symbols for the minidump
uploads a sample minidump, which creates an event we can check for

This test is based on the test here: https://github.com/getsentry/sentry/blob/28f4f36b4adc3d39ea80f2f1864cff8e311b6af3/tests/symbolicator/test_minidump_full.py#L72


.github/workflows/test.yml
```diff
@@ -68,6 +68,8 @@ jobs:
           sudo mkdir -p "${{ matrix.compose_path }}"
           sudo curl -L https://github.com/docker/compose/releases/download/${{ matrix.compose_version }}/docker-compose-`uname -s`-`uname -m` -o "${{ matrix.compose_path }}/docker-compose"
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
+      - name: Install sentry-cli
```
ethanhs, line 71: This adds a dependency on sentry-cli, I expect that is ok to add as a CI-only dependency, will add it to the e2e tests action if we go ahead with that. I couldn't use the docker container because on macos the network runs in a VM which is a different network than the host :/

_integration-test/run.sh
```diff
@@ -51,7 +53,7 @@ echo "${_endgroup}"
 
 echo "${_group}Running tests ..."
 get_csrf_token() { awk '$6 == "sc" { print $7 }' $COOKIE_FILE; }
-sentry_api_request() { curl -s -H 'Accept: application/json; charset=utf-8' -H "Referer: $SENTRY_TEST_HOST" -H 'Content-Type: application/json' -H "X-CSRFToken: $(get_csrf_token)" -b "$COOKIE_FILE" -c "$COOKIE_FILE" "$SENTRY_TEST_HOST/api/0/$1" ${@:2}; }
+sentry_api_request() { curl -s -H 'Accept: application/json; charset=utf-8' -H "Referer: $SENTRY_TEST_HOST" -H 'Content-Type: application/json' -H "X-CSRFToken: $(get_csrf_token)" -b "$COOKIE_FILE" -c "$COOKIE_FILE" "$SENTRY_TEST_HOST/$1" ${@:2}; }
```
ethanhs, line 54: We make a request against relay so we must make this generic over the API path used (see where we calculate `EVENT_ID` below)

_integration-test/run.sh
```diff
@@ -130,6 +132,42 @@ cat debug.log
...
+SCOPES=$(jq -n -c --argjson scopes '["event:admin", "event:read", "member:read", "org:read", "team:read", "project:read", "project:write", "team:write"]' '$ARGS.named')
+SENTRY_AUTH_TOKEN=$(sentry_api_request "api/0/api-tokens/" -X POST --data "$SCOPES" | jq -r '.token')
+SENTRY_DSN=$(sentry_api_request "api/0/projects/sentry/native/keys/" | jq -r '.[0].dsn.secret')
+# Then upload the symbols to that project (note the container mounts pwd to /work)
+SENTRY_URL="$SENTRY_TEST_HOST" sentry-cli upload-dif --org "$SENTRY_ORG" --project "$SENTRY_PROJECT" --auth-token "$SENTRY_AUTH_TOKEN" windows.sym
+# Get public key for minidump upload
+PUBLIC_KEY=$(sentry_api_request "api/0/projects/sentry/native/keys/" | jq -r '.[0].public')
+# Upload the minidump to be processed, this returns the event ID of the crash dump
+EVENT_ID=$(sentry_api_request "api/$NATIVE_PROJECT_ID/minidump/?sentry_key=$PUBLIC_KEY" -X POST -F 'upload_file_minidump=@windows.dmp' | sed 's/\-//g')
+# We have to wait for the item to be processed
```
ethanhs, line 153: Unfortunately, self-hosted runs symbolicator with caches, and those can mean things aren't processed right away. I'm not aware of a way to flush the caches, the tests in Sentry say to run without them(!)

chadwhitacre: > I'm not aware of a way to flush the caches

Did you ask in #discuss-ingest?

> the tests in Sentry say to run without them

Is that what we're doing here?

_integration-test/run.sh
```diff
@@ -130,6 +130,42 @@ cat debug.log
 echo '------------------------------------------'
 echo "${_endgroup}"
 
+echo "${_group}Test symbolicator works ..."
+SENTRY_ORG="${SENTRY_ORG:-sentry}"
+SENTRY_PROJECT="${SENTRY_PROJECT:-native}"
+SENTRY_TEAM="${SENTRY_TEAM:-sentry}"
+# First set up a new project if it doesn't exist already
+PROJECT_JSON=$(jq -n -c --arg name "$SENTRY_PROJECT" --arg slug "$SENTRY_PROJECT" '$ARGS.named')
```
chadwhitacre, line 138: You're really getting familiar with `jq`, aren't you? :D

_integration-test/run.sh
```diff
@@ -130,6 +130,42 @@ cat debug.log
...
+NATIVE_PROJECT_ID=$(sentry_api_request "api/0/teams/$SENTRY_ORG/$SENTRY_TEAM/projects/" | jq -r '.[]|select(.slug == "'"$SENTRY_PROJECT"'")|.id')
```
hubertdeng123, line 139: nit: `"'"$SENTRY_PROJECT"'" -> "'$SENTRY_PROJECT'"`

_integration-test/run.sh
```diff
@@ -130,6 +130,42 @@ cat debug.log
 echo '------------------------------------------'
 echo "${_endgroup}"
 
+echo "${_group}Test symbolicator works ..."
+SENTRY_ORG="${SENTRY_ORG:-sentry}"
```
hubertdeng123, line 134: Under what circumstance are `SENTRY_ORG`, `SENTRY_PROJECT`, or `SENTRY_TEAM` not declared for integration tests?

_integration-test/run.sh
ethanhs: No, we are running with caches enabled, since that is more true to how self-hosted runs. I can ask in #discuss-ingest but since the Sentry tests don't flush the cache, and there is no documentation about how to do so, I presume it isn't possible.

_integration-test/run.sh
ethanhs: That wouldn't escape the variable. Because its the same as `'string' + $var + 'end'`, whereas we want `'string' + "$var" + 'end'`. Not really an issue since this is only a single word, but for consistency's sake I try to always quote variables.

_integration-test/run.sh
ethanhs: They aren't by default, but I am mimicking the line above where `SENTRY_TEST_HOST` gets set.

_integration-test/run.sh
```diff
@@ -130,6 +130,42 @@ cat debug.log
...
+  EVENT_PROCESSED=$(sentry_api_request "api/0/projects/$SENTRY_ORG/$SENTRY_PROJECT/events/" | jq -r '.[]|select(.id == "'"$EVENT_ID"'")|.id')
+  if [ -z "$EVENT_PROCESSED" ]; then
+    sleep $i
+  else
+    break
+  fi
+done
+if [ -z "$EVENT_PROCESSED" ]; then
+  echo "Hm, the event $EVENT_ID didn't exist... listing events that do exist:"
+  sentry_api_request "api/0/projects/$SENTRY_ORG/$SENTRY_PROJECT/events/" | jq .
```
chadwhitacre, line 164: Yay for some debugging output.

_integration-test/run.sh
```diff
@@ -130,6 +130,42 @@ cat debug.log
...
+PUBLIC_KEY=$(sentry_api_request "api/0/projects/sentry/native/keys/" | jq -r '.[0].public')
+# Upload the minidump to be processed, this returns the event ID of the crash dump
+EVENT_ID=$(sentry_api_request "api/$NATIVE_PROJECT_ID/minidump/?sentry_key=$PUBLIC_KEY" -X POST -F 'upload_file_minidump=@windows.dmp' | sed 's/\-//g')
+# We have to wait for the item to be processed
+for i in {0..60..10}; do
+  EVENT_PROCESSED=$(sentry_api_request "api/0/projects/$SENTRY_ORG/$SENTRY_PROJECT/events/" | jq -r '.[]|select(.id == "'"$EVENT_ID"'")|.id')
+  if [ -z "$EVENT_PROCESSED" ]; then
+    sleep $i
+  else
+    break
```
chadwhitacre, line 159: So the success case is having our event processed at all? Is it worth implementing the [second half of the test](https://github.com/getsentry/sentry/blob/28f4f36b4adc3d39ea80f2f1864cff8e311b6af3/tests/symbolicator/test_minidump_full.py#L86-L103) to check the integrity of the file?

chadwhitacre: Talked on a call ... I'm good with the limited smoke test here.

chadwhitacre:
[Passing case](https://github.com/getsentry/self-hosted/actions/runs/3944220970/jobs/6749899133#step:5:6619). 💃 

Let's see it catch a failure! 😁

_integration-test/run.sh
ethanhs: The cache should not be persistent. It seems that there is no way to manually trigger a flush :/

BYK:
Observation: I don't recall why I opted for using `bash` for these tests (probably for something quick and dirty) but at this point, it feels to me that the tests should be a bit more organized and manageable? Maybe using something like Playwright?

ethanhs:
@BYK eventually I'd like to rewrite the installer in Python and make it so we can use pytest and all the creature comforts that come from that (probably use playwright in tests as well!). But that is a much larger project 😅

ethanhs:
Here is a sample of a failure: https://github.com/getsentry/self-hosted/actions/runs/3952966281/jobs/6768689917#step:5:6808

Symbol upload will fail of course, so we fail at the sentry-cli point.

BYK:
>@BYK eventually I'd like to rewrite the installer in Python and make it so we can use pytest and all the creature comforts that come from that (probably use playwright in tests as well!). But that is a much larger project 😅

This is kind of orthogonal to what I'm suggesting though. There's no reason for integration tests to be in the same language or environment with the installer (maybe it was my faulty brain which fell into this misconseption in the first place, don't really remember). We can right now at this moment, start creating Playwright tests and start benefitting from them which would cut through most of this crazy bash/curl/jq dance with CSRF tokens etc. It would:

- Cover more surface area
- Be easier to understand and maintain
- Have videos of the failures if we wanted to
- Have screenshot tests if we wanted to

Regarding rewriting the installer in Python (or any other language tbh), I'd suggest finding ways of getting rid of it completely (such as by adopting Ansible, Helm charts, or whatever). It would be a challenge to use Python unless the installer gets compiled into a self-contained binary for each system by using something like [PyOxidizer](https://pyoxidizer.readthedocs.io/en/stable/)

chadwhitacre:
Also [benefiting from](https://github.com/getsentry/self-hosted/actions/runs/3952966281/jobs/6768689917#step:6:45) https://github.com/getsentry/self-hosted/issues/1762: 😁

```
self-hosted-1-symbolicator-1                               "/bin/bash /docker-e…"   symbolicator                               exited (137)        
```

chadwhitacre:
I'm open to Playwright, def separate from installer lang.

ethanhs:
> We can right now at this moment, start creating Playwright tests and start benefitting from them which would cut through most of this crazy bash/curl/jq dance with CSRF tokens etc.

This is true. I am hesitant to start on this since we are near the end of the quarter and I don't want the project delayed much more, but may when I add the snuba tests I will add them as playwright tests.

> Regarding rewriting the installer in Python (or any other language tbh), I'd suggest finding ways of getting rid of it completely (such as by adopting Ansible, Helm charts, or whatever). It would be a challenge to use Python unless the installer gets compiled into a self-contained binary for each system by using something like [PyOxidizer](https://pyoxidizer.readthedocs.io/en/stable/)

The idea with that was to make the installer run entirely in docker except for a minimal kickoff script. I agree Ansible might be better, but I'm unsure if making that a dependency is acceptable for our users (right now we only require GNU core utils and docker).

hubertdeng123 (approved):

chadwhitacre (approved):

PR #1937, opened 2023-01-26
Updates Redis to 6.2.10

danielhartnell:
Redis patched two CVEs in release `6.2.9`. We would like to update the Sentry container to apply those fixes:

https://github.com/redis/redis/releases/tag/6.2.9

After that release, it looks like they pushed a bug fix in `6.2.10` so it probably makes sense to bump up to that.

If there's anything I can do to help test or evaluate this change, let me know.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -112,7 +112,7 @@ services:
       test: echo stats | nc 127.0.0.1 11211
   redis:
     <<: *restart_policy
-    image: "redis:6.2.4-alpine"
+    image: "redis:6.2.10-alpine"
     healthcheck:
       <<: *healthcheck_defaults
       test: redis-cli ping


BYK (approved):

chadwhitacre (approved):
LGTM but what version are we on in SaaS prod for the record @hubertdeng123?

hubertdeng123 (approved):

chadwhitacre:
TIL image-mirror. Are we sure that's reflective of what's in SaaS?

hubertdeng123:
ok, let me confirm to be sure. Looks like redis 5.0 image is used for dev and I saw it being pulled in for getsentry tests

hubertdeng123:
Looks like depending on the cluster, we're running either 5.x of 6.x

chadwhitacre:
Sounds good. Ship it! 🚢

chadwhitacre:
Thanks @danielhartnell! :-)

PR #1938, opened 2023-01-26
Fix post-release.sh for `git pull`

BYK:
Adds a workaround for the error in https://github.com/getsentry/publish/actions/runs/4018136005

(not sure if this is fool-proof but it follows the suggestion in the output)

### Legal Boilerplate


BYK:
Well, this would not fix the issue but would remove the annoying message.

chadwhitacre (approved):

hubertdeng123 (approved):

PR #1940, opened 2023-01-27
ci: Add test for alerts to integration tests

### LINKED ISSUE #1887: Add test for alerts/notifications to test the snuba pipeline
This is part of https://github.com/getsentry/self-hosted/issues/1761, should hopefully better test snuba.

ethanhs:
Alright, so I'm kicking myself a bit here... turns out it was way simpler to just use curl and jq rather than use playwright 🤦‍♂️😭.

I'm not totally sad I tried out playwright, as it gave good information on what we'd need for future perhaps more complicated tests (i.e. greater integration with frontend).

I initially was stuck on this route because I kept getting access denied issues, but it turns out if you keep your cookie around (like we do in the integration tests!) it doesn't block access. I'm not totally happy we're using an undocumented API (I plan to seek out if there is an ideal path for either making this API documented/stable in some way, or if we should be doing something else entirely), but I think this should be fine.

I decided to break this up into two parts: set the alert before other tests which create an event, then check that
a) the event has fired according to Sentry
b) the smtp server logs indicate we tried to look up the email of our test user (aka we tried to email them)

Should we change the test user email to some invalid domain? 🤔 

Fixes #1887

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -154,7 +154,7 @@ SENTRY_AUTH_TOKEN=$(sentry_api_request "api/0/api-tokens/" -X POST --data "$SCOP
 sleep 10
 SENTRY_DSN=$(sentry_api_request "api/0/projects/sentry/native/keys/" | jq -r '.[0].dsn.secret')
 # Then upload the symbols to that project (note the container mounts pwd to /work)
-SENTRY_URL="$SENTRY_TEST_HOST" sentry-cli upload-dif --org "$SENTRY_ORG" --project "$SENTRY_PROJECT" --auth-token "$SENTRY_AUTH_TOKEN" windows.sym --log-level=debug
+SENTRY_URL="$SENTRY_TEST_HOST" sentry-cli upload-dif --org "$SENTRY_ORG" --project "$SENTRY_PROJECT" --auth-token "$SENTRY_AUTH_TOKEN" windows.sym --log-level=debug --wait
 # Get public key for minidump upload
 PUBLIC_KEY=$(sentry_api_request "api/0/projects/sentry/native/keys/" | jq -r '.[0].public')
 # Upload the minidump to be processed, this returns the event ID of the crash dump
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -152,8 +152,6 @@ SCOPES=$(jq -n -c --argjson scopes '["event:admin", "event:read", "member:read",
 SENTRY_AUTH_TOKEN=$(sentry_api_request "api/0/api-tokens/" -X POST --data "$SCOPES" | jq -r '.token')
 # wait to make sure the new token and keys have been generated
 sleep 10
-$dc stop post-process-forwarder-errors
-$dc stop post-process-forwarder-transactions
 SENTRY_DSN=$(sentry_api_request "api/0/projects/sentry/native/keys/" | jq -r '.[0].dsn.secret')
 # Then upload the symbols to that project (note the container mounts pwd to /work)
 SENTRY_URL="$SENTRY_TEST_HOST" sentry-cli upload-dif --org "$SENTRY_ORG" --project "$SENTRY_PROJECT" --auth-token "$SENTRY_AUTH_TOKEN" windows.sym --log-level=debug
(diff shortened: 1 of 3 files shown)


chadwhitacre:
Can you link to failing and succeeding test runs?

chadwhitacre:
Also did you have a pattern test for this in `sentry` like you did for the symbolicator test? Would be good to compare ...

_integration-test/run.sh
```diff
@@ -105,7 +105,7 @@ EVENT_PATH="api/0/projects/sentry/internal/events/$TEST_EVENT_ID/"
 export -f sentry_api_request get_csrf_token
 export SENTRY_TEST_HOST COOKIE_FILE EVENT_PATH
 printf "Getting the test event back"
-timeout 60 bash -c 'until $(sentry_api_request "$EVENT_PATH" -Isf -X GET -o /dev/null); do printf '.'; sleep 0.5; done'
+timeout 90 bash -c 'until $(sentry_api_request "$EVENT_PATH" -Isf -X GET -o /dev/null); do printf '.'; sleep 0.5; done'
```
hubertdeng123, line 108: is this to fix flakiness?

_integration-test/run.sh
```diff
@@ -130,16 +130,23 @@ cat debug.log
...
+SENTRY_PROJECT="${SENTRY_PROJECT:-native}"
 PROJECT_JSON=$(jq -n -c --arg name "$SENTRY_PROJECT" --arg slug "$SENTRY_PROJECT" '$ARGS.named')
 NATIVE_PROJECT_ID=$(sentry_api_request "api/0/teams/$SENTRY_ORG/$SENTRY_TEAM/projects/" | jq -r '.[]|select(.slug == "'"$SENTRY_PROJECT"'")|.id')
 if [ -z "${NATIVE_PROJECT_ID}" ]; then
   NATIVE_PROJECT_ID=$(sentry_api_request "api/0/teams/$SENTRY_ORG/$SENTRY_TEAM/projects/" -X POST --data "$PROJECT_JSON" | jq -r '. // null | .id')
+  if [ -z "${NATIVE_PROJECT_ID}" ]; then
+    sentry_api_request "api/0/teams/$SENTRY_ORG/$SENTRY_TEAM/projects/" -X POST --data "$PROJECT_JSON" | jq
+  fi
 fi
+sentry_api_request "api/0/projects/$SENTRY_TEAM/$SENTRY_PROJECT/rules/?duplicateRule=false&wizardV3=true" -X POST -d @./alert_settings/settings.json | jq '.actions'
```
hubertdeng123, line 146: what is `wizardV3`? 🤔

_integration-test/run.sh
ethanhs: Yes

_integration-test/run.sh
ethanhs: This API call is usually done from the frontend, so I am just copying what the frontend sends request-wise.

ethanhs:
> Also did you have a pattern test for this in sentry like you did for the symbolicator test? Would be good to compare ...

No, this just makes an alert and checks that notifications went out. Lynn said if alerting works then post processing should work as well.

Here's a failing run: https://github.com/getsentry/self-hosted/actions/runs/4027673632/jobs/6923674649
Here's a successful run: https://github.com/getsentry/self-hosted/actions/runs/4028467450/jobs/6925385304

hubertdeng123 (approved):

chadwhitacre:
Can you link the commits for the failing and successful tests? Hard to get at from the action runs.

ethanhs:
@chadwhitacre 
Here is the successful commit: https://github.com/getsentry/self-hosted/pull/1940/commits/cdf0d0660e9d9d12e7bf9ff05226503aa65a9c0e
Failed commit: https://github.com/getsentry/self-hosted/pull/1940/commits/d62929451d49d8ddfa72fbbcb2633759233e25bc

chadwhitacre:
Why does d62929451d49d8ddfa72fbbcb2633759233e25bc fail? I would expect the failing commit to modify a fixture or kill a service or something like we did for the symbolicator test.

ethanhs:
@chadwhitacre there wasn't an event that caused the alert to fire, so it failed. After that commit I moved the test for alert firing to after the symbolicator event so it would succeed.

PR #1990, opened 2023-02-28
feat(replays): add replays to self hosted

JoshFerge:
- [x] Enables the replays feature flags by default for self hosted
- [x] Adds the consumers `ingest-replay-recordings` and `snuba-replays-consumer` to the docker-compose
- [x] Adds `ingest-replay-recordings` to kafka creation script (`ingest-replay-events` is automatically handled by the snuba creation script)
- [x] Adds an end-to-end test to confirm that a basic replay makes it through the system. This test checks two endpoints, one for clickhouse, and one for replay storage. 

also confirmed this works w/ self hosted by running a webpage with our SDK pointed at the DSN of the self hosted server, then seeing it in the UI.

### DIFF
--- a/_integration-test/run.sh
+++ b/_integration-test/run.sh
@@ -171,6 +171,28 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
 source _integration-test/custom-ca-roots/teardown.sh
 echo "${_endgroup}"
 
+echo "${_group}Test that replays work ..."
+echo "Creating test replay..."
+TEST_REPLAY_ID=$(
+  export LC_ALL=C
+  head /dev/urandom | tr -dc "a-f0-9" | head -c 32
+)
+TIME_IN_SECONDS=$(date +%s)
+curl -sf --data '{"event_id":"'"$TEST_REPLAY_ID"'","sdk":{"name":"sentry.javascript.browser","version":"7.38.0"}}
+{"type":"replay_event"}
+{"type":"replay_event","replay_start_timestamp":$TIME_IN_SECONDS,"timestamp":$TIME_IN_SECONDS,"error_ids":[],"trace_ids":[],"urls":["example.com"],"replay_id":"'"$TEST_REPLAY_ID"'","segment_id":0,"replay_type":"session","event_id":"'"$TEST_REPLAY_ID"'","environment":"production","sdk":{"name":"sentry.javascript.browser","version":"7.38.0"},"request":{"url":"example.com","headers":{"platform":"javascript","contexts":{"replay":{"session_sample_rate":1,"error_sample_rate":1}}}
+{"type":"replay_recording","length":19}
+{"segment_id":0}
+[]' -H 'Content-Type: application/json' -H "X-Sentry-Auth: Sentry sentry_version=7, sentry_key=$SENTRY_KEY, sentry_client=test-bash/0.1" "$SENTRY_TEST_HOST/api/$PROJECT_ID/envelope/" -o /dev/null
+
+printf "Getting the test replay back"
+REPLAY_SEGMENT_PATH="api/0/projects/sentry/internal/replays/$TEST_EVENT_ID/recording-segments/?download"
+REPLAY_EVENT_PATH="api/0/projects/sentry/internal/replays/$TEST_EVENT_ID/"
+timeout 60 bash -c 'until $(sentry_api_request "$REPLAY_EVENT_PATH" -Isf -X GET -o /dev/null); do printf '.'; sleep 0.5; done'
+timeout 60 bash -c 'until $(sentry_api_request "$REPLAY_SEGMENT_PATH" -Isf -X GET -o /dev/null); do printf '.'; sleep 0.5; done'
+echo " got it!"
+echo "${_endgroup}"
+
 # Table formatting based on https://stackoverflow.com/a/39144364
 COMPOSE_PS_OUTPUT=$(docker compose ps --format json | jq -r \
   '.[] |
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -258,6 +258,9 @@ services:
   snuba-transactions-consumer:
     <<: *snuba_defaults
     command: consumer --storage transactions --consumer-group transactions_group --auto-offset-reset=latest --max-batch-time-ms 750
+  snuba-replays-consumer:
+    <<: *snuba_defaults
+    command: consumer --storage replays --auto-offset-reset=latest --max-batch-time-ms 750
   snuba-replacer:
     <<: *snuba_defaults
     command: replacer --storage errors --auto-offset-reset=latest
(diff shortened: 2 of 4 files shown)


_integration-test/run.sh
```diff
@@ -171,6 +171,28 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
...
 echo "${_endgroup}"
 
+echo "${_group}Test that replays work ..."
+echo "Creating test replay..."
+TEST_REPLAY_ID=$(
+  export LC_ALL=C
+  head /dev/urandom | tr -dc "a-f0-9" | head -c 32
+)
+TIME_IN_SECONDS=$(date +%s)
+curl -sf --data '{"event_id":"'"$TEST_REPLAY_ID"'","sdk":{"name":"sentry.javascript.browser","version":"7.38.0"}}
```
JoshFerge, line 181: for some reason the bash syntax of `$'` for C escaped strings for newlines wasn't working so just put in the real values

_integration-test/run.sh
```diff
@@ -171,6 +171,28 @@ $dcr --no-deps web python3 /etc/sentry/test-custom-ca-roots.py
...
+{"type":"replay_event"}
+{"type":"replay_event","replay_start_timestamp":$TIME_IN_SECONDS,"timestamp":$TIME_IN_SECONDS,"error_ids":[],"trace_ids":[],"urls":["example.com"],"replay_id":"'"$TEST_REPLAY_ID"'","segment_id":0,"replay_type":"session","event_id":"'"$TEST_REPLAY_ID"'","environment":"production","sdk":{"name":"sentry.javascript.browser","version":"7.38.0"},"request":{"url":"example.com","headers":{"platform":"javascript","contexts":{"replay":{"session_sample_rate":1,"error_sample_rate":1}}}
+{"type":"replay_recording","length":19}
+{"segment_id":0}
+[]' -H 'Content-Type: application/json' -H "X-Sentry-Auth: Sentry sentry_version=7, sentry_key=$SENTRY_KEY, sentry_client=test-bash/0.1" "$SENTRY_TEST_HOST/api/$PROJECT_ID/envelope/" -o /dev/null
+
+printf "Getting the test replay back"
+REPLAY_SEGMENT_PATH="api/0/projects/sentry/internal/replays/$TEST_EVENT_ID/recording-segments/?download"
+REPLAY_EVENT_PATH="api/0/projects/sentry/internal/replays/$TEST_EVENT_ID/"
+timeout 60 bash -c 'until $(sentry_api_request "$REPLAY_EVENT_PATH" -Isf -X GET -o /dev/null); do printf '.'; sleep 0.5; done'
```
hubertdeng123, line 191: So this test itself will take more than 2 minutes? Is there any way to cut down on that time? I know dev-infra isn't super happy about self-hosted e2e tests taking over 15 minutes

JoshFerge: hmm it shouldn't, i just copied the defaults from the other tests.

JoshFerge: are we noticing its increasing the time of the test by two minutes?

JoshFerge: i can lower the timeout here if needed i don't think the tests take more than a few seconds

hubertdeng123: ahh I read that wrong, guess I'm still waking up 😅

hubertdeng123:
Awesome work @JoshFerge and the rest of the @getsentry/replay  team. Approximately how much more disk space will need to be used for replays? Right now we have a minimum requirement of 20 GB of free space to run self-hosted

hubertdeng123:
> I don't think it meaningfully impacts the amount of default disk space. if they are sending a lot of replays, they'll want to add some more disk space but i assume that's the same with any other usage of the software.

Yeah I agree with this about the default disk space part. Maybe this comes from me not understanding how replays work, but I assumed that replays would take up a good amount of storage compared to before. Is that the case? If so I'm thinking it might be useful to warn them about this. 

Reason I say this is we've gotten multiple issues of people filing issues and complaining about disk space usage in the past and it'd be nice to point them to some documentation warning them if it's expected

JoshFerge:
> > I don't think it meaningfully impacts the amount of default disk space. if they are sending a lot of replays, they'll want to add some more disk space but i assume that's the same with any other usage of the software.
> 
> Yeah I agree with this about the default disk space part. Maybe this comes from me not understanding how replays work, but I assumed that replays would take up a good amount of storage compared to before. Is that the case? If so I'm thinking it might be useful to warn them about this.
> 
> Reason I say this is we've gotten multiple issues of people filing issues and complaining about disk space usage in the past and it'd be nice to point them to some documentation warning them if it's expected

replays are about ~10x larger than an error / transaction, but generally with how people use them they send a lot less. so just kind of hard to say. the average replay is ~200KB (will vary on the characteristics of their website), so users can calculate how much extra space they need based on their expected volume.

hubertdeng123:
Gotcha, thanks for the context 😁. I think it'd be a nice to add that somewhere for our users. Maybe under https://develop.sentry.dev/self-hosted/? Any thoughts @chadwhitacre?

JoshFerge:
> Gotcha, thanks for the context 😁. I think it'd be a nice to add that somewhere for our users. Maybe under https://develop.sentry.dev/self-hosted/? Any thoughts @chadwhitacre?

it seems like we don't really call out resource usage anywhere on this page -- would lean towards not adding it for now from what I see.

docker-compose.yml
```diff
@@ -322,6 +325,9 @@ services:
   ingest-consumer:
     <<: *sentry_defaults
     command: run ingest-consumer --all-consumer-types
+  ingest-replay-recordings:
+    <<: *sentry_defaults
+    command: run ingest-replay-recordings
```
chadwhitacre, line 330: Dang 33 services 😭

```
clickhouse
cron
geoipupdate
ingest-consumer
ingest-replay-recordings
kafka
memcached
nginx
post-process-forwarder-errors
post-process-forwarder-transactions
postgres
redis
relay
sentry-cleanup
smtp
snuba-api
snuba-cleanup
snuba-consumer
snuba-outcomes-consumer
snuba-replacer
snuba-replays-consumer
snuba-sessions-consumer
snuba-subscription-consumer-events
snuba-subscription-consumer-transactions
snuba-transactions-cleanup
snuba-transactions-consumer
subscription-consumer-events
subscription-consumer-transactions
symbolicator
symbolicator-cleanup
web
worker
zookeeper
```

chadwhitacre: 11 snubas 😖

hubertdeng123:
We've decided to document the changes in the changelog for the next release and include some documentation around resource usage for replays there. Documentation in develop can come later as another issue of migrating README details to the develop docs. PR is good to merge by me

hubertdeng123 (approved):

chadwhitacre:
Seeing a working test:

https://github.com/getsentry/self-hosted/actions/runs/4288646161/jobs/7470889652#step:5:7978

docker-compose.yml
BYK: https://github.com/getsentry/snuba/issues/1670

That said this would only help with a 33% reduction.

Looks like we need a similar issue for Sentry to consolidate all those `query-subscription-consumer` and `post-process-forwarder` services.

Issue #1995, opened 2023-02-28
Remove headers from backup/restore 


hubertdeng123:
### Problem Statement

Backup/restore is currently using the command from the [docs](https://develop.sentry.dev/self-hosted/backup/):
`docker-compose run --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export > sentry/backup.json`

This is adding headers to the backup file that need to be removed before import. Try and find another way of handling this

This looks something like:
```
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
```

PR #1998, opened 2023-03-01
Remove cert upgrade headers in backup file

hubertdeng123:
closes https://github.com/getsentry/self-hosted/issues/1995

This cert upgrade was causing headers to be written to the backup file.

### DIFF
--- a/sentry/entrypoint.sh
+++ b/sentry/entrypoint.sh
@@ -2,7 +2,12 @@
 set -e
 
 if [ "$(ls -A /usr/local/share/ca-certificates/)" ]; then
-  update-ca-certificates
+  # for backups, let's redirect the console output to avoid it being written to backup file
+  if [[ $SENTRY_BACKUP == true ]]; then
+    update-ca-certificates >/dev/null
+  else
+    update-ca-certificates
+  fi
 fi
 
 if [ -e /etc/sentry/requirements.txt ]; then


sentry/entrypoint.sh
```diff
@@ -2,7 +2,12 @@
 set -e
 
 if [ "$(ls -A /usr/local/share/ca-certificates/)" ]; then
-  update-ca-certificates
+  # for backups, let's redirect the console output to avoid it being written to backup file
+  if [[ $SENTRY_BACKUP == true ]]; then
```
hubertdeng123, line 6: Probably a good idea to keep the output regularly in case it errors out for some reason. For backups, we don't want this in the json file so 👋

hubertdeng123: This will also need an update in the develop documentation as the command for backup will be
`docker-compose run --rm -T -e SENTRY_LOG_LEVEL=CRITICAL SENTRY_BACKUP=true web export > sentry/backup.json`

hubertdeng123: develop PR:
https://github.com/getsentry/develop/pull/851

hubertdeng123: Reasonable, better approach is to use something like this. Going to close this PR 
`docker-compose run -v $(pwd)/sentry:/sentry-data/backup  --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json`

writes directly to a file directly

hubertdeng123: As for 2, it feels like adding another level of indirection for logging errors can be confusing for users

PR #2000, opened 2023-03-01
fix(tests): fix replay integration test

JoshFerge:
There were various things wrong with the test and wasn't working as expected. should work now. can verify on CI with a couple of commits.


_integration-test/run.sh
```diff
@@ -173,23 +173,32 @@ echo "${_endgroup}"
...
+{"type":"replay_event","replay_start_timestamp":"'"$TIME_IN_SECONDS"'","timestamp":"'"$TIME_IN_SECONDS"'","error_ids":[],"trace_ids":[],"urls":["http://example.com/"],"replay_id":"'"$TEST_REPLAY_ID"'","segment_id":0,"replay_type":"session","event_id":"'"$TEST_REPLAY_ID"'","environment":"production","sdk":{"integrations":["InboundFilters","FunctionToString","TryCatch","Breadcrumbs","GlobalHandlers","LinkedErrors","Dedupe","HttpContext","Replay"],"version":"7.31.1","name":"sentry.javascript.react"},"sdkProcessingMetadata":{},"request":{"url":"http://example.com/","headers":{"User-Agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36"}},"platform":"javascript","tags":{"sessionSampleRate":1,"errorSampleRate":1}}
 {"type":"replay_recording","length":19}
 {"segment_id":0}
```
BYK, line 186: Looks like this data can live in a fixture file and things like `$TEST_REPLAY_ID` can be plugged in with some help from `jq`?

_integration-test/run.sh
```diff
@@ -173,23 +173,32 @@ echo "${_endgroup}"
...
+
+for i in {1..30}; do
+  REPLAY_EVENT_PROCESSED=$(sentry_api_request "$REPLAY_EVENT_PATH" -X GET | jq -r '.data.id')
+  REPLAY_RECORDING_PROCESSED=$(sentry_api_request "$REPLAY_SEGMENT_PATH" -X GET | jq .)
+  if ! [ $REPLAY_EVENT_PROCESSED = $TEST_REPLAY_ID ] || [ -z "$REPLAY_RECORDING_PROCESSED" ]; then
+    echo $i
+    sleep $i
+  else
+    break
+  fi
```
BYK, line 200: The code this replaced clearly showed the intention with the `timeout 60` part. With this one, all I see is a magic number, `30` and increased magical sleeping.

1. Why do we need to wait longer between retries?
2. Please clearly document the reasoning and the working logic of the code with comments.

JoshFerge: 1. this was a bug, thank you for the catch. 
2. added a couple of comments

_integration-test/run.sh
chadwhitacre: Fixture + jq is a good idea, but I'm not seeing existing fixtures to use as an example. Given that, I don't think we should block.

chadwhitacre:
> can verify on CI with a couple of commits

a90250d7ecae1eda92b6ed04588bc4f82328a41a ⇒ https://github.com/getsentry/self-hosted/actions/runs/4307414343

[Still working](https://github.com/getsentry/self-hosted/actions/runs/4307414343/jobs/7512452078#step:5:8335)? 🙈

hubertdeng123:
I've tested replays locally in self-hosted  and they seem to work fine. We've decided that we're ok to ship on the 15th and will circle back about writing the test case in python later on when we iron out how we want other teams to write integration tests for stuff they own.

chadwhitacre:
> circle back about writing the test case in python later on when we iron out how we want other teams to write integration tests for stuff they own.

See https://github.com/getsentry/team-ospo/issues/91

Issue #1918, opened 2023-01-17
Add integration tests for backup/restore workflow


ethanhs:
This is to split out of https://github.com/getsentry/self-hosted/issues/1761.

@BYK said the following:
> the backup/restore dance is woefully undertested. I did some fixes and improvements there while writing the docs (it did not even work before that 😅). In the upgrade scenario, you may wanna add a backup/restore scenario too. That said it should only matter when there's a model/migration change.

This is part of the overall backup/restore improvement project for Q1 2024 (TBD)

PR #2012, opened 2023-03-04
Add backup/restore integration tests

hubertdeng123:
closes https://github.com/getsentry/self-hosted/issues/1918

Since there is a unit test in sentry already covering backup/restore, this is a more broad test. It tests these functionalities
1. Backup/restore is correctly writing any sort of data to the `backup.json` file. 
2. The export command doesn't throw an exception
3. The import command doesn't throw an exception

### DIFF
--- a/_integration-test/ensure-backup-restore-works.sh
+++ b/_integration-test/ensure-backup-restore-works.sh
@@ -6,7 +6,7 @@ source install/dc-detect-version.sh
 
 echo "${_group}Test that backup/restore works..."
 echo "Creating backup..."
-$dcr -v $(pwd)/sentry:/sentry-data/backup  --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
+$dcr -v $(pwd)/sentry:/sentry-data/backup --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
 test -f sentry/backup.json
 
 # bring postgres down and recreate the docker volume
--- a/_integration-test/ensure-backup-restore-works.sh
+++ b/_integration-test/ensure-backup-restore-works.sh
@@ -14,6 +14,7 @@ $dc stop postgres
 sleep 5
 $dc rm -f -v postgres
 export SKIP_USER_CREATION=1
+source install/create-docker-volumes.sh
 source install/set-up-and-migrate-database.sh
 $dc up -d
 
--- a/_integration-test/ensure-backup-restore-works.sh
+++ b/_integration-test/ensure-backup-restore-works.sh
@@ -30,4 +30,4 @@ docker-compose up -d
 echo "Importing backup..."
 docker-compose run --rm -T web import /etc/sentry/backup.json
 
-rm $(pwd)/sentry/backup/backup.json
+rm $(pwd)/sentry/backup.json
--- a/_integration-test/ensure-backup-restore-works.sh
+++ b/_integration-test/ensure-backup-restore-works.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -ex
+
+source install/_lib.sh
+source install/dc-detect-version.sh
+
+echo "${_group}Test that backup/restore works..."
+echo "Creating backup..."
+$dcr -v $(pwd)/sentry:/sentry-data/backup  --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
+test -f sentry/backup.json
+
+# bring postgres down and recreate the docker volume
+$dc stop postgres
+sleep 5
+$dc rm -f -v postgres
+export SKIP_USER_CREATION=1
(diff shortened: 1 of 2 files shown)


_integration-test/ensure-backup-restore-works.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+set -ex
+
+source install/_lib.sh
+source install/dc-detect-version.sh
+
+echo "${_group}Test that backup/restore works..."
+echo "Creating backup..."
+$dcr -v $(pwd)/sentry:/sentry-data/backup --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
```
BYK, line 9: Probably need an ownership/permission set here. I suggest something like this:
```bash
mkdir $(pwd)/sentry/backup
chmod +w $(pwd)/sentry/backup
$dcr -v $(pwd)/sentry/backup:/sentry-data/backup --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json

Comment (hubertdeng123) [reply _integration-test/ensure-backup-restore-works.sh]: Darn looks like that isn't working, will play around and see if I can figure something out. Script seems to run fine locally though

Comment (hubertdeng123) [reply _integration-test/ensure-backup-restore-works.sh]: For some reason, write access for group + owner wasn't enough. I needed write access for all users

Comment (BYK) [reply _integration-test/ensure-backup-restore-works.sh]: Probably because the user and group IDs are different when running the sentry container (from the one GitHub actions bash script runner)

Comment (BYK) [reply _integration-test/ensure-backup-restore-works.sh]: Probably because the user and group IDs are different when running the sentry container (from the one GitHub actions bash script runner)

File: integration-test.sh [hunk 2]
```diff
@@ -25,3 +26,4 @@ echo "Testing in-place upgrade and customizations"
 ./install.sh --minimize-downtime
 _integration-test/run.sh
 _integration-test/ensure-customizations-work.sh
+_integration-test/ensure-backup-restore-works.sh
```
chadwhitacre, line 29: A little goofy that we have to call this twice, but I guess that's where we're at and we clean up the test suite under https://github.com/getsentry/team-ospo/issues/91.

_integration-test/ensure-backup-restore-works.sh
```diff
@@ -0,0 +1,35 @@
...
+$dc stop postgres
+sleep 5
+$dc rm -f -v postgres
+docker volume rm sentry-postgres
+export SKIP_USER_CREATION=1
+source install/create-docker-volumes.sh
+source install/set-up-and-migrate-database.sh
+$dc up -d
+
+# echo "Importing backup..."
```
chadwhitacre, line 32: echo?

_integration-test/ensure-backup-restore-works.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo "${_group}Test that backup/restore works..."
+echo "Creating backup..."
+backup_path="$(pwd)/sentry/backup"
+mkdir -p $backup_path
+# Docker was giving me permissioning issues when trying to create this file and write to it even after giving read + write access
+# to group and owner. Instead, try creating the empty file and then give everyone write access to the backup file
+touch $backup_path/backup.json
+chmod 666 $backup_path/backup.json
+$dcr -v $backup_path:/sentry-data/backup -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
```
chadwhitacre, line 15: Would love to see this match [the docs](https://develop.sentry.dev/self-hosted/backup/#backup) exactly.

chadwhitacre: With a comment saying as much.

chadwhitacre: If you can't get it _exactly_ like the docs that's fine, just say so in the comment.

_integration-test/ensure-backup-restore-works.sh
```diff
@@ -0,0 +1,35 @@
...
+sleep 5
+$dc rm -f -v postgres
+docker volume rm sentry-postgres
+export SKIP_USER_CREATION=1
+source install/create-docker-volumes.sh
+source install/set-up-and-migrate-database.sh
+$dc up -d
+
+# echo "Importing backup..."
+$dcr --rm -T web import /etc/sentry/backup/backup.json
```
chadwhitacre, line 33: This matches the docs basically. Fine to use $dcr I guess vs. writing it out (like in the docs) if $dcr is necessary to get the test infra to work.

hubertdeng123: it's not necessary to get it to work, For this use case I can use the exact command the docs is using

_integration-test/ensure-backup-restore-works.sh
hubertdeng123: ah that makes sense

chadwhitacre (approved):

PR #2029, opened 2023-03-14
Add Backup/restore scripts

hubertdeng123:
1. adds backup and restore scripts
2. added error reporting to both of the scripts
3. consolidate utility scripts to a helper script file


utilities.sh
```diff
@@ -0,0 +1,68 @@
...
+  done
+
+  # If we have a version given, switch to it.
+  if [ -n "$version" ]; then
+    git checkout "$version"
+  fi
+}
+
+function backup() {
+  docker-compose run -v $(pwd)/sentry:/sentry-data/backup --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
```
BYK, line 63: I'd ensure the permissions here with `chmod +w` or something.

utilities/set-up-error-reporting-for-scripts.sh
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -eEuo pipefail
```
hubertdeng123, line 2: Is there a reason why we're not including the `-E` flag generally? If an error is thrown in a function, our cleanup function actually doesn't catch it properly and treats it as the `EXIT` trap code instead of `ERR`. Found that out in this PR

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

chadwhitacre: > Is there a reason why we're not including the `-E` flag generally?

No good one that I know of. :)

chadwhitacre (changes requested):
Brainstormed #2035, buuuuuuuuut ...

- move everything from `utilities` into `scripts` for now
- rename `clean.sh` back to `reset.sh`
- DRY up the `{backup,reset,restore}.sh` scripts amap (while keeping separate .sh files)

chadwhitacre (approved):
Ship it! :)

PR #2050, opened 2023-03-28
Support external sourcemaps bigger, than 1Mb

le0pard:
Self hosted sentry limited in sourcemap external size, which can fetch and cache to 1Mb. 

> Remote file too large for caching

This is because `memcached` the maximum size of a value you can store in is 1 megabyte by default. But this can be fixed by `-I` param.

This PR allow to fetch and cache external sourcemaps for sentry up to 25Mb.

image

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,6 +70,7 @@ x-sentry-defaults: &sentry_defaults
     # on the host system (or in the .env file)
     SENTRY_EVENT_RETENTION_DAYS:
     SENTRY_MAIL_HOST:
+    SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
   volumes:
     - "sentry-data:/data"
     - "./sentry:/etc/sentry"
@@ -107,10 +108,6 @@ services:
     <<: *restart_policy
     image: "memcached:1.6.9-alpine"
     command: ["-I", "${SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:-1M}"]
-    environment:
-      # Leaving the value empty to just pass whatever is set
-      # on the host system (or in the .env file)
-      SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
     healthcheck:
       <<: *healthcheck_defaults
       # From: https://stackoverflow.com/a/31877626/5155484
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -106,7 +106,7 @@ services:
   memcached:
     <<: *restart_policy
     image: "memcached:1.6.9-alpine"
-    command: "-I 25M"
+    command: ["-I", "25M"]
     healthcheck:
       <<: *healthcheck_defaults
       # From: https://stackoverflow.com/a/31877626/5155484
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -106,6 +106,7 @@ services:
   memcached:
     <<: *restart_policy
     image: "memcached:1.6.9-alpine"
+    command: "-I 25M"
```
BYK, line 109: Better to use the array notation?
```suggestion
    command: ["-I", "25M"]
```

sentry/sentry.conf.example.py
```diff
@@ -107,6 +107,9 @@ def get_internal_network():
         "BACKEND": "django.core.cache.backends.memcached.MemcachedCache",
         "LOCATION": ["memcached:11211"],
         "TIMEOUT": 3600,
+        "OPTIONS": {
+            "server_max_value_length": 1024 * 1024 * 25,
```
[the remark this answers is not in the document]
le0pard: @BYK changed in https://github.com/getsentry/self-hosted/pull/2050/commits/6117ec862a115f8c263911b0ef3a88388fb40189

docker-compose.yml
```diff
@@ -106,6 +106,11 @@ services:
   memcached:
     <<: *restart_policy
     image: "memcached:1.6.9-alpine"
+    command: ["-I", "${SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:-1M}"]
+    environment:
+      # Leaving the value empty to just pass whatever is set
+      # on the host system (or in the .env file)
+      SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
```
BYK, line 113: Don't think you need this as the memcached image should be oblivious to any `SENTRY_` env variable :) Instead, you should move `SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:` part to line 73, next to `x-sentry-defaults`' environment section

sentry/sentry.conf.example.py
```diff
@@ -3,6 +3,15 @@
...
 from sentry.conf.server import *  # NOQA
 
+UNIT_SIZES={
+    "K": 1024,
+    "M": 1024**2,
+    "G": 1024**3,
+}
+def unit_text_to_bytes(text):
+    multiplier = UNIT_SIZES[text[-1]]
+    return float(text[:-1])*multiplier
```
BYK, line 13: Now that you put my code directly there, I think we can make it even simpler 😅 
```suggestion
BYTE_MULTIPLIER = 1024
UNITS = ("K", "M", "G")
def unit_text_to_bytes(text):
		unit = text[-1].upper()
		power = UNITS.index(unit) + 1
    return float(text[:-1])*(BYTE_MULTIPLIER**power)
```

BYK (changes requested):
Rejecting for the `environment` docker compose config placement

docker-compose.yml
le0pard: fixed by https://github.com/getsentry/self-hosted/pull/2050/commits/4881f244bdc866f8e195b7371bc402d75ed46073

hubertdeng123 (approved):

PR #2074, opened 2023-04-06
Upgrade Postgres to 14

hubertdeng123:
This bumps the default version of Postgres in self-hosted to 14

https://github.com/getsentry/self-hosted/issues/1610

This needs to go in before the sentry PR to change the default version of postgres due to CI failures in sentry
https://github.com/getsentry/sentry/pull/46999


.pre-commit-config.yaml
```diff
@@ -16,7 +16,6 @@ repos:
   rev: v4.3.0
   hooks:
   - id: check-case-conflict
-  - id: check-executables-have-shebangs
```
hubertdeng123, line 19: all of our `*.sh` files in the `install/` directory don't have shebangs, let's get rid of this

docker-compose.yml
```diff
@@ -126,7 +126,7 @@ services:
         hard: 10032
   postgres:
     <<: *restart_policy
-    image: "postgres:9.6"
+    image: "ghcr.io/getsentry/image-mirror-library-postgres:14"
```
aldy505, line 129: why aren't we using the original `postgres` image?

note: iirc, I know that sentry mirrors the original postgres image, but people might ask these stuff

.pre-commit-config.yaml
chadwhitacre: They're also not executable.

```
/Users/chadwhitacre/workbench/getsentry/self-hosted/install
total 88
-rw-r--r--  1 chadwhitacre  staff  1666 Mar 17 15:04 _lib.sh
-rw-r--r--  1 chadwhitacre  staff   208 Mar 17 15:04 _min-requirements.sh
-rw-r--r--  1 chadwhitacre  staff   164 Mar 17 15:04 bootstrap-snuba.sh
-rw-r--r--  1 chadwhitacre  staff   446 Mar 17 15:04 build-docker-images.sh
-rw-r--r--  1 chadwhitacre  staff   642 Mar 17 15:04 check-latest-commit.sh
-rw-r--r--  1 chadwhitacre  staff  2937 Mar 17 15:04 check-minimum-requirements.sh
-rw-r--r--  1 chadwhitacre  staff   520 Mar 17 15:04 create-docker-volumes.sh
-rw-r--r--  1 chadwhitacre  staff   694 Mar 17 15:04 create-kafka-topics.sh
-rw-r--r--  1 chadwhitacre  staff   716 Mar 17 15:04 dc-detect-version.sh
-rw-r--r--  1 chadwhitacre  staff  1151 Mar 17 15:04 detect-platform.sh
-rw-r--r--  1 chadwhitacre  staff   211 Mar 17 15:04 ensure-files-from-examples.sh
-rw-r--r--  1 chadwhitacre  staff  1684 Mar 17 15:04 ensure-relay-credentials.sh
-rw-r--r--  1 chadwhitacre  staff  7610 Mar 17 15:04 error-handling.sh
-rw-r--r--  1 chadwhitacre  staff   641 Mar 17 15:04 generate-secret-key.sh
-rw-r--r--  1 chadwhitacre  staff   964 Mar 17 15:04 geoip.sh
-rw-r--r--  1 chadwhitacre  staff  1531 Mar 17 15:04 install-wal2json.sh
-rw-r--r--  1 chadwhitacre  staff  2719 Mar 17 15:04 parse-cli.sh
-rw-r--r--  1 chadwhitacre  staff   383 Mar 17 15:04 set-up-and-migrate-database.sh
-rw-r--r--  1 chadwhitacre  staff   352 Mar 17 15:04 turn-things-off.sh
-rw-r--r--  1 chadwhitacre  staff   608 Mar 17 15:04 update-docker-images.sh
-rw-r--r--  1 chadwhitacre  staff   972 Mar 17 15:04 wrap-up.sh
```

hubertdeng123: 🤦‍♂️

docker-compose.yml
hubertdeng123: Great question, this mirrors the original postgres image but uses an older version of the tagged `postgres:14` image, particularly postgres 14.6. I figured it might be good to use the same image that we use in `getsentry/sentry` dev here for consistency purposes, although it shouldn't make a difference

chadwhitacre: Worth commenting that in the file?

hubertdeng123: done

chadwhitacre (approved):

docker-compose.yml
BYK: Please don't do this as it will block users in China who are blocked by sanctions or some other reasons and are not able to use Google container hosting service. I fear same may apply for GitHub's. See #445.

install/upgrade-postgres.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "Due to glibc change re-indexing"
+  echo "Starting up new PostgreSQL version"
+  $dc up -d postgres
+
+  # Wait for postgres
+  RETRIES=5
+  until docker exec sentry-self-hosted-postgres-1 psql -U postgres -c "select 1" >/dev/null 2>&1 || [ $RETRIES -eq 0 ]; do
+    echo "Waiting for postgres server, $((RETRIES--)) remaining attempts..."
+    sleep 1
+  done
```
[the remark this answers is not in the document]
hubertdeng123: Seems like docker compose v1 will no longer be supported in June 2023 🤔. Would rather not drop support for v1 here in this PR and wait a bit longer to do that

We currently use `pg_isready` for a healthcheck which seems sufficient enough for a healthcheck for Postgres

docker-compose.yml
hubertdeng123: fair point, will change this

install/upgrade-postgres.sh
```diff
@@ -0,0 +1,44 @@
+echo "${_group}Ensuring proper PostgreSQL version ..."
+
+if [[ -n "$(docker volume ls -q --filter name=sentry-postgres)" && "$(docker run --rm -v sentry-postgres:/db busybox cat /db/PG_VERSION 2>/dev/null)" == "9.6" ]]; then
+  docker volume rm sentry-postgres-new || true
```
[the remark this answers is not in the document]
hubertdeng123: I don't like introducing an additional prompt asking for the user to continue with a warning since for people who have automated upgrade scripts it will cause those to start hanging (us and our daily upgrades 😅). It feels like adding an additional install option like we do for `minimize-downtime` and `skip-commit-check` also isn't super applicable here since this is a one time thing too.

stayallive:
Am I reading this right that it basically duplicates the database to a new volume?

Doesn't this require double-ish disk space for the duration? This seems like a non-trivial thing to not mention if so 😅

hubertdeng123:
You're right @stayallive. I'll add this to the release notes as this should be made more clear

stayallive:
Just an FYI, I am manually running this script line by line and am running into some issues I want to document for other readers.

I started when this was still called `onpremise` and thus my Postgres containers boot up as: `sentry_onpremise_postgres_1` instead of `sentry-self-hosted-postgres-1` which fails to reindex, manually changing the container names works just fine of course. Not sure what would happen if I would have run the script (assuming it just skips the indexing).

Also I am running the migration on one of my modest ~200GB deployments and it takes quite a hot minute to run this migration causing multi-hour donwtime (currently projecting 3-4 hours). The upgrade copies the data over to a new volume but copying the data back to the old volume name also takes a long time and reindex is also not that quick. Most of this is a docker limitation (not being able to rename the container) but was a bit unexpected going into this.

This is all not an issue that needs to be fixed (althought the container names might be fixable by running this through docker-compose instead) but I feel like the warning should be a little harder since this will cause long downtimes and undefined behaviour if you do not have double disk space available for the temp postgres volumes on larger deployments.

BYK:
>I started when this was still called onpremise and thus my Postgres containers boot up as: sentry_onpremise_postgres_1 instead of sentry-self-hosted-postgres-1 which fails to reindex, manually changing the container names works just fine of course. Not sure what would happen if I would have run the script (assuming it just skips the indexing).

I think if you run `docker compose down` it will just remove the containers and then later recreate them and you should not have any issues. There's no data stored inside containers, only in volumes, which have consistent names that did not change. Especially for the database.

hubertdeng123:
I've put up a PR to resolve the inconsistencies between docker volume names by using `docker compose exec postgres` instead:
https://github.com/getsentry/self-hosted/pull/2096

Just tested it locally and it seems to work for me. Unfortunately this won't go in until the next release but hopefully that will make the upgrade process more smooth. Thanks for bringing these issues up 🙏

PR #2099, opened 2023-04-20
Enable upstream keepalive

otbutz:
This enables keepalive connections for both upstream blocks, following the best practices as recommended by nginx:

https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#no-keepalives

> We recommend setting the parameter to twice the number of servers listed in the upstream{} block.

As there's only a single server in each block, we keep 2 idle connections.

### Legal Boilerplate

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -55,10 +55,12 @@ http {
 
 	upstream relay {
 		server relay:3000;
+		keepalive 2;
 	}
 
 	upstream sentry {
 		server web:9000;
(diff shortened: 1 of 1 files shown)


BYK (approved):
As long as the tests pass, this sounds like a great change to me. Did you observe any performance improvements?

otbutz:
Our instance is probably too small to see any measurable differences.

hubertdeng123 (approved):

PR #2102, opened 2023-04-21
Fix issue to run some queries and VACUUM command in database after update to PostgreSQL 14

le0pard:
Fix issue to run some queries and VACUUM command in database after update to PostgreSQL 14

Without this changes some SQL command start return error: `ERROR: could not resize shared memory segment . . . : No space left on device`

### Legal Boilerplate


BYK:
I feel like this is an instance-specific setting and value and may not be applicable to everyone. I recommend using docket-compose.override.yml file to set this only for yourself instead of hard-coding a value in the repo.

le0pard:
Ok, thanks

PR #2111, opened 2023-04-27
Slightly relax reporting consent

chadwhitacre:
We're talking about sending CSPs to dogfood. We're like 99% covered by existing ("runtime errors" is pretty broad). This should put us at 100% covered.


hubertdeng123 (approved):

chadwhitacre:
We need to be explicit about what we collect, no loopholes.

PR #2114, opened 2023-04-28
Uninstall Docker Compose v1 from CI so it's not used for tests

hubertdeng123:
This will uninstall docker compose v1.29 from CI so `docker-compose` cannot be used.

CI runs to indicate this is the case:
1. https://github.com/getsentry/self-hosted/actions/runs/4832862658/jobs/8612163711?pr=2114
1. https://github.com/getsentry/self-hosted/actions/runs/4832862658/jobs/8612163856?pr=2114

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,8 @@ jobs:
           sudo mkdir -p "${{ matrix.compose_path }}"
           sudo curl -L https://github.com/docker/compose/releases/download/${{ matrix.compose_version }}/docker-compose-`uname -s`-`uname -m` -o "${{ matrix.compose_path }}/docker-compose"
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
+          # Docker Compose v1 is installed here, remove it
+          sudo rm -f "/usr/local/bin/docker-compose"
 
       - name: Integration Test
         run: ./integration-test.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,12 +64,12 @@ jobs:
           # Always remove `docker compose` support as that's the newer version
           # and comes installed by default nowadays.
           sudo rm -f "/usr/local/lib/docker/cli-plugins/docker-compose"
+          # Docker Compose v1 is installed here, remove it
+          sudo rm -f "/usr/local/bin/docker-compose"
           sudo rm -f "${{ matrix.compose_path }}/docker-compose"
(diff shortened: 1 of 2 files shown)


_integration-test/ensure-backup-restore-works.sh
```diff
@@ -11,24 +11,24 @@ echo "Creating backup..."
 touch $(pwd)/sentry/backup.json
 chmod 666 $(pwd)/sentry/backup.json
 # Command here matches exactly what we have in our docs https://develop.sentry.dev/self-hosted/backup/#backup
-docker-compose run -v $(pwd)/sentry:/sentry-data/backup --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
+$dc run -v $(pwd)/sentry:/sentry-data/backup --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export /sentry-data/backup/backup.json
```
hubertdeng123, line 14: using `$dc` since our scripts do some pre-script setup that is failing CI (error-handling, dc-detect-version, detect-platform, trap)

.github/workflows/test.yml
```diff
@@ -68,6 +68,8 @@ jobs:
           sudo mkdir -p "${{ matrix.compose_path }}"
           sudo curl -L https://github.com/docker/compose/releases/download/${{ matrix.compose_version }}/docker-compose-`uname -s`-`uname -m` -o "${{ matrix.compose_path }}/docker-compose"
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
+          # Docker Compose v1 is installed here, remove it
+          sudo rm -f "/usr/local/bin/docker-compose"
```
BYK, line 72: Let's move this up, next to other `rm` commands.

_integration-test/ensure-backup-restore-works.sh
BYK: Yisss, exactly the cases we wanted to fix.

BYK (approved):

.github/workflows/test.yml
aminvakil: This should be after downloading docker-compose, right?

aminvakil (approved):

.github/workflows/test.yml
BYK: Doesn't really matter. I'd do it before to be consistent and not accidentally delete the downloaded one.

chadwhitacre (approved):

PR #2115, opened 2023-04-29
feat: add suggested fix feature

### LINKED ISSUE #2068: Add Suggested Fix
[New feature](https://docs.sentry.io/product/issues/issue-details/suggested-fix/)! Wat do?

aldy505:
Closes #2068 

Users should only need to add `OPENAI_API_TOKEN=<their token value>` to their `.env.` or `.env.custom` file. That's about it. It works magically.

A screenshot from my self-hosted instance:

image_2023-04-29_20-20-18

### Legal Boilerplate

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -291,3 +291,19 @@ GEOIP_PATH_MMDB = '/geoip/GeoLite2-City.mmdb'
 
 # BITBUCKET_CONSUMER_KEY = 'YOUR_BITBUCKET_CONSUMER_KEY'
 # BITBUCKET_CONSUMER_SECRET = 'YOUR_BITBUCKET_CONSUMER_SECRET'
+
+##############################################
+# Suggested Fix Feature / OpenAI Integration #
+##############################################
+
+# See https://docs.sentry.io/product/issues/issue-details/ai-suggested-solution/
+# for more information about the feature. Make sure the OpenAI's privacy policy is
+# aligned with your company.
+
+# Set the OPENAI_API_KEY on the .env or .env.custom file with a valid
+# OpenAI API key to turn on the feature.
+OPENAI_API_KEY = env("OPENAI_API_KEY", "")
+
+if OPENAI_API_KEY and not OPENAI_API_KEY.isspace():
+  SENTRY_FEATURES["organizations:open-ai-suggestion"] = True
+  SENTRY_FEATURES["organizations:open-ai-suggestion-new-design"] = True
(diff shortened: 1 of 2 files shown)


sentry/sentry.conf.example.py
```diff
@@ -291,3 +291,19 @@ def get_internal_network():
...
+# for more information about the feature. Make sure the OpenAI's privacy policy is
+# aligned with your company.
+
+# Set the OPENAI_API_KEY on the .env or .env.custom file with a valid
+# OpenAI API key to turn on the feature.
+OPENAI_API_KEY = env("OPENAI_API_KEY", "")
+
+if OPENAI_API_KEY and not OPENAI_API_KEY.isspace():
+  SENTRY_FEATURES["organizations:open-ai-suggestion"] = True
+  SENTRY_FEATURES["organizations:open-ai-suggestion-new-design"] = True
```
aldy505, line 309: As a side note:

You can actually just enable the `organizations:open-ai-suggestion` feature flag. Enabling both will gives you 2 section of suggested fix (before the stacktraces and after the breadcrumbs). But there's no way we can enable the `new-design` flag without enabling the first flag. Tried it before.

If we want some stability we can just enable only the `organizations:open-ai-suggestion` flag.

chadwhitacre: Let's hold off on the `-new-design` flag for now since it seems to still be under development and not ready for prime time.

aldy505: Done

hubertdeng123 (approved):

sentry/sentry.conf.example.py
```diff
@@ -291,3 +291,18 @@ def get_internal_network():
...
+
+# See https://docs.sentry.io/product/issues/issue-details/ai-suggested-solution/
+# for more information about the feature. Make sure the OpenAI's privacy policy is
+# aligned with your company.
+
+# Set the OPENAI_API_KEY on the .env or .env.custom file with a valid
+# OpenAI API key to turn on the feature.
+OPENAI_API_KEY = env("OPENAI_API_KEY", "")
+
+if OPENAI_API_KEY and not OPENAI_API_KEY.isspace():
```
chadwhitacre, line 307: Interesting. I would usually do `env("OPENAI_API_KEY", "").strip()`. ¯\\\_(ツ)\_/¯

chadwhitacre (approved):

sentry/sentry.conf.example.py
aldy505: I don't do Python so much, so that's what I ended up doing :D

chadwhitacre: Ah, gotcha. :)

Honestly I don't think we even really need the check. In general it would be unusual (possibly even difficult) to end up with an envvar value that was only whitespace. If it were a likely problem it should be solved in [the `env` helper function](https://github.com/getsentry/sentry/blob/23.4.0/src/sentry/conf/server.py#L67-L79) (it's not). Let's do without it.

```suggestion
if OPENAI_API_KEY:
```

PR #2146, opened 2023-05-16
fix(suggested-fix): key should be 'key', not 'token'

aldy505:
I was upgrading my Sentry instance to the latest version, then I spent like 15 minutes wondering on why the Suggested Fix feature I [made the PR of](https://github.com/getsentry/self-hosted/pull/2115) is gone. I thought the Sentry team changed the logic of that. It turns out I made a typo. 

This is how I feel right now:

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,7 +71,7 @@ x-sentry-defaults: &sentry_defaults
     SENTRY_EVENT_RETENTION_DAYS:
     SENTRY_MAIL_HOST:
     SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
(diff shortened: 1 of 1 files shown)


aldy505:
It uses new design now (by default, see https://github.com/getsentry/sentry/pull/47531). Here are a few screenshots:

image

image

image

BYK (approved):
This is why we should have tests 🙂 (not your fault, the repo is not in great shape to add more tests for new features).

Bump on that Playwright patch maybe @chadwhitacre @hubertdeng123 ?

chadwhitacre (approved):

Issue #1838, opened 2022-11-29
Add Profiling


chadwhitacre:
Arriving here from a hallway chat with @jernejstrasner @indragiek @phacops @hubertdeng123. We should aim to ship Profiling in self-hosted as close to GA as possible, ideally part of the same announcement but it's understood this is not a GA blocker.

## To Do

- [x] https://github.com/getsentry/team-ospo/issues/51
- [x] https://github.com/getsentry/team-profiling/issues/188
- [x] Add something to e2e test.
- [x] Add to `docker-compose.yml`.
- [x] Add to CalVer release pipeline.

jernejstrasner:
@indragiek what's the plan on this one?

chadwhitacre:
Putting this to backlog for now, let's revisit in Jan/Feb when closer to GA.

irudoy:
Any ETA on this?

phacops:
@irudoy a bit late to the party but we're targetting end of March for all this to be completed.

chadwhitacre:
See #1873 and #1993 for examples to follow.

jamesob:
Would love to see this feature in self-hosted.

aldy505:
Just a heads up: when this is up and running, and people are submitting way too much profiling data from their applications, they can offload the profiling data into Azure Blob Storage, AWS S3 bucket, or GCP's GCS by providing (or maybe overriding) the value on the `SENTRY_BUCKET_PROFILES` environment variable. The default value is that it will write to a local filesystem (that will probably mounted to a Docker volume. See [here](https://github.com/getsentry/vroom/blob/6cc05b2e2652015f75f6b9b12c88f46fa486d2b8/Dockerfile#L17)).

A few examples:
- AWS S3: `s3://my-bucket?region=us-west-1&awssdk=2&endpoint=minio.selfhosted.com&s3ForcePathStyle=false&disableSSL=true`

    Do provide additional environment variables (if needed):
    - AWS_ACCESS_KEY=foobar
    - AWS_SECRET_KEY=foobar
    - AWS_SESSION_TOKEN=foobar (not required)

   Or, if you have machine running on AWS already, see https://docs.aws.amazon.com/sdk-for-go/api/aws/session/#pkg-overview for other methods on authenticating to S3.
- Azure Blob: `azblob://my-container?protocol=http&domain=localhost:10001&protocol=string&localemu=false&cdn=false`

    Do provide additional environment variables (if needed):
[...]

PR #2154, opened 2023-05-17
feat(profiling): Run profiling on self-hosted

phacops:
This PR aims to add support for profiling on self-hosted.

Fixes https://github.com/getsentry/self-hosted/issues/1838.


docker-compose.yml
```diff
@@ -384,6 +384,12 @@ services:
         <<: *depends_on-healthy
       web:
         <<: *depends_on-healthy
+  vroom:
+    <<: *restart_policy
+    image: "VROOM_IMAGE"
```
aldy505, line 389:
```suggestion
    image: "$VROOM_IMAGE"
```

Should be inferred from variable

aldy505 (commented):
I don't see any configuration for integrating it with the `web` or `snuba` containers? Or is it meant to be just like this? I'd like to try it on my own server 😄

phacops:
@aldy505 Good point, I added those.

docker-compose.yml
```diff
@@ -400,6 +420,8 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
+  sentry-vroom:
+    external: true
```
aldy505, line 424: Don't forget to add a line about this here: https://github.com/getsentry/self-hosted/blob/master/install/create-docker-volumes.sh

docker-compose.yml
```diff
@@ -283,6 +283,12 @@ services:
   snuba-subscription-consumer-transactions:
     <<: *snuba_defaults
     command: subscriptions-scheduler-executor --dataset transactions --entity transactions --auto-offset-reset=latest --no-strict-offset-reset --consumer-group=snuba-transactions-subscriptions-consumers --followed-consumer-group=transactions_group --delay-seconds=60 --schedule-ttl=60 --stale-threshold-seconds=900
+  snuba-profiling-profiles-consumer:
+    <<: *snuba_defaults
+    command: consumer --storage profiles --auto-offset-reset=latest --max-batch-time-ms 1000 --no-strict-offset-reset
+  snuba-profiling-functions-consumer:
+    <<: *snuba_defaults
+    command: consumer --storage functions --auto-offset-reset=latest --max-batch-time-ms 1000 --no-strict-offset-reset
```
aldy505, line 291: Should these be added to the list of sentry image's depends_on field?

Here: https://github.com/getsentry/self-hosted/blob/8aa571cb758406eccb7aefe13b3b7ebf1a34e5c2/docker-compose.yml#L39-L50

docker-compose.yml
```diff
@@ -384,6 +393,19 @@ services:
...
       web:
         <<: *depends_on-healthy
+  vroom:
+    <<: *restart_policy
+    image: "$VROOM_IMAGE"
+    environment:
+      SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
+      SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
+      SENTRY_BUCKET_PROFILES:
+      SENTRY_SNUBA_HOST: "$SNUBA"
```
aldy505, line 403: I don't think this reads from environment variable defined here, as this is hardcoded.

https://github.com/getsentry/self-hosted/blob/8aa571cb758406eccb7aefe13b3b7ebf1a34e5c2/docker-compose.yml#L60

Should we also hardcode the address of Snuba API host?

Unless you want to do `"${SNUBA:-http://snuba-api:3128}"`

docker-compose.yml
```diff
@@ -384,6 +393,19 @@ services:
         <<: *depends_on-healthy
       web:
         <<: *depends_on-healthy
+  vroom:
```
aldy505, line 396: Shouldn't we add the `vroom` address to the `x-sentry-default` key (or `web` container) as defined here?

https://github.com/getsentry/sentry/blob/39233a25d1bee5fd4dffa754b0b618128ef3deaa/src/sentry/conf/server.py#L3201

aldy505 (commented):
Actually I'm a bit curious about the `SENTRY_DSN` environment variable that can be applied to snuba and vroom container. If they're left empty, I suppose no error message will be sent to the `internal` project of people's self-hosted  instance.

Vroom's defined here: https://github.com/getsentry/vroom/blob/c2cf9653f5bb9b2e4a02f0548956b790547e9048/cmd/vroom/config.go#L8

docker-compose.yml
```diff
@@ -351,6 +357,9 @@ services:
         BASE_IMAGE: sentry-self-hosted-local
     entrypoint: "/entrypoint.sh"
     command: '"0 0 * * * gosu sentry sentry cleanup --days $SENTRY_EVENT_RETENTION_DAYS"'
+  ingest-profiles:
```
hubertdeng123, line 360: small thing, can this be moved up to be under these lines?

https://github.com/getsentry/self-hosted/blob/8aa571cb758406eccb7aefe13b3b7ebf1a34e5c2/docker-compose.yml#L332-L337

This file is pretty large and it would keep things a bit more organized

docker-compose.yml
hubertdeng123: +1, seeing this in CI

https://github.com/getsentry/self-hosted/actions/runs/5015430579/jobs/9018870128?pr=2154

``` shell
+ docker compose --ansi never run --rm web createuser --force-update --superuser --email test@example.com -***
  time="2023-05-19T21:14:23Z" level=warning msg="The \"SNUBA\" variable is not set. Defaulting to a blank string."
```

hubertdeng123 (commented):
Will profiling be released under CalVer? If so, would need to add `VROOM` to 
https://github.com/getsentry/self-hosted/blob/1fa77341a49180755edf84450419f9733fce0bab/scripts/bump-version.sh#L10

phacops:
> Actually I'm a bit curious about the `SENTRY_DSN` environment variable that can be applied to snuba and vroom container. If they're left empty, I suppose no error message will be sent to the `internal` project of people's self-hosted instance.

That's correct, the Sentry SDK won't be activated and nothing will be sent. I was expecting the user to pass the right DSN via an environment variable. Do you know how other services are handling that?

aldy505:
> That's correct, the Sentry SDK won't be activated and nothing will be sent. I was expecting the user to pass the right DSN via an environment variable. Do you know how other services are handling that?

We might as well need to ping @hubertdeng123 for help. 

There is an "internal" project that should've been set up during the first install of Sentry, but I don't know on how to acquire the DSN and put it into the environment variable / configuration that can be consumed by Snuba and Vroom.

hubertdeng123:
> There is an "internal" project that should've been set up during the first install of Sentry, but I don't know on how to acquire the DSN and put it into the environment variable / configuration that can be consumed by Snuba and Vroom.

Snuba is very similar, [retrieves `SENTRY_DSN` from the environment](https://github.com/getsentry/snuba/blob/9d503a951019a4ebad5df6d418c9a950ca6c8595/snuba/settings/settings_self_hosted.py#L25). I've never played around with it before but it should just be possible to put it into the configuration through the `docker-compose.yml` file. I don't think this is a large concern for this PR though. For the record, looks like a DSN for the internal project can be shown by going to `/settings/${org}/projects/internal/keys/`

aldy505:
I guess we can skip that one for now?

hubertdeng123:
> I guess we can skip that one for now?

Yep

_unit-test/create-docker-volumes-test.sh
```diff
@@ -15,14 +15,13 @@ sentry-kafka
...
 sentry-redis
 sentry-symbolicator
+sentry-vroom
 sentry-zookeeper"
 
 before=$(get_volumes)
 
 test "$before" == "" || test "$before" == "$expected_volumes"
 
-source install/create-docker-volumes.sh
```
hubertdeng123, line 24: not sure why this was ever duplicated even after looking through the file history, and tests seem to be passing fine without it

.env
```diff
@@ -9,7 +9,9 @@ SENTRY_IMAGE=getsentry/sentry:nightly
 SNUBA_IMAGE=getsentry/snuba:nightly
 RELAY_IMAGE=getsentry/relay:nightly
 SYMBOLICATOR_IMAGE=getsentry/symbolicator:nightly
+VROOM_IMAGE=getsentry/vroom:latest
```
aldy505, line 12: wait.. whaaaat? 🫨

_unit-test/create-docker-volumes-test.sh
aldy505: maybe testing for duplicate volume creation?

.env
phacops: This should be `nightly`, for testing purposes i changed it but now we are finally publishing `nightly` tags properly.

aldy505: Ah ok. Understandable.

.env
```diff
@@ -9,7 +9,9 @@ SENTRY_IMAGE=getsentry/sentry:nightly
 SNUBA_IMAGE=getsentry/snuba:nightly
 RELAY_IMAGE=getsentry/relay:nightly
 SYMBOLICATOR_IMAGE=getsentry/symbolicator:nightly
+VROOM_IMAGE=getsentry/vroom:nightly
 WAL2JSON_VERSION=latest
 HEALTHCHECK_INTERVAL=30s
 HEALTHCHECK_TIMEOUT=60s
 HEALTHCHECK_RETRIES=10
+SENTRY_BUCKET_PROFILES=file://localhost//var/lib/sentry-profiles
```
BYK, line 17: Why is this an env variable and not in a config file?

_unit-test/create-docker-volumes-test.sh
```diff
@@ -15,14 +15,13 @@ sentry-kafka
 sentry-postgres
 sentry-redis
 sentry-symbolicator
+sentry-vroom
```
BYK, line 18: I feel like this is not a great practice (creating and relying on global volumes). Can we not make this a "local" volume? Do we need to absolutely make sure the data here should survive a potentially accidental `docker compose down -v`?

docker-compose.yml
```diff
@@ -283,6 +283,12 @@ services:
   snuba-subscription-consumer-transactions:
     <<: *snuba_defaults
     command: subscriptions-scheduler-executor --dataset transactions --entity transactions --auto-offset-reset=latest --no-strict-offset-reset --consumer-group=snuba-transactions-subscriptions-consumers --followed-consumer-group=transactions_group --delay-seconds=60 --schedule-ttl=60 --stale-threshold-seconds=900
+  snuba-profiling-profiles-consumer:
```
BYK, line 286: Yay, more snuba containers! Let's add another strike to getsentry/snuba#1670

docker-compose.yml
BYK: It should be in the config file below (or through the env variable as you suggested, yes). Otherwise it won't (shouldn't?) work

scripts/bump-version.sh
```diff
@@ -7,7 +7,7 @@ NEW_VERSION="$2"
 WAL2JSON_VERSION=${WAL2JSON_VERSION:-$(curl -s "https://api.github.com/repos/getsentry/wal2json/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")')}
 
 sed -i -e "s/^WAL2JSON_VERSION=\([^:]\+\):.\+\$/WAL2JSON_VERSION=\1:$WAL2JSON_VERSION/" .env
-sed -i -e "s/^\(SENTRY\|SNUBA\|RELAY\|SYMBOLICATOR\)_IMAGE=\([^:]\+\):.\+\$/\1_IMAGE=\2:$NEW_VERSION/" .env
+sed -i -e "s/^\(SENTRY\|SNUBA\|RELAY\|SYMBOLICATOR|VROOM\)_IMAGE=\([^:]\+\):.\+\$/\1_IMAGE=\2:$NEW_VERSION/" .env
```
BYK, line 10: Maybe we should consider making this a blocklist now since almost all images get updated?

BYK (commented):
There should be at least a cursory test sending a profile and making sure it is ingested before this is merged. Otherwise someone will eventually break and nobody will notice it until its too late :) (true story)

BYK:
Regarding `SENTRY_DSN`, it _is_ created during the initial setup. It's just tricky to extract it out of the container and then putting in a "permanent" and shared place for other containers to consume. Can be dumped into a shared volume maybe?

.env
aldy505: It's because.. uhhh... https://github.com/getsentry/vroom/blob/bdda64cc531de12ff86fc2a547850712a231ceeb/cmd/vroom/config.go#L4-L19

aldy505: (Git blamed to my commit lol)

_unit-test/create-docker-volumes-test.sh
phacops: This will end up with profile data, it's something you'd likely want persisted by default.

phacops:
> There should be at least a cursory test sending a profile and making sure it is ingested before this is merged. Otherwise someone will eventually break and nobody will notice it until its too late :) (true story)

It was added.

_unit-test/create-docker-volumes-test.sh
BYK: Local volumes are also persisted by default. They are just not "globally" available by docker containers outside of this project and you can remove them with `docker compose down -v`.

I'd say we should stop adding new global volumes and keep everything local unless we have _very_ good reasons and we understand them fully.

_unit-test/create-docker-volumes-test.sh
BYK: Yes, this was to make sure trying to create volumes over and over again does not error. I think, since we run the whole installation twice now, this is redundant.

fixtures/envelope-with-profile.json
```diff
@@ -0,0 +1,5 @@
...
...,11,12,13,14,15,16],[21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[50,16],[49,44,45,46,47,43,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[51,52,48,49,44,45,46,47,43,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[52,48,49,44,45,46,47,43,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16],[53,54,55,51,52,48,49,44,45,46,47,43,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37,38,39,40,41,42,9,10,11,12,13,14,15,16]],"thread_metadata":{"0":{"name":"main"}}},"transaction":{"name":"fib: sourcemaps here","id":"66578634d48d433db0ad52882d1efe5b","trace_id":"33321bfbd5304bcc9663d1b53b08f84e","active_thread_id":"0"}}
```
BYK, line 5: Not saying this is too large but if there's a "minimally viable profile file" that's smaller than this, I'd prefer that 😅

_unit-test/create-docker-volumes-test.sh
aldy505: Profiling data should survive accidental `docker compose down`, though. Although AFAIK, [vroom](https://github.com/getsentry/vroom) doesn't have a cleanup mechanism for profiling data that exceeds the maximum sentry retention period. I guess for now, every profiling data stored on this volume will be persisted forever.

fixtures/envelope-with-profile.json
hubertdeng123, line 1: This doesn't seem to be a valid json file, seems to be an envelope file as far as I can tell? Maybe that's why the test is failing?

_unit-test/create-docker-volumes-test.sh
chadwhitacre: Switching from global to local volumes seems like a reasonable improvement, but I don't think we should make @phacops implement that for us. Let's let him use the status quo patterns and make the switch in a separate PR.

BYK: @chadwhitacre this one does not require a refactor. Just not adding it here and removing the `external: true` line in the compose file will do it. We already have other local volumes.

fixtures/envelope-with-profile.json
phacops: Test was failing because `--data` wasn't respecting new lines. Using `--data-binary` solves this.

_unit-test/create-docker-volumes-test.sh
phacops: @BYK It's definitely an issue we didn't really think about since we were relying on GCS retention settings to remove profiles when needed. I should be addressed separately in `vroom`.

phacops: @BYK Now, the reason why I set the `external: true` was because I wanted the volume to be persisted and according to the comment we have for non-external volumes, it suggests they won't be persisted. Seems like the comment is misguided though.

docker-compose.yml
aldy505: We should be good to go after this one is addressed, right? @phacops

docker-compose.yml
```diff
@@ -421,8 +421,9 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
+  # not being external will still persist data across restarts
+  # it won't persist if someone does a docker compose down -v
   sentry-vroom:
-    external: true
   # These store ephemeral data that needn't persist across restarts.
```
BYK, line 427: We should remove this comment as you suggested earlier to not confuse future folks. Definitely not accurate. (I mean we don't _need_ this data to persist through restarts but they do regardless)
```suggestion
```

BYK (approved):
🚀

.env
phacops: @BYK I put this here as I understood that's where you put environment variables users are susceptible to edit for their needs. Now, this one is a bit special because if you wanted to put a different storage location (like a bucket on GCS or S3), you'd have to pass additional environment variables and thus modify the `docker-compose.yml` file anyway.

Is this usually how you are doing this to customize the behavior of some services?

docker-compose.yml
phacops: I added `vroom` as a `depends_on` since `web` needs to POST to `vroom`, but then the consumers only depends on `kafka` technically. I think it's good like this.

hubertdeng123 (approved):

.env
BYK: @phacops not really unless the setting applies to multiple services (like retention days). The typical way is to have a `<service_name>` folder with an example config we provide for defaults and then folks can customize it to their needs. See https://github.com/getsentry/self-hosted/tree/master/sentry or https://github.com/getsentry/self-hosted/tree/master/relay for examples.

hubertdeng123:
23.5.2 is out now, this can be merged now

chadwhitacre (approved):

docker-compose.yml
```diff
@@ -384,6 +395,21 @@ services:
...
+  vroom:
+    <<: *restart_policy
+    image: "$VROOM_IMAGE"
+    environment:
+      SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
+      SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
+      SENTRY_BUCKET_PROFILES: file://localhost//var/lib/sentry-profiles
+      SENTRY_SNUBA_HOST: "http://snuba-api:1218"
+    ports:
+      - "8085:8085/tcp"
```
aldy505, line 407: The port shouldnt be exposed. 

```suggestion
```

docker-compose.yml
aldy505: +1

docker-compose.yml
```diff
@@ -384,6 +395,21 @@ services:
         <<: *depends_on-healthy
       web:
         <<: *depends_on-healthy
+  vroom:
+    <<: *restart_policy
+    image: "$VROOM_IMAGE"
+    environment:
+      SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
+      SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
+      SENTRY_BUCKET_PROFILES: file://localhost//var/lib/sentry-profiles
```
aldy505, line 404: I'd prefer:

```suggestion
      SENTRY_BUCKET_PROFILES: "${SENTRY_BUCKET_PROFILES:-file://localhost//var/lib/sentry-profiles}"
```

That way, people can adjust their vroom storage backend config easier (ie. they can use GCS or S3 or Azure Blob), defined from .env or .env.custom without having them to change docker-compose.yml

aldy505: Or, you might prefer `VROOM_BUCKET_PROFILES`.

aldy505 (commented):
Just a little bit more..

Note about the exposed port: if you don't specify it as exposed port, that port will still be opened to other containers. Having a new exposed ports might cause a vulnerability for people who didn't set up a proper OS firewall.

docker-compose.yml
BYK: I'd discourage this as it might be better to override the compose file with `docker-compose.override.yml` until we get a dedicated `vroom` config file.

aldy505: Okay, agreed.

aldy505 (commented):
LGTM!

docker-compose.yml
phacops: I added to the comment instead, noting we don't need to persist this data but it's persisted anyway. I think removing it removes information about the kind of data stored.

scripts/bump-version.sh
chadwhitacre: https://github.com/getsentry/self-hosted/pull/2207 ftr, missing escape of the `|` led to [a corrupt 23.6.0 release](https://github.com/getsentry/self-hosted/blob/23.6.0/.env#L11-L12). 🐭

PR #2006, opened 2023-03-03
feat(snuba): Add snuba sessions subscription service

klboke:
Related issues, pr

- https://github.com/getsentry/snuba/issues/3696
- https://github.com/getsentry/snuba/pull/3697
- https://github.com/getsentry/snuba/pull/3804

Add snuba sessions subscription to self-hosted. cc @lynnagara 

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #2181, opened 2023-06-01
feat(subscription): Add sentry sessions subscription service

klboke:
Resolve the same issue as #2006 .

### Legal Boilerplate


hubertdeng123:
I've been putting this PR on hold because I'd like to know more information about sessions deprecation, which is ongoing right now.

klboke:
@hubertdeng123  Are there any latest updates regarding this PR?

hubertdeng123:
@klboke Unfortunately, we won't be making this change. We need to deprecate sessions completely and it's on our plate for next quarter. Currently, features for Sentry are being built on the metrics backend.

Issue #2076, opened 2023-04-11
Drop Support for Docker Compose v1


hubertdeng123:
We should drop support for Docker Compose v1 since it will no longer be supported starting June 2023. Also potentially add Docker Compose v3 sometime in the future

PR #2187, opened 2023-06-05
Remove docker compose v1

hubertdeng123:
docker is removing support for v1 after June 2023, so let's get rid of it in self-hosted

closes https://github.com/getsentry/self-hosted/issues/2076


install/dc-detect-version.sh
```diff
@@ -8,12 +8,10 @@ fi
 
 echo "${_group}Initializing Docker Compose ..."
 
-# Some environments still use `docker-compose` even for Docker Compose v2.
```
hubertdeng123, line 11: Is this ever true? I think some environments allow `docker-compose` to be used even on Docker Compose v2, but the official syntax should be `docker compose` instead of `docker-compose`. At least specified in the docs:
https://docs.docker.com/compose/migrate/

chadwhitacre: We found it to be true in the wild, hence the accommodation. Could dig in history and find the PR for more detail. It was a linux distro that munged the name, don't remember which.

.github/workflows/test.yml
```diff
@@ -44,11 +44,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Disabled due to https://github.com/getsentry/self-hosted/issues/1415
-          # - compose_version: "1.28.0"
-          #   compose_path: "/usr/local/bin"
-          # - compose_version: "1.29.2"
-          #   compose_path: "/usr/local/bin"
```
chadwhitacre, line 51: 😬

install/_min-requirements.sh
```diff
@@ -1,6 +1,6 @@
 # Don't forget to update the README and othes docs when you change these!
 MIN_DOCKER_VERSION='19.03.6'
-MIN_COMPOSE_VERSION='1.28.0'
+MIN_COMPOSE_VERSION='2.0.0'
```
chadwhitacre, line 3: Let's have this match exactly (including patch version) the min version in our test matrix. Does 2.0.0 really work? 🤔 If not, update the test matrix (good way to find out btw ;), else update this to 2.0.1.

hubertdeng123: Yep, it exists https://github.com/docker/compose/releases/tag/v2.0.0

install/check-minimum-requirements.sh
```diff
@@ -20,14 +20,14 @@ if [[ "$(vergte ${DOCKER_VERSION//v/} $MIN_DOCKER_VERSION)" -eq 1 ]]; then
 fi
 echo "Found Docker version $DOCKER_VERSION"
 
-COMPOSE_VERSION=$($dc_base version --short || echo '')
+COMPOSE_VERSION=$(docker compose version --short || echo '')
```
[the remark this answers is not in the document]
hubertdeng123: `$dc_base` is removed, but I can keep it too if it's useful to keep that abstracted. The reason I removed it is because it seemed to only be used to determine if `docker compose` or `docker-compose` should be used

install/_min-requirements.sh
hubertdeng123: ah looks like there's a reason why we're using 2.0.1 here
https://stackoverflow.com/questions/69524003/cannot-install-docker-compose-v2-invalid-plugin-exec-format-error

install/check-minimum-requirements.sh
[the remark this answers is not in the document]
BYK: +1 for removal for now.

BYK (approved):
🔥

install/check-minimum-requirements.sh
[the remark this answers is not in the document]
chadwhitacre: I bet we're going to hear about this from users. 🐭

install/dc-detect-version.sh
```diff
@@ -8,12 +8,10 @@ fi
...
-dc_base="$(docker compose version &>/dev/null && echo 'docker compose' || echo 'docker-compose')"
```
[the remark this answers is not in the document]
hubertdeng123: fair enough, we can keep this around then 👍

aminvakil (approved):

chadwhitacre (approved):

install/dc-detect-version.sh
NotSqrt: See https://docs.docker.com/compose/install/standalone/ 
> For example type `docker-compose up` when using Compose standalone, instead of `docker compose up`.

PR #2193, opened 2023-06-07
ref(compose): Separate ingest consumers

jan-auer:
Ingest consumers will stop to support the `--all-consumer-types` option, and
instead will need to run separate instances for events, attachments, and
transactions.

See https://github.com/getsentry/sentry/pull/49957

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -332,9 +332,15 @@ services:
   worker:
     <<: *sentry_defaults
     command: run worker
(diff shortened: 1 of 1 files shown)


BYK:
Anyway we can keep the `--all` option around for small-scale self-hosted versions?

ashwoods (approved):

untitaker (approved):

PR #2203, opened 2023-06-14
ref: Move most consumers to unified consumer CLI

untitaker:
Context: https://github.com/getsentry/sentry/pull/50364

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -334,19 +334,19 @@ services:
     command: run worker
   events-consumer:
     <<: *sentry_defaults
-    command: run ingest-consumer --consumer-type=events
+    command: run consumer ingest-events --consumer-group ingest-consumer
   attachments-consumer:
     <<: *sentry_defaults
-    command: run ingest-consumer --consumer-type=attachments
+    command: run consumer ingest-attachments --consumer-group ingest-consumer
   transactions-consumer:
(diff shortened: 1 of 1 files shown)


BYK:
```
KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: profiles: Broker: Unknown topic or partition"}
```

This is from profiles consumer. There might be others, check the logs for more 🙂

untitaker:
thank byk for helping out here!

> KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: profiles: Broker: Unknown topic or partition"}

The problem is, I am not even changing the topic here -- the old CLI is also depending on the same topic.

if I understand it right, this script was supposed to create those topics anyway: https://github.com/getsentry/self-hosted/blob/a52de4ab1bdb6e1cc2d33b8c0f72414b28d48649/install/create-kafka-topics.sh#L1

hubertdeng123:
seems odd that the integration test on docker compose v2.0.1 is succeeding 🤔

hubertdeng123:
now it's passing, for the record I've never seen that error before and it should not have failed twice in a row

hubertdeng123 (approved):
🚀🚀

PR #2211, opened 2023-06-17
feat: vroom cleanup script that respects default retention days

aldy505:
Conversation context: https://github.com/getsentry/vroom/issues/268

I haven't test this one out, I'll have to turn on a baremetal server on my office that's dedicated for experiment purposes to make this (https://github.com/getsentry/opentelemetry-demo) up and running, send stuff to my selfhosted sentry, and see the result on this new container.

A few Q&A:

Q: A new container???
A: Relax, this one is very lightweight and does not involve anything other than cron & file deletion.

Q: Why are you using `debian:stable-slim`, not `$VROOM_IMAGE`?
A: This:
```
[+] Running 1/1
 ! vroom-cleanup Warning                                                                                                                                                                                    3.3s
[+] Building 0.7s (6/9)
 => [vroom-cleanup internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 469B                                                                                                                                                                        0.0s
 => [vroom-cleanup internal] load .dockerignore                                                                                                                                                             0.0s
 => => transferring context: 2B                                                                                                                                                                             0.0s
 => [vroom-cleanup internal] load metadata for docker.io/getsentry/vroom:23.6.1                                                                                                                             0.0s
 => [vroom-cleanup 1/5] FROM docker.io/getsentry/vroom:23.6.1                                                                                                                                               0.1s
 => [vroom-cleanup internal] load build context                                                                                                                                                             0.0s
 => => transferring context: 35B                                                                                                                                                                            0.0s
 => ERROR [vroom-cleanup 2/5] RUN if [ -z "${http_proxy}" ]; then echo "Acquire::http::proxy "${http_proxy}";" >> /etc/apt/apt.conf; fi                                                                     0.6s
------
 > [vroom-cleanup 2/5] RUN if [ -z "${http_proxy}" ]; then echo "Acquire::http::proxy "${http_proxy}";" >> /etc/apt/apt.conf; fi:
#0 0.520 /bin/sh: can't create /etc/apt/apt.conf: nonexistent directory
------
failed to solve: process "/bin/sh -c if [ -z \"${http_proxy}\" ]; then echo \"Acquire::http::proxy \\\"${http_proxy}\\\";\" >> /etc/apt/apt.conf; fi" did not complete successfully: exit code: 1
```

### Legal Boilerplate


BYK:
At this point, the command seems like it can be part of `sentry-cleanup`?

BYK:
>At first, I'm thinking of making a new binary at vroom repo to handle cleanup (although, yes this might be more overkill than just doing find -delete). So that everything that's profiling related goes to the vroom service.

Don't think it would be overkill as I'm sure there will be more logic need to be baked into that one-liner as time passes. Consolidating everything related to a product into its binary/code makes total sense to me.

I wish we could just use the `vroom` image for this tho. I'm not familiar with it but if it is a binary, can we make the docker image a very plain one with a flexible entrypoint/cmd (similar to Sentry image actually) so we can just override that without it running extra processes in the background automatically?

aldy505:
What do you guys think about that? @phacops @hubertdeng123 

If migrating the vroom binary might be an issue, I can think of building multiple binaries on the Dockerfile and changing the entrypoint on the self-hosted compose file.

hubertdeng123:
>Don't think it would be overkill as I'm sure there will be more logic need to be baked into that one-liner as time passes. Consolidating everything related to a product into its binary/code makes total sense to me.

Upvote. Using the vroom image is something I would like to see and would be preferred here if we must use a separate container to run the cleanup for vroom, if adding the one liner into sentry-cleanup is not feasible. 

From conversation in https://github.com/getsentry/vroom/issues/268
> If we need something more generic (which, I'm against since it's better handled differently for each cloud providers), we'll make a new command in this repository, listing files needing to be deleted from ClickHouse and we'll use the library we have to delete them.

@phacops how likely do you see this evolving from the one-liner cleanup mentioned [here](https://github.com/getsentry/vroom/issues/268#issuecomment-1590074754)? From the self-hosted side here at least, it seems better to have the cleanup logic live in vroom

aldy505:
I won't be free for the next 2 weeks, so I'll drop this here https://github.com/getsentry/vroom/pull/284. Feel free to decide anything.

aldy505:
I'm free now. Any updates?

hubertdeng123:
Talked internally with @phacops, we've agreed to use debian on the vroom image so this PR should be able to use the vroom image without the error you're seeing. It seems like since this cleanup job won't need to evolve over time it's safe to use the one-liner.

hubertdeng123:
Waiting for https://github.com/getsentry/vroom/pull/290 to go in

aldy505:
Updated to use vroom's image

phacops (approved):

hubertdeng123 (approved):

Issue #857, opened 2021-02-11
Support Kafka topic configuration migration/synchronization.


BYK:
There are several topic configurations that are required for efficient or correct message processing in certain circumstances:

- log compaction settings for the commit log topic
- timestamp type for topics that are consumed as ticks subscription data

/cc @chadwhitacre @mattrobenolt @getsentry/sns

mattrobenolt:
We have plans to execute on work for this, we just haven't done it yet.

BYK:
Hence the "Status: Accepted" label :) Is this on @getsentry/single-tenant's roadmap or should I keep it on myself?

mattrobenolt:
I guess we don't have this in Asana, I'll figure out how to add this. We've talked about it very recently, I guess didn't track it anywhere, but I'd say yeah, this is our responsibility.

BYK:
>I guess we don't have this in Asana, I'll figure out how to add this. 

I just moved this from Asana (was tracked on an OSS Team board) for the sake of external visibility and our effort to use GitHub more. Up to you if you still want a copy there but they should be linked.

/cc @chadwhitacre

mattrobenolt:
That's fine, I dunno where we stand on the tools we use, but I know all of our stuff is in Asana these days. I don't make these decisions. I don't even know how to use the tools.

chadwhitacre:
Does this belong in `onpremise` or should we move to `sentry`? Where will the code live?

BYK:
I would assume the code to live here at the beginning until it is more broadly adopted (may move into its own repo or under getsentry/sentry).

PR #2224, opened 2023-06-21
ref: Move all consumers to unified consumer CLI

hubertdeng123:
Putting this up again since PR was reverted here: https://github.com/getsentry/self-hosted/pull/2223

Will need to investigate e2e test failures related to profiles topic not being found, resulting in CI being blocked on `getsentry/sentry`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -346,7 +346,7 @@ services:
     command: run consumer ingest-replay-recordings --consumer-group ingest-replay-recordings
   ingest-profiles:
     <<: *sentry_defaults
-    command: run consumer ingest-profiles --consumer-group ingest-profiles --no-strict-offset-reset
+    command: run consumer --no-strict-offset-reset ingest-profiles --consumer-group ingest-profiles
   post-process-forwarder-errors:
     <<: *sentry_defaults
     command: run post-process-forwarder --entity errors
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -334,19 +334,19 @@ services:
     command: run worker
   events-consumer:
     <<: *sentry_defaults
-    command: run ingest-consumer --consumer-type=events
+    command: run consumer ingest-events --consumer-group ingest-consumer
   attachments-consumer:
     <<: *sentry_defaults
-    command: run ingest-consumer --consumer-type=attachments
+    command: run consumer ingest-attachments --consumer-group ingest-consumer
   transactions-consumer:
     <<: *sentry_defaults
-    command: run ingest-consumer --consumer-type=transactions
+    command: run consumer ingest-transactions --consumer-group ingest-consumer
   ingest-replay-recordings:
     <<: *sentry_defaults
-    command: run ingest-replay-recordings
+    command: run consumer ingest-replay-recordings --consumer-group ingest-replay-recordings
   ingest-profiles:
     <<: *sentry_defaults
-    command: run ingest-profiles --no-strict-offset-reset
+    command: run consumer ingest-profiles --consumer-group ingest-profiles --no-strict-offset-reset
   post-process-forwarder-errors:
     <<: *sentry_defaults
     command: run post-process-forwarder --entity errors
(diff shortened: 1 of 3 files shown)


BYK:
We may need to add some topics here: https://github.com/getsentry/self-hosted/blob/master/install/create-kafka-topics.sh

hubertdeng123:
> We may need to add some topics here: https://github.com/getsentry/self-hosted/blob/master/install/create-kafka-topics.sh

I've double checked, profiles seems to be in that file. It's also the name of the topic used in sentry 🤔
https://github.com/getsentry/sentry/blob/d1c17c650ec94cc85b1277e61b971ee89703b0b2/src/sentry/runner/commands/run.py#L638

Why is this change introducing so much flakiness in our e2e tests...

docker-compose.yml
```diff
@@ -334,19 +334,19 @@ services:
...
-    command: run ingest-consumer --consumer-type=transactions
+    command: run consumer ingest-transactions --consumer-group ingest-consumer
   ingest-replay-recordings:
     <<: *sentry_defaults
-    command: run ingest-replay-recordings
+    command: run consumer ingest-replay-recordings --consumer-group ingest-replay-recordings
   ingest-profiles:
     <<: *sentry_defaults
-    command: run ingest-profiles --no-strict-offset-reset
+    command: run consumer ingest-profiles --consumer-group ingest-profiles --no-strict-offset-reset
```
BYK, line 349: I think this change may fix the issue:
```suggestion
    command: run consumer --no-strict-offset-reset ingest-profiles --consumer-group ingest-profiles
```

Reason being, the new `consumer` command expecting a positional `topic` argument at the end and I suspect putting `--no-strict-offset-reset` at the end might be confusing that.

hubertdeng123: thanks for the suggestion, let me try that

BYK:
Sorry this didn't work. Looks like this maybe a bug in Kafka itself: https://github.com/confluentinc/confluent-kafka-dotnet/issues/1766

The solution maybe to just add some retry logic there for this particular error. I think Snuba have something like this already.

lynnagara:
Isn't the issue this line? https://github.com/getsentry/self-hosted/blob/a81cb1e9d9e3407933cfdb1420ee85189f86749b/install/create-kafka-topics.sh#L8

Since there are many topics which contain the substring `profiles` at this point, we entirely skip over creation.

BYK:
@lynnagara that's a good catch. That said I _think_ @hubertdeng123 tried to creat topics without checking at all earlier here: [`55441d0` (#2224)](https://github.com/getsentry/self-hosted/pull/2224/commits/55441d055c03d262f78c8b81e3ed33cd39c71657)

untitaker:
@lynnagara indeed good catch. `-w` flag was supposed to take care of this (i think), but it didn't work for me locally

untitaker:
tests are green 😎

untitaker:
[`55441d0` (#2224)](https://github.com/getsentry/self-hosted/pull/2224/commits/55441d055c03d262f78c8b81e3ed33cd39c71657) would've almost worked as well if we added `|| true` to catch the error when the topic already exists. why the topic would already exist I don't know

lynnagara:
> looking at the CI output it seems that [`55441d0` (#2224)](https://github.com/getsentry/self-hosted/pull/2224/commits/55441d055c03d262f78c8b81e3ed33cd39c71657) would've almost worked as well if we added `|| true` to catch the error when the topic already exists. why the topic would already exist I don't know

A few topics exist b/c `snuba bootstrap` runs before this and creates them

lynnagara (approved):

untitaker:
> A few topics exist b/c snuba bootstrap runs before this and creates them

should we remove it? having two ways to create topics sounds sketchy. I don't know if we can make it so that sentry creates some of the topics and snuba some of the others, it's unclear to me which service "owns" a particular topic

lynnagara:
> > A few topics exist b/c snuba bootstrap runs before this and creates them
> 
> should we remove it? having two ways to create topics sounds sketchy. I don't know if we can make it so that sentry creates some of the topics and snuba some of the others, it's unclear to me which service "owns" a particular topic

The thing is right now Snuba actually does create them with the correct configs, so it's a bit better to create those ones there. We should move the source of truth for all topics and topic configuration into `sentry-kafka-schemas` and avoid all this?

BYK:
Wow, great catch @lynnagara and @untitaker. I think I'm responsible for that fuzzy topic matching so I apologize for the pain I caused 😅 

> We should move the source of truth for all topics and topic configuration into sentry-kafka-schemas and avoid all this?

I present you another ancient issue then: #857 😀

PR #2235, opened 2023-06-26
Update nginx to 1.25.1

williamdes:
LGTM the changelog: https://nginx.org/en/CHANGES-1.24
1.23 LGTM: https://nginx.org/en/CHANGES
and 1.22 only had a .1 with CVE fixes

1.25 is still a bit new, maybe we should wait before bumping: https://endoflife.date/nginx

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -372,7 +372,7 @@ services:
     <<: *restart_policy
     ports:
       - "$SENTRY_BIND:80/tcp"
-    image: "nginx:1.22.0-alpine"
+    image: "nginx:1.24.0-alpine"
     volumes:
       - type: bind
         read_only: true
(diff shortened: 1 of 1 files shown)


williamdes:
Gosh the CI is flaky: https://github.com/getsentry/self-hosted/actions/runs/5381345860/jobs/9765286778#step:4:12472

aminvakil (changes requested):
We have a very simple use of nginx and I don't think we will face a problem using 1.25.1, let's bump it to latest version.

aminvakil:
Can you also please correct the PR title?

aminvakil (approved):

hubertdeng123 (approved):

PR #2236, opened 2023-06-26
Update geoipupdate to 5.1.1

williamdes:
I have read all changelogs between the two versions:
- https://github.com/maxmind/geoipupdate/releases
- https://github.com/maxmind/geoipupdate/releases?page=2

Since https://github.com/maxmind/geoipupdate/releases/tag/v4.8.0 there is two new ENVs to avoid overriding the entrypoint 
@chadwhitacre you did comment on #766 back in the day https://github.com/getsentry/self-hosted/pull/766/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3
Should we use ENVs now ?

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -243,7 +243,7 @@ services:
       timeout: 10s
       retries: 30
   geoipupdate:
-    image: "maxmindinc/geoipupdate:v4.7.1"
+    image: "maxmindinc/geoipupdate:v5.1.1"
     # Override the entrypoint in order to avoid using envvars for config.
     # Futz with settings so we can keep mmdb and conf in same dir on host
     # (image looks for them in separate dirs by default).


docker-compose.yml
```diff
@@ -243,7 +243,7 @@ services:
       timeout: 10s
       retries: 30
   geoipupdate:
-    image: "maxmindinc/geoipupdate:v4.7.1"
+    image: "maxmindinc/geoipupdate:v5.1.1"
```
chadwhitacre, line 246: [Looks like](https://hub.docker.com/r/maxmindinc/geoipupdate) they are moving to [GHCR](https://github.com/maxmind/geoipupdate/pkgs/container/geoipupdate). Should we update here?

chadwhitacre:
> Should we use ENVs now?

Seems reasonable. https://github.com/maxmind/geoipupdate/pull/143 does look like it addresses our use-case directly. :)

docker-compose.yml
williamdes: It will cause setup problems as the ghcr images require github ghcr docker login to pull

chadwhitacre: Hrm, okay. Looks like 5.1.1 is on both so this is fine for now.

hubertdeng123 (approved):

chadwhitacre:
Thanks for the contribution, @williamdes! :)

PR #2254, opened 2023-07-06
Fix command called in reset script

stayallive:
Looks like this was missed in #2029, running reset.sh now runs the bash `reset` and just resets the console 😅

### DIFF
--- a/scripts/_lib.sh
+++ b/scripts/_lib.sh
@@ -17,7 +17,7 @@ function confirm() {
 # The purpose of this script is to make it easy to reset a local self-hosted
 # install to a clean state, optionally targeting a particular version.
 
-function clean() {
+function reset() {
   # If we have a version given, validate it.
   # ----------------------------------------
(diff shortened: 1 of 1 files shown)


hubertdeng123:
The purpose of this script is actually to run the clean function that lives in the `scripts/_lib.sh` file. Was this something that was unclear? What would you have wanted the functionality to be?

stayallive:
Well yes. But the cmd in the file was “reset” and not “clean”, see the diff. 

So as is it calls “reset” instead of the expected “clean”.

hubertdeng123:
> Well yes. But the cmd in the file was defined as “reset” and not “clean”, see the diff, so the wrong function is called.

There actually isn't a `reset` function in the file. Maybe what might be cleaner is changing clean -> reset in `scripts/_lib.sh`?

stayallive:
Sure, that also works, updated 👍

hubertdeng123:
thanks for catching this @stayallive

hubertdeng123 (approved):

PR #2274, opened 2023-07-17
Move open ai key from env variables

hubertdeng123:
Let's not store this key in the env variables to reduce the risk of potentially leaking secrets.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,6 +74,7 @@ x-sentry-defaults: &sentry_defaults
     SENTRY_EVENT_RETENTION_DAYS:
     SENTRY_MAIL_HOST:
     SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
+    # Set this value if you plan on using the Suggested Fix Feature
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -74,6 +74,7 @@ x-sentry-defaults: &sentry_defaults
     SENTRY_EVENT_RETENTION_DAYS:
     SENTRY_MAIL_HOST:
     SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
+    # Set this value if you plan on using the Suggested Fix Feature
```
azaslavsky, line 77: What is this feature? Are there docs for it anywhere? This implies we're using ChatGPT under the hood for it?

azaslavsky (approved):

docker-compose.yml
hubertdeng123: Docs are here: 
https://docs.sentry.io/product/issues/issue-details/ai-suggested-solution/

We're using ChatGPT to recommend solutions for issues

Issue #2260, opened 2023-07-09
healthcheck incorrectly assumes  gnu nc in redis alpine image


aep:
### Self-Hosted Version

23.6.2

### CPU Architecture

x86_64

### Docker Version

24.0.4

### Docker Compose Version

v2.19.1

### Steps to Reproduce

git clone https://github.com/getsentry/self-hosted.git
cd self-hosted
./install.sh

### Expected Result

working

### Actual Result

install script gets stuck likely  because redis never becomes healthy

this is because the healthcheck fails with

```
nc: unrecognized option: q
BusyBox v1.36.1 (2023-06-02 00:42:02 UTC) multi-call binary.

Usage: nc [OPTIONS] HOST PORT  - connect\nnc [OPTIONS] -l -p PORT [HOST] [PORT]  - listen\n\n\t-e PROG\tRun PROG after connect (must be last)\n\t-l\tListen mode, for inbound connects\n\t-lk\tWith -e, provides persistent server\n\t-p PORT\tLocal port\n\t-s ADDR\tLocal address\n\t-w SEC\tTimeout for connects and final net reads\n\t-i SEC\tDelay interval for lines sent\n\t-n\tDon't do DNS resolution\n\t-u\tUDP mode\n\t-b\tAllow broadcasts\n\t-v\tVerbose\n\t-o FILE\tHex dump traffic\n\t-z\tZero-I/O mode (scanning)\n"

```

hubertdeng123:
Thanks for bringing this up, we may need to put up a fix for this. Seems like this is coming from this line:
https://github.com/getsentry/self-hosted/blob/5e1e74b39131d8d7e56c388089c25a2574d9734f/docker-compose.yml#L178

Could you try removing the -q option to see if it works for you?

aep:
Yeah works without q

PR #2275, opened 2023-07-17
Remove nc -q option

hubertdeng123:
The nc -q option is failing for some users as seen [here](https://github.com/getsentry/self-hosted/issues/2260). Seems like -q does this from [docs](https://docs.oracle.com/cd/E86824_01/html/E54763/netcat-1.html) and seems safe to remove:
After receiving EOF on stdin, wait for specified number of seconds and quit.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -175,7 +175,7 @@ services:
     healthcheck:
       <<: *healthcheck_defaults
       test:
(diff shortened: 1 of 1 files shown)


azaslavsky (approved):

chadwhitacre:
Did this show up in our dogfooding instance? 👀

PR #2297, opened 2023-07-25
Resolve Safari Content-Disposition header bug

### LINKED ISSUE #2285: Unable to open Sentry with Safari, possibly related to Content-Encoding: gzip
### Self-Hosted Version

23.7.0

### CPU Architecture

x86_64

### Docker Version

24.0.2

### Docker Compose Version

2.18.1

### Steps to Reproduce

Go to the sentry main page with the latest Safari browser (16.5.1).

### Expected Result

I was expecting a styled and working login page.

### Actual Result

It doesn't work. Browser Console shows:
 
screenshot_2023-07-20_at_07 26 03_720

This appears to be related to app.js being transfer-encoded with gzip, but Safari not picking this up and trying to parse the gzipped data as javascript. This does work with Chrome for example.

Note that we're running Sentry behind Caddy, which may be interfering with headers.

### Event ID

_No response_

azaslavsky:
With release 23.7.0, the upstream sentry@23.7.0 upgraded Django to 3.2.20, a breaking change for that dependency. This introduced a number of bugs into self-hosted, including one where the Content-Disposition response header is now set by Django in a manner that is illegible to Safari, as documenting in https://code.djangoproject.com/ticket/31082.

This change modifies the default nginx config to override this header with the simpler `inline` setting, allowing Safari to properly decompress gzipped resources like CSS and Javascript bundles.

Fixes #2285

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -43,6 +43,9 @@ http {
 	proxy_next_upstream error timeout invalid_header http_502 http_503 non_idempotent;
 	proxy_next_upstream_tries 2;
 
+	# Must be removed to play nice with Safari, see the discussion in: getsentry/self-hosted#2285.
+	proxy_hide_header Content-Disposition;
+
 	# Remove the Connection header if the client sends it,
 	# it could be "close" to close a keepalive connection
 	proxy_set_header Connection '';


hubertdeng123 (approved):

mwarkentin:
I've opened a more targeted fix here: https://github.com/getsentry/self-hosted/pull/2455

PR #2355, opened 2023-08-18
Add metrics and generic metrics backend

hubertdeng123:


aldy505:
10 new containers on this PR. As far as I experienced this, each Snuba container consumes around 50-100 MB of RAM, and each Sentry ingest container consumes around 60-130 MB of RAM. I'd assume this feature would require at least additional 1 GB of RAM tops (assuming every each of it consumes 100 MB of RAM). Should we bump the minimum specs on the requirements? https://github.com/getsentry/self-hosted/blob/8c294f577bab901272d536c35a082879d6417331/README.md?plain=1#L5-L11

BYK:
Or should we finally invest in a unified Snuba instance and unified ingest processes?

hubertdeng123:
Yep, was pretty painful adding another 10 containers here.

> Or should we finally invest in a unified Snuba instance and unified ingest processes?

Let me see what I can do here in the following weeks.

williamdes:
> Or should we finally invest in a unified Snuba instance and unified ingest processes?

That would be cool !
There is already too much containers to look after

And if you want a supervisor alternative, there is https://github.com/FedericoPonzi/Horust that I got into alpine edge testing ;)

aldy505:
> > Or should we finally invest in a unified Snuba instance and unified ingest processes?
> 
> 
> 
> That would be cool !
> 
> There is already too much containers to look after
> 
> 
> 
> And if you want a supervisor alternative, there is https://github.com/FedericoPonzi/Horust that I got into alpine edge testing ;)

I don't really think that trimming down the containers to a single container with some kind of supervisor would be different to the consumed resources. As most of us are more likely to have CPU and RAM usage problems instead of storage problems. 

Python is a language that requires runtime. If you spawn another app (whether it's on a separate container or same container with supervisor), it will be (most likely) the same, because you'd have to have the runtime consuming some minimum CPU and RAM resource to get it running. I'd argue that it'll be better if we have snuba (and sentry ingest) to have unified consumer/ingester instead of having everything supervised.

williamdes:
> Python is a language that requires runtime. 

Yup, but using Horust, well it's Rust ;)
The consumption should be pretty minimal.

> I'd argue that it'll be better if we have snuba (and sentry ingest) to have unified consumer/ingester instead of having everything supervised.

It's maybe best not to mix everything, but maybe some tasks could be unified in one container per "topic/section"

hubertdeng123:
Re: unified snuba/ingest instances

Will have to scope that out, on the radar for Q4:
https://github.com/getsentry/team-ospo/issues/189

sentry/sentry.conf.example.py
```diff
@@ -188,6 +188,13 @@ def get_internal_network():
 
 SENTRY_DIGESTS = "sentry.digests.backends.redis.RedisBackend"
 
+###################
+# Metrics Backend #
+###################
+
+SENTRY_RELEASE_HEALTH = "sentry.release_health.metrics.MetricsReleaseHealthBackend"
```
hubertdeng123, line 195: should this live in https://github.com/getsentry/snuba/blob/master/snuba/settings/settings_self_hosted.py instead?

hubertdeng123:
Looks good to me locally
Stuff in clickhouse:
Screenshot 2023-08-30 at 10 19 53 AM

Stuff in postgres:
Screenshot 2023-08-30 at 10 18 56 AM

sentry/sentry.conf.example.py
```diff
@@ -271,6 +278,11 @@ def get_internal_network():
             "organizations:session-replay",
             "organizations:issue-platform",
             "organizations:profiling",
+            "organizations:dashboards-mep",
```
azaslavsky, line 281: Are these orgs designed to test the new snuba containers somehow?

azaslavsky (approved):

sentry/sentry.conf.example.py
hubertdeng123: those are feature flags that are set at the org level, and utilize the new snuba containers

Issue #2364, opened 2023-08-23
Shut up the clickhouse logs


williamdes:
### Problem Statement

I noticed that my clickhouse volume is way too large with "useless" logs.
In `/var/lib/docker/volumes/sentry-self-hosted_sentry-clickhouse-log/_data/`

```
2023.08.23 10:07:35.671362 [ 53524 ] {} <Information> HTTPHandler: Done processing query
2023.08.23 10:07:36.047713 [ 87 ] {3f54e9b0-34b0-4793-9488-f1d88d701e0d} <Information> executeQuery: Read 1 rows, 180.00 B in 0.014 sec., 72 rows/sec., 12.79 KiB/sec.
2023.08.23 10:07:36.054002 [ 87 ] {} <Information> HTTPHandler: Done processing query
2023.08.23 10:07:38.152329 [ 87 ] {dec81e4a-f8b0-4e75-a771-15775e4bf239} <Information> executeQuery: Read 2 rows, 261.00 B in 0.011 sec., 183 rows/sec., 23.34 KiB/sec.
2023.08.23 10:07:38.152671 [ 87 ] {} <Information> HTTPHandler: Done processing query
2023.08.23 10:07:38.672870 [ 89 ] {693acf83-4c9f-4aad-bb5c-ff074a10cc40} <Information> executeQuery: Read 4 rows, 246.00 B in 0.007 sec., 605 rows/sec., 36.39 KiB/sec.
2023.08.23 10:07:38.675604 [ 89 ] {} <Information> HTTPHandler: Done processing query
2023.08.23 10:07:38.687769 [ 53524 ] {1a022911-4e4e-4078-84ad-89e001030e58} <Information> executeQuery: Read 4 rows, 9.20 KiB in 0.020 sec., 199 rows/sec., 458.22 KiB/sec.
2023.08.23 10:07:38.688075 [ 53524 ] {} <Information> HTTPHandler: Done processing query
2023.08.23 10:07:40.358819 [ 87 ] {24968ab2-d8b3-42b1-98b3-c1d07f2d91e1} <Information> executeQuery: Read 1 rows, 180.00 B in 0.013 sec., 76 rows/sec., 13.40 KiB/sec.
2023.08.23 10:07:40.359303 [ 87 ] {} <Information> HTTPHandler: Done processing query
[...]

PR #2382, opened 2023-09-05
Reduce logs coming from clickhouse

hubertdeng123:
Fixes https://github.com/getsentry/self-hosted/issues/2364


azaslavsky (approved):

aldy505 (commented):
✅

PS. Sometimes I wonder at the irony of this being a repository for the community, yet we can't publicly press approve of request changes. IMO it should be opened though.

PR #2412, opened 2023-09-20
Parameterize backup restore script

hubertdeng123:
Allows users to call
```
./scripts/backup.sh (users/organizations/config)
./scripts/restore.sh (users/organizations/config)
```


scripts/_lib.sh
```diff
@@ -60,13 +60,15 @@ function reset() {
 }
 
 function backup() {
+  type=$1
```
aminvakil, line 63:
```suggestion
  type=${1:-"global"}
```

aminvakil (commented):
To keep it backward compatible for anyone who might use this script some another way.

ricardoapaes:
It will be of great help to us. I'm waiting for the merge!

scripts/_lib.sh
hubertdeng123: good suggestion!

azaslavsky (approved):
We should update the docs to mention this as well

hubertdeng123:
> We should update the docs to mention this as well
agreed

PR #2426, opened 2023-09-25
Add system.url-prefix to config for visibility

hubertdeng123:


BYK:
This is being asked during the initial setup though, isn't it. Even more, if it's set here people cannot change it from the UI.

hubertdeng123:
> This is being asked during the initial setup though, isn't it.

Good point, there was a user complaining that they were missing the option in their self-hosted setup. Let me follow up there to see how that happened

https://github.com/getsentry/sentry-cli/issues/1750

azaslavsky (approved):

aldy505 (commented):
✅

PR #2445, opened 2023-10-03
feat: upgrade to zookeeper-less kafka

joshuarli:
We've upgraded to zookeeper-less kafka upstream: https://github.com/getsentry/sentry/pull/55742

This simplifies the self-hosted environment and makes things leaner with less moving parts.


install/create-docker-volumes.sh
```diff
@@ -6,6 +6,5 @@ echo "Created $(docker volume create --name=sentry-kafka)."
 echo "Created $(docker volume create --name=sentry-postgres)."
 echo "Created $(docker volume create --name=sentry-redis)."
 echo "Created $(docker volume create --name=sentry-symbolicator)."
-echo "Created $(docker volume create --name=sentry-zookeeper)."
```
BYK, line 9: May wanna add clean up for this external volume on upgrades.

docker-compose.yml
```diff
@@ -156,43 +156,27 @@ services:
...
+    image: "confluentinc/cp-kafka:7.5.0"
     environment:
-      KAFKA_ZOOKEEPER_CONNECT: "zookeeper:2181"
-      KAFKA_ADVERTISED_LISTENERS: "PLAINTEXT://kafka:9092"
+      # https://docs.confluent.io/platform/current/installation/docker/config-reference.html#cp-kakfa-example
+      KAFKA_PROCESS_ROLES: "broker,controller"
+      KAFKA_CONTROLLER_QUORUM_VOTERS: "1@127.0.0.1:29093"
+      KAFKA_CONTROLLER_LISTENER_NAMES: "CONTROLLER"
+      KAFKA_NODE_ID: "1"
+      CLUSTER_ID: "MkU3OEVBNTcwNTJENDM2Qk"
```
aminvakil, line 168: Could we generate a UUID for `CLUSTER_ID` value instead of hard-coding a single string for every instance?

joshuarli: There aren't any benefits I can think of from doing so. It's just used for cluster coordination.

BYK: We can leverage this existing code if we wanted to: https://github.com/getsentry/self-hosted/blob/5ffb6a8f0382ce6d79e82dbbe671256ef9d42384/install/generate-secret-key.sh#L9

install/create-docker-volumes.sh
hubertdeng123: +1

hubertdeng123: Can probably add this to `turn-things-off.sh`

hubertdeng123 (approved):

install/create-docker-volumes.sh
hubertdeng123: pushed this up

PR #2455, opened 2023-10-06
Update nginx.conf

### LINKED ISSUE #2285: Unable to open Sentry with Safari, possibly related to Content-Encoding: gzip

mwarkentin:
Fixes #2285

Applies the same `proxy_hide_header Content-Disposition;`, but only to paths that start with `_static/` which should avoid the issue introduced previously with attachment downloads.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -76,6 +76,7 @@ http {
 			proxy_pass http://sentry;
 		}
 		location /_static/ {
+			proxy_pass http://sentry;
 			proxy_hide_header Content-Disposition;
 		}
 	}
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -75,5 +75,8 @@ http {
 		location / {
 			proxy_pass http://sentry;
 		}
+		location /_static/ {
+			proxy_hide_header Content-Disposition;
+		}
 	}
 }


nginx/nginx.conf
```diff
@@ -75,5 +75,8 @@ http {
 		location / {
 			proxy_pass http://sentry;
 		}
+		location /_static/ {
+			proxy_hide_header Content-Disposition;
```
aminvakil, line 79: This location does not inherit `proxy_pass http://sentry` from anywhere, with this change, it should have the line itself.
```suggestion
                        proxy_pass http://sentry;
			proxy_hide_header Content-Disposition;
```

mwarkentin: Thanks, good catch!

azaslavsky (approved):
Thanks for submitting this! :)

aminvakil (commented):
LGTM

PR #2491, opened 2023-10-17
configure dependabot

mdtro:
Configures dependabot to open PRs for security updates on our docker images.

I'm also including updates for our Github Actions, but if they end up too noisy -- we can re-assess.


hubertdeng123 (approved):
Thanks for putting this up :)

PR #2494, opened 2023-10-18
Remove kafka volume and remake with zookeeper removal

hubertdeng123:
Encountered this in our dogfood instance while upgrading. I needed to remove the kafka volume and remake.

### DIFF
--- a/install/turn-things-off.sh
+++ b/install/turn-things-off.sh
@@ -9,6 +9,9 @@ else
   # TODO(getsentry/self-hosted#2489)
   if docker volume ls | grep -qw sentry-zookeeper; then
     docker volume rm sentry-zookeeper
+    if docker volume ls | grep -qw sentry-kafka; then
+      docker volume rm sentry-kafka
+    fi
   fi
 fi
 


BYK:
Won't this cause data loss if there's anything yet to be processed?

hubertdeng123:
> Won't this cause data loss if there's anything yet to be processed?
Ugh ok yes, looks like the process for doing this is actually not very straightforward. I needed to do this on our dogfood instance so the instance was able to recover again.

Looks like the migration from zookeeper to KRaft is still in EA.
https://docs.confluent.io/platform/current/installation/migrate-zk-kraft.html

hubertdeng123:
I think there's a few ways moving forwards here:

1. Revert the change to remove zookeeper from self-hosted installations :disappointed:
2. Remove the sentry-kafka volume with the sentry-zookeeper removal, which will cause data loss if events have not been processed
3. Wait until there's a better process for migrating zookeeper to KRaft for self-hosted

BYK:
I don't think silently deleting possible data is a good path to upgrade. The current setup works for fresh installs so the only issue is for upgrades.

I'm pretty sure we can ensure full drainage (talk to SRE folks) and issue a warning/confirmation before deleting the volume.

hubertdeng123:
> I don't think silently deleting possible data is a good path to upgrade. The current setup works for fresh installs so the only issue is for upgrades.

I agree. I'll get back on this

hubertdeng123:
Looks like this is not possible to cutover cleanly right now, the removal of zookeeper is only done for dev env for sentry and getsentry. Looks like we'll need to hold off on removing zookeeper in self-hosted at a later date. Although it looks like that will be soon! Kafka 3.6 seems to support it, but none of the `cp-kafka` images are on it yet.

PR #2536, opened 2023-11-01
Upgrade Clickhouse to 21.8

hubertdeng123:
Upgrades clickhouse for self-hosted users. Tested it out on our own self-hosted Sentry instance and it seems to work smoothly 🤞

### DIFF
--- a/install/detect-platform.sh
+++ b/install/detect-platform.sh
@@ -20,10 +20,10 @@ fi
 export DOCKER_ARCH=$(docker info --format '{{.Architecture}}')
 if [[ "$DOCKER_ARCH" = "x86_64" ]]; then
   export DOCKER_PLATFORM="linux/amd64"
-  export CLICKHOUSE_IMAGE="yandex/clickhouse-server:20.3.9.70"
+  export CLICKHOUSE_IMAGE="yandex/clickhouse-server:21.8.13.1.altinitystable"
 elif [[ "$DOCKER_ARCH" = "aarch64" ]]; then
   export DOCKER_PLATFORM="linux/arm64"
-  export CLICKHOUSE_IMAGE="altinity/clickhouse-server:21.6.1.6734-testing-arm"
+  export CLICKHOUSE_IMAGE="altinity/clickhouse-server:21.8.12.29.altinitydev.arm"
 else
   echo "FAIL: Unsupported docker architecture $DOCKER_ARCH."
   exit 1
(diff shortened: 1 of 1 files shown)


lynnagara:
can you explain the `21.8.13.1.altinitystable` vs `21.8.12.29.altinitydev.arm` part? i'm wondering if this is also a solution we can use to upgrade https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L2753

hubertdeng123:
Sure, `21.8.12.29.altinitydev.arm`, is an image that is used for dev environments primarily those running on Apple silicon. So I think you can use this to upgrade sentry dev. There sadly doesn't seem to be an arm image for 21.8.13.1 though, so the versions won't be completely aligned.

hubertdeng123:
tested in dev, it's working for me

lynnagara (approved):

lynnagara:
> tested in dev, it's working for me

Thanks, will update to this version in sentry then

lynnagara:
Should we be using the `altinitydev.arm` versions? Would `altinity/clickhouse-server:21.8.12.1.testingarm` be better?

hubertdeng123:
> Should we be using the altinitydev.arm versions? Would altinity/clickhouse-server:21.8.12.1.testingarm be better?

Don't think there is much of a difference there as I couldn't find much online. It's newer and the version of clickhouse used is more close to the .testingarm version. Seems fine to use and I've tested it locally

azaslavsky (approved):

lynnagara:
> > Should we be using the altinitydev.arm versions? Would altinity/clickhouse-server:21.8.12.1.testingarm be better?
> 
> Don't think there is much of a difference there as I couldn't find much online. It's newer and the version of clickhouse used is more close to the .testingarm version. Seems fine to use and I've tested it locally

@nikhars Do you know which we should be using here? It was my understanding that the testing versions are generally more stable than the dev ones but I couldn't find details about it online.

williamdes:
Does anyone has a reference of why the name changes that would help me trust the new image rather than the Yandex namespace?
Can I just pick the same version on the Yandex repo?

MeredithAnya:
> Does anyone has a reference of why the name changes that would help me trust the new image rather than the Yandex namespace? Can I just pick the same version on the Yandex repo?

The Altinity versions are maintained for 3 years, we use Altinity builds in Sentry's production clusters so that we can have more stability when it comes to upgrading our clusters. According to their docs, they are 100% compatible with the community builds: https://docs.altinity.com/altinitystablebuilds/

williamdes:
Thank you!

Cybernisk:
Hi @hubertdeng123 should this fix the issue with Replays and Clickhouse mentioned by @cmanallen in https://github.com/getsentry/sentry/issues/57326#issuecomment-1785229083 ?
I've moved to `altinity/clickhouse-server:21.8.15.15.altinitystable` as Allen advised.

hubertdeng123:
@Cybernisk Yes, this should fix the issue that you have.

PR #2565, opened 2023-11-13
Enable Cron monitoring in self-hosted version

mohsenmottaghi:
Sentry has had a Cron Monitoring feature in beta for a long time. Since Sentry is open-source, we can develop applications in our development with the self-hosted version before shipping them to the production environment without being worried about false alarms and spamming events. So, if it's possible, we would like to see this feature in the self-hosted too.  

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -368,6 +368,9 @@ services:
   ingest-profiles:
     <<: *sentry_defaults
     command: run consumer --no-strict-offset-reset ingest-profiles --consumer-group ingest-profiles
+  ingest-monitors:
+    <<: *sentry_defaults
+    command: run consumer --no-strict-offset-reset ingest-monitors --consumer-group ingest-monitors
   post-process-forwarder-errors:
     <<: *sentry_defaults
     command: run consumer post-process-forwarder-errors --consumer-group post-process-forwarder --synchronize-commit-log-topic=snuba-commit-log --synchronize-commit-group=snuba-consumers
--- a/install/create-kafka-topics.sh
+++ b/install/create-kafka-topics.sh
@@ -3,7 +3,7 @@ echo "${_group}Creating additional Kafka topics ..."
 # NOTE: This step relies on `kafka` being available from the previous `snuba-api bootstrap` step
 # XXX(BYK): We cannot use auto.create.topics as Confluence and Apache hates it now (and makes it very hard to enable)
 EXISTING_KAFKA_TOPICS=$($dcr -T kafka kafka-topics --list --bootstrap-server kafka:9092 2>/dev/null)
-NEEDED_KAFKA_TOPICS="ingest-attachments ingest-transactions ingest-events ingest-replay-recordings profiles ingest-occurrences ingest-metrics ingest-performance-metrics"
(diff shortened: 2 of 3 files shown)


hubertdeng123:
This is still in beta, we will put a PR to enable this as soon as it gets to GA. For the time being we encourage you to test out the Crons feature but we're not ready to enable this for everyone just yet.

huangdijia:
Still looking forward to joining

alireza1371:
This is what we need also in our instance, dig into it and find out it almost a year that this feature has existed, so I'm gonna thankful if you guys reconsider the releasing of this feature sooner.

BYK:
Folks, if you need this, it shouldn't be too hard to apply these changes to your local copy. One is in the config file anyway so that's not even changing the repo.

The other one is in the docker-compose file which you can amend with a `docker-compose.overrides.yml` without changing the repo.

Then the only step left is creating the Kafka topic manually and you should be good to go.

huangdijia:
> Folks, if you need this, it shouldn't be too hard to apply these changes to your local copy. One is in the config file anyway so that's not even changing the repo.
> 
> 
> 
> The other one is in the docker-compose file which you can amend with a `docker-compose.overrides.yml` without changing the repo.
> 
> 
> 
> Then the only step left is creating the Kafka topic manually and you should be good to go.

I tried to adjust the docker-compose according to the PR, but it doesn't work. Can you give me some guidance?

BYK:
>Can you give me some guidance?

Not without you sharing what went wrong 😅

mohsenmottaghi:
> > Can you give me some guidance?
> 
> Not without you sharing what went wrong 😅

Fair enough,
It's been quite some time since this feature was released in beta. Can you please provide us with an estimate of when this feature will be fully available? We need to make a decision as to whether we should wait longer or look for another solution.

BYK:
@mohsenmottaghi - I am not working for Sentry so I cannot provide you anything like that.

If you tried to direct this to @hubertdeng123 - my 2 cents would still be "cannot give you anything solid" as this is an open source project with no guarantees around anything. If you want guarantees, a real service etc. you should consider the SaaS version of Sentry.

hubertdeng123:
@mohsenmottaghi I cannot give you a good estimate as dates have been moving around internally as well. @BYK has summed it up perfectly above

ckdake:
@hubertdeng123 This confused us as we're using sentry local and have crons visible in the UI and can create crons in the UI, but without the background workers doing their thing.

cron checkins with curl/sentry-cli/go  respond with a HTTP 202 and look like they are successful, but never make it to the UI. I was diagnosing why and ran into this PR.

The UI shouldn't be visible and/or work for crons if it's not going to actually be enabled in the opensource codebase.

hubertdeng123:
@ckdake The UI should not be visible for Crons in self-hosted by default, have you messed around with your `sentry.conf.py` file?

ckdake:
@hubertdeng123 Yes, we enabled the `organizations:monitors` in SENTRY_FEATURES, expecting that to turn the feature on.

hubertdeng123:
Then, it seems like things are performing as expected. This feature is not enabled by default, and the UI is not visible as it is still in beta. If you'd like to give it a shot, you'll need to add a kafka topic and an ingest consumer

PR #2592, opened 2023-11-21
Patch for dev self-hosted environments

hubertdeng123:
Recently this error started appearing in dev environments on Apple silicon chips. 
```
▶ Setting up / migrating database ...
 smtp Pulling
no matching manifest for linux/arm64/v8 in the manifest list entries
Error in install/set-up-and-migrate-database.sh:13.
'$dcr web upgrade' exited with status 18
```

This is a workaround for that

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,6 +102,7 @@ x-snuba-defaults: &snuba_defaults
 services:
   smtp:
     <<: *restart_policy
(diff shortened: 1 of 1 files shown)


azaslavsky (approved):

aminvakil (commented):
LGTM

PR #2594, opened 2023-11-21
feat: Add sentry-admin.sh tool

azaslavsky:
`sentry-admin.sh` is just a bash wrapper around some admin commands that are already technically exposed in the `web` container, though a bit hard to find today. This command just forwards the `sentry` command-line administration tool out from the container onto the host machine, slightly editing the help docs to make it a bit easier to use in the process.

### DIFF
--- a/_integration-test/ensure-sentry-admin-works.sh
+++ b/_integration-test/ensure-sentry-admin-works.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+set -ex
+
+source install/_lib.sh
+source install/dc-detect-version.sh
+
+echo "${_group}Test that sentry-admin works..."
+
+echo "Global help documentation..."
+
+global_help_doc=$(/bin/bash --help)
+if ! echo "$global_help_doc" | grep -q "^Usage: ./sentry-admin.sh"; then
+  echo "Assertion failed: Incorrect binary name in global help docs"
+  exit 1
+fi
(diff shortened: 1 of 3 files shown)


_integration-test/ensure-sentry-admin-works.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
hubertdeng123, line 1: I don't think we should have two instances of backup in integration tests if it takes too long. It already takes over 25 minutes to run integration tests and this adds to that overhead

sentry-admin.sh
hubertdeng123, line 1: If we have this sentry-admin tool, I think we can delete the `/scripts/backup.sh` and `/scripts/restore.sh`, thoughts?

hubertdeng123: Along the lines of that, I do think `reset.sh` could be incorporated here, but that can be done in a follow up PR.

azaslavsky: I'd like to keep them for now, since users may be relying on them. We can deprecate-then-delete them in future releases.

Agree on `reset.sh`.

_integration-test/ensure-sentry-admin-works.sh
azaslavsky: Removed this 2nd backup test, and modified the original to use `sentry-admin.sh` instead. We'll modify the docs once this PR lands.

hubertdeng123 (approved):

Issue #2251, opened 2023-07-04
Sentry responded with an error: HTTPSConnectionPool


rajivmucheli:
### Self-Hosted Version

Self hosted 9.1.4

### CPU Architecture

x86_64

### Docker Version

20.10.17

### Docker Compose Version

k8s

### Steps to Reproduce

Hi,

I see the below Sentry events in sentry logs, how can i fix it ?

b"Sentry responded with an error: HTTPSConnectionPool(host='', port=443): Read timed out. (read timeout=5) (url: https://sentry.eu-de-1.cloud.sap/api/4/store/)"

2023-07-04 04:45:33,138 24 ERROR sentry.errors [req-4c00dcd6-4bdf-4a8e-864d-9333179bff7f gNone - - - - -] Sentry responded with an error: HTTPSConnectionPool(host='xxx', port=443): Read timed out. (read timeout=5) (url: https://xxxx/api/22/store/): requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='xxx', port=443): Read timed out. (read timeout=5)

1. I looked into https://github.com/getsentry/self-hosted/issues/350 but had no luck.
2. i set socket.setdefaulttimeout(10) in sentry.conf

### Expected Result

How to fix the connectionpool error.

### Actual Result

b"Sentry responded with an error: HTTPSConnectionPool(host='', port=443): Read timed out. (read timeout=5) (url: https://sentry.eu-de-1.cloud.sap/api/4/store/)"

2023-07-04 04:45:33,138 24 ERROR sentry.errors [req-4c00dcd6-4bdf-4a8e-864d-9333179bff7f gNone - - - - -] Sentry responded with an error: HTTPSConnectionPool(host='xxx', port=443): Read timed out. (read timeout=5) (url: https://xxxx/api/22/store/): requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='xxx', port=443): Read timed out. (read timeout=5)

azaslavsky:
Hi! We only offer limited support for non-standard setups, which seems to be the case here due to the use of kubernetes.

[...]

PR #2606, opened 2023-11-27
fix: set postgres shm_size to 256mb

jboutsicaris:
update shared memory size for postrgres container to allow vacuum operations: https://github.com/docker-library/docs/blob/master/postgres/README.md#caveats

Capture_could-not-resize-shared-memory-segment

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -149,6 +149,7 @@ services:
     environment:
       POSTGRES_HOST_AUTH_METHOD: "trust"
     entrypoint: /opt/sentry/postgres-entrypoint.sh
+    shm_size: "256mb"
     volumes:
       - "sentry-postgres:/var/lib/postgresql/data"
       - type: bind


BYK:
Not sure if this is widely applicable, especially after reading this comment: https://dba.stackexchange.com/questions/275378/dev-shm-size-recommendation-for-postgres-database-in-docker#comment634037_275385

Why not make this specific to your instance with a `docker-compose.overrides.yml` file?

jboutsicaris-OD:
Capture_sentry-cleanup-connection-error

The main reason I was vacuuming is because I wasnt sure about the successful exit of the sentry cleanup command. I initiated this command with `docker compose exec -it sentry-cleanup bash` and ran `sentry cleanup --days 30`. It would be nice to probably stick to using the sentry commands. Unfortunately I don't think the exit code is set on failure so I resorted to psql interface to make sure I had enough room. I also see this potentially related issue: https://github.com/getsentry/self-hosted/issues/2251

I'd rather use sentry cleanup than modify posgtres directly. Thanks.

hubertdeng123:
Closing this as it is not a general use case for most users.

jboutsicaris-OD:
@hubertdeng123 Do you think I should make a GitHub issue for the cleanup error I posted here? I don't see that particular error there yet. Thanks

PR #2649, opened 2023-12-13
ref: get self-hosted sentry image from ghcr instead of dockerhub

asottile-sentry:

### DIFF
--- a/.env
+++ b/.env
@@ -5,7 +5,7 @@ SENTRY_EVENT_RETENTION_DAYS=90
 SENTRY_BIND=9000
 # Set SENTRY_MAIL_HOST to a valid FQDN (host/domain name) to be able to send emails!
 # SENTRY_MAIL_HOST=example.com
-SENTRY_IMAGE=getsentry/sentry:nightly
+SENTRY_IMAGE=ghcr.io/getsentry/sentry-self-hosted:latest
 SNUBA_IMAGE=getsentry/snuba:nightly
 RELAY_IMAGE=getsentry/relay:nightly
 SYMBOLICATOR_IMAGE=getsentry/symbolicator:nightly


hubertdeng123:
I'm guessing something similar to https://github.com/getsentry/sentry/pull/61434 can be applied to snuba, relay, and symbolicator too? It would be good to add a todo for getting those onto ghcr for nightly/latest images

Also, I believe this will break the release scripts here:
https://github.com/getsentry/self-hosted/blob/master/scripts/bump-version.sh

asottile-sentry:
> I'm guessing something similar to [getsentry/sentry#61434](https://github.com/getsentry/sentry/pull/61434) can be applied to snuba, relay, and symbolicator too? It would be good to add a todo for getting those onto ghcr for nightly/latest images
> 
> Also, I believe this will break the release scripts here: https://github.com/getsentry/self-hosted/blob/master/scripts/bump-version.sh

that script seems like it should still work?  I think it'll replace `:latest` with `:whatever` -- unless I'm reading it wrong...

asottile-sentry:
> > I'm guessing something similar to [getsentry/sentry#61434](https://github.com/getsentry/sentry/pull/61434) can be applied to snuba, relay, and symbolicator too? It would be good to add a todo for getting those onto ghcr for nightly/latest images
> > Also, I believe this will break the release scripts here: https://github.com/getsentry/self-hosted/blob/master/scripts/bump-version.sh
> 
> that script seems like it should still work? I think it'll replace `:latest` with `:whatever` -- unless I'm reading it wrong...

oh except we don't have the versioned tagged images in `ghcr` -- what do

asottile-sentry:
alright validated the new image works -- will have to finish this some other time

PR #2670, opened 2023-12-22
Tweak the template now that we can see it

chadwhitacre:
Screenshot 2023-12-22 at 11 48 25 AM

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.yml
+++ b/.github/ISSUE_TEMPLATE/release.yml
@@ -7,7 +7,7 @@ body:
       label: Body
       description: "Edit YY.M.N in the title and three times in the first line of the body, then submit. 👍"
       value: |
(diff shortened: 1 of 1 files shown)


BYK:
You can probably tie this to a scheduled GitHub action and remove all manual editing. Just saying 🙂

chadwhitacre:
We could schedule the creation of the monthly release issue since that is known and can be planned for. We do of course have out-of-band releases for which it is still valuable to have this template. The scheduled issue creation can use this template as well so I don't think this is a bad incremental step. Not sure what you mean by manual editing, exactly.

BYK:
>Edit YY.M.N in the title and three times in the first line of the body, then submit. 👍

This is the manual editing I was referring to.

Regarding patch releases, I'm fairly sure there was also some good auto versioning script somewhere in the release pipeline. So the process can be "go run this workflow then follow the instructions on the issue it assigned to you"

hubertdeng123 (approved):

PR #2722, opened 2024-01-18
Fix cache error self hosted

hubertdeng123:


chadwhitacre (approved):

hubertdeng123:
Looks like `server_max_value_length` is not available for PyMemcacheCache

https://docs.djangoproject.com/en/5.0/topics/cache/#memcached

bobvandevijver:
Looks like this change broke our installations, see https://github.com/getsentry/self-hosted/issues/2728#issuecomment-1913081251 for the log.

PR #2732, opened 2024-01-23
simplify if for open-ai-suggestion

LvckyAPI:
I simplified the if request at the `sentry.conf.example.py`.

### Legal Boilerplate

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -317,7 +317,6 @@ GEOIP_PATH_MMDB = '/geoip/GeoLite2-City.mmdb'
 # OpenAI API key to turn on the feature.
 OPENAI_API_KEY = env("OPENAI_API_KEY", "")
 
-OPENAI_API_KEY = env("OPENAI_API_KEY", "")
 SENTRY_FEATURES["organizations:open-ai-suggestion"] = bool(OPENAI_API_KEY)
 
 ##############################################
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -317,8 +317,8 @@ GEOIP_PATH_MMDB = '/geoip/GeoLite2-City.mmdb'
 # OpenAI API key to turn on the feature.
 OPENAI_API_KEY = env("OPENAI_API_KEY", "")
 
-if OPENAI_API_KEY:
-  SENTRY_FEATURES["organizations:open-ai-suggestion"] = True
+OPENAI_API_KEY = env("OPENAI_API_KEY", "")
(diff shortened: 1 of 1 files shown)


sentry/sentry.conf.example.py
```diff
@@ -317,8 +317,8 @@ def get_internal_network():
 # OpenAI API key to turn on the feature.
 OPENAI_API_KEY = env("OPENAI_API_KEY", "")
 
-if OPENAI_API_KEY:
-  SENTRY_FEATURES["organizations:open-ai-suggestion"] = True
+OPENAI_API_KEY = env("OPENAI_API_KEY", "")
```
LvckyAPI, line 320:
```suggestion
```

Tildashilds (commented):
Me

sentry/sentry.conf.example.py
```diff
@@ -317,8 +317,7 @@ def get_internal_network():
 # OpenAI API key to turn on the feature.
 OPENAI_API_KEY = env("OPENAI_API_KEY", "")
 
-if OPENAI_API_KEY:
-  SENTRY_FEATURES["organizations:open-ai-suggestion"] = True
+SENTRY_FEATURES["organizations:open-ai-suggestion"] = bool(OPENAI_API_KEY)
```
aldy505, line 320: I'd prefer this to be just:
```python
SENTRY_FEATURES["organizations:open-ai-suggestion"] = False
```
So users will have to be aware that they're toggling the OpenAI feature. No billing surprise because of their Sentry deployments.

hubertdeng123: But would you argue that only users who would want the OpenAI feature would be setting the `OPENAI_API_KEY`?

aldy505: @hubertdeng123 Pretty much yeah, OpenAI billing is so scary when you're forget that you have to pay for them

sentry/sentry.conf.example.py
```diff
@@ -317,8 +317,7 @@ def get_internal_network():
 # OpenAI API key to turn on the feature.
 OPENAI_API_KEY = env("OPENAI_API_KEY", "")
```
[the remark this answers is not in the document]
aldy505: I'm not that good with python, but at least it's similar from what's implemented on the upstream sentry.

https://github.com/getsentry/sentry/blob/76906375a34c8fd566d20264db6fa87ad50f70d2/src/sentry/api/endpoints/event_ai_suggested_fix.py#L320-L322

hubertdeng123 (approved):

Issue #2725, opened 2024-01-22
Question about installation with http_proxy


lemrouch:
### Problem Statement

I'm trying to install Sentry behind firewall.
Code in https://github.com/getsentry/self-hosted/pull/1543/files is
```
RUN if [ -z "${http_proxy}" ]; then echo "Acquire::http::proxy \"${http_proxy}\";" >> /etc/apt/apt.conf; fi
RUN if [ -z "${https_proxy}" ]; then echo "Acquire::https::proxy \"${https_proxy}\";" >> /etc/apt/apt.conf; fi
```
In my experience and documentation `-z` in test means _if the string is empty_ therefore the config won't be set if there are actual values in http_proxy and/or https_proxy variables.

### Solution Brainstorm

I guess the code was meant to be
```
RUN if [ ! -z "${http_proxy}" ]; then echo "Acquire::http::proxy \"${http_proxy}\";" >> /etc/apt/apt.conf; fi
RUN if [ ! -z "${https_proxy}" ]; then echo "Acquire::https::proxy \"${https_proxy}\";" >> /etc/apt/apt.conf; fi
```
or
```
RUN if [ -n "${http_proxy}" ]; then echo "Acquire::http::proxy \"${http_proxy}\";" >> /etc/apt/apt.conf; fi
RUN if [ -n "${https_proxy}" ]; then echo "Acquire::https::proxy \"${https_proxy}\";" >> /etc/apt/apt.conf; fi
```
Am I right or too tired?

azaslavsky:
I think that's correct. Any interest in submitting a PR with the fix?

PR #1543, opened 2022-06-26
Improve installation through proxy

goganchic:
Fixes #1537

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #2734, opened 2024-01-24
Fix apt config generation when http_proxy is set (#2725)

lemrouch:
Fix inverted logic for http_proxy condition in cron container configuration.
Closes https://github.com/getsentry/self-hosted/issues/2725

### Legal Boilerplate

### DIFF
--- a/cron/Dockerfile
+++ b/cron/Dockerfile
@@ -1,8 +1,8 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 USER 0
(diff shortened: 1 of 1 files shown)


aminvakil (commented):
LGTM

@goganchic Since you've added this in https://github.com/getsentry/self-hosted/pull/1543, did it work for you afterwards?

aldy505 (approved):

hubertdeng123 (approved):

PR #2749, opened 2024-01-30
[FEAT]: Scaler/update secret key

ymeghna:
1. Added script for adding private github key script 
2. Add Private github key in yml

### Legal Boilerplate


hubertdeng123:
Closing as this does not seem like it belongs here

Issue #2698, opened 2024-01-10
Add Developer Metrics


aldy505:
### Problem Statement

Sentry has Developer Metrics in alpha. We should bring that into self-hosted after GA.

### Solution Brainstorm

Discussion: https://github.com/getsentry/sentry/discussions/58584
Feature docs: https://develop.sentry.dev/delightful-developer-metrics/
SDK: https://develop.sentry.dev/sdk/metrics/

SDK guide on Python: https://develop.sentry.dev/delightful-developer-metrics/sending-metrics-sdk/

aldy505:
To enable the feature, you'd need to edit your `sentry/sentry.conf.py` file and add these feature flags:

```python
SENTRY_FEATURES.update(
  {
    feature: True
    for feature in (
      # ......
      # For DDM
      "organizations:ddm-ui",
      "organizations:custom-metrics",
      "organizations:ddm-experimental",
      "organizations:use-metrics-layer-in-alerts"
    )
  }
)
```

Add new Snuba container on your `docker-compose.yml`:

```yaml
  snuba-generic-metrics-gauges-consumer:
    <<: *snuba_defaults
    command: consumer --storage generic_metrics_gauges_raw --consumer-group snuba-gen-metrics-gauges-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
```

 Then do `./install.sh` all over again.

hubertdeng123:
Thanks for creating this issue

DarkByteZero:
These are required Database Migrations:

```sh
docker compose run --rm snuba-api migrations migrate -g spans
```

aldy505:
Hello @ayirr7, since GA for Metrics is near, do you want to re-create this PR https://github.com/getsentry/self-hosted/pull/2757? (You can't reopen since it's already locked, you need to create a new PR)

I believe adding the gauge consumer now is a good idea, since it will ingest some data beforehand, that can be viewed later on the UI when GA's arrived -- so the Metrics UI won't be empty when it's enabled.

marianhlavac:
> To enable the feature, make sure you're on **23.11.2** or higher.

Note for anyone trying to test this on their self-hosted instance, the minimum version is probably actually higher. Had no luck on 24.1.1 (the Metrics page was not showing at all), after upgrading to **24.7.0** it worked well.

xuweixi11:
anyone used the metric alert success?  it's never triggered, my version is 24.7.0

[...]

Issue #2646, opened 2023-12-12
Support new Beta User Feedback feature in self-hosted


bruno-garcia:
New User Feedback workflow and JS widget to capture feedback outside of runtime errors (event_id link).

This will allow us have the feature on the Sandbox. And single tenant.

Feature flags:

```
"organizations:user-feedback-ui",
"organizations:user-feedback-ingest",
"organizations:issue-platform",
"organizations:feedback-ingest",
"organizations:feedback-post-process-group",
"organizations:feedback-visible",
```

## Documentation

Product: https://docs.sentry.io/product/user-feedback/
SDK: https://docs.sentry.io/platforms/javascript/user-feedback/

aldy505:
Mandatory question: Do we need additional Sentry worker or Snuba consumer containers?

hubertdeng123:
@bruno-garcia do you know the answer to this? I peeked into `sentry` and it doesn't look like any additional infra should be required but not entirely sure

SebastiaanKloos:
I found nothing more than a feature flag that should be enabled. Weird thing is that it did not ingest any feedbacks.

holtgrewe:
I can confirm, that after adding the feature flags, I can see the UI but nothing is ingested.

JoshFerge:
Hello, please also add these features:

```
"organizations:issue-platform",
"organizations:feedback-ingest",
"organizations:feedback-post-process-group",
"organizations:feedback-visible",
```

let me know if this does not work!

SebastiaanKloos:
> Hello, please also add these features:
> 
> ```
> "organizations:issue-platform",
> "organizations:feedback-ingest",
> "organizations:feedback-post-process-group",
> "organizations:feedback-visible",
> ```
> 
> let me know if this does not work!

Thank you so much. This fixed it. Woohoo!

![CleanShot 2024-01-02 at 23 14 38](https://github.com/getsentry/self-hosted/assets/6506510/e24bf201-cc6b-4886-b42d-89523c777ae6)

schettn:
Anyone getting this after adding the features flags?

![image](https://github.com/getsentry/self-hosted/assets/52858351/c3de7b37-cd36-4f11-b29f-ea50ff0af56c)

mlch911:
With the addition of these 6 flags, I can now see the user feedback tab, but it can't receive any data. The user feedback tab on the details page of a single issue also has no data.
I'm using sentry 24.1.0 right now.

[...]

Issue #1993, opened 2023-02-28
Add Crons


hubertdeng123:
### Problem Statement

Sentry has a Cron Monitoring feature in beta. We'd like to make this available to everyone. 

### Solution Brainstorm

Once Crons goes to GA, we'll make Crons generally available for self-hosted

See Replays in #1990 for a template.

hubertdeng123:
For now, users can enable it by editing their `sentry.conf.py` file. In `SENTRY_FEATURES`, add the `"organizations:monitors"` flag. This will look something like this:

```
SENTRY_FEATURES.update(
    {
        feature: True
        for feature in (
            "organizations:discover",
            ...
            "organizations:monitors",
        )
    }
)
```

chadwhitacre:
@hubertdeng123 Can you post this back in that slack thread wherever it was if you haven't?

hubertdeng123:
already posted there 😃

viktor2097:
> For now, users can enable it by editing their `sentry.conf.py` file. In `SENTRY_FEATURES`, add the `"organizations:monitors"` flag. This will look something like this:
> 
> ```
> SENTRY_FEATURES.update(
>     {
>         feature: True
>         for feature in (
>             "organizations:discover",
>             ...
>             "organizations:monitors",
>         )
>     }
> )
> ```

Do we have an equivalent to turn on the new profiling feature?

chadwhitacre:
> Do we have an equivalent to turn on the new profiling feature?

If so let's post it on https://github.com/getsentry/self-hosted/issues/1838.

aldy505:
> Do we have an equivalent to turn on the new profiling feature?

@viktor2097 Not yet. At least from my understanding, profiling feature requires [additional backend](https://github.com/getsentry/vroom) to be added to the docker compose file and set the profiling backend address on the `sentry` container. Right now, the `vroom` is not published on the Docker Hub yet, so we're all waiting on that.

aldy505:
@hubertdeng123 any idea when this will be on GA?

aldy505:
For those brave souls who wants to activate Cron feature:

On `sentry/sentry.conf.py`:
```python
# Add this line
SENTRY_FEATURES["organizations:monitors"] = True
```

On `docker-compose.yml`: (yes you have to deal with changing this everytime you upgrade Sentry)
```yaml
services:
  # Add new container for Cron
  ingest-monitors:
    <<: *sentry_defaults
[...]

PR #2757, opened 2024-02-01
feat(generic-metrics): Add gauges to docker compose file

ayirr7:
This was missing by default, adding this consumer to the docker-compose file as well

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -173,8 +173,7 @@ services:
       - "sentry-secrets:/etc/zookeeper/secrets"
     healthcheck:
       <<: *healthcheck_defaults
-      test:
-        ["CMD-SHELL", 'echo "ruok" | nc -w 2 localhost 2181 | grep imok']
+      test: ["CMD-SHELL", 'echo "ruok" | nc -w 2 localhost 2181 | grep imok']
   kafka:
     <<: *restart_policy
     depends_on:
@@ -203,7 +202,11 @@ services:
       - "sentry-secrets:/etc/kafka/secrets"
     healthcheck:
       <<: *healthcheck_defaults
-      test: ["CMD-SHELL", "/usr/bin/kafka-topics --bootstrap-server kafka:9092 --list"]
+      test:
+        [
+          "CMD-SHELL",
+          "/usr/bin/kafka-topics --bootstrap-server kafka:9092 --list",
+        ]
       interval: 10s
       timeout: 10s
       retries: 30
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -283,6 +287,9 @@ services:
   snuba-generic-metrics-counters-consumer:
     <<: *snuba_defaults
     command: consumer --storage generic_metrics_counters_raw --consumer-group snuba-gen-metrics-counters-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
+  snuba-generic-metrics-gauges-consumer:
```
hubertdeng123, line 290: is this a new consumer that was missed? 😅

hubertdeng123 (approved):

aldy505 (commented):
I know that this is missing and as far as I know, this is only used by the DDM backend? If that's the case, can we hold this until DDM is at least in beta or GA?

Adding a new container to the compose file is... an undesirable thing to do for self-hosted Sentry.

hubertdeng123:
If this is only used for DDM, I agree with @aldy505 that we should hold off on this for now

hubertdeng123 (commented):
If this is only used for DDM, I agree with @aldy505 that we should hold off on this for now

ale-cota:
Since a lot of people are bound to run into this issue while trying to run metrics on self-hosted, would it be possible/acceptable to add it as a comment at least? So as to inform users that if they want to use DDM, they need to add the additional consumer to their docker-compose file.

aldy505:
> Since a lot of people are bound to run into this issue while trying to run metrics on self-hosted, would it be possible/acceptable to add it as a comment at least? So as to inform users that if they want to use DDM, they need to add the additional consumer to their docker-compose file.

@ale-cota The self-hosted community never do that. For new features we always put an issue and point people to run/enable some things on their config in order to enable a new feature. For this specific issue, it's here: https://github.com/getsentry/self-hosted/issues/2698

For other features, see https://github.com/getsentry/self-hosted/issues/2646, https://github.com/getsentry/self-hosted/issues/1993#issuecomment-1675633779, and.. I forgot where is the issue for Performance Queries

chadwhitacre:
> to inform users that if they want to use DDM, they need to add the additional consumer to their docker-compose file

I like using a GitHub issue for this. Let's start formalizing that as our process. I've added [a `Type: Pre-release Feature` label](https://github.com/getsentry/self-hosted/labels/Type%3A%20Pre-release%20Feature) to this repo and applied it to https://github.com/getsentry/self-hosted/issues/2698. I think it'd be great to have a paragraph on our [self-hosted docs](https://develop.sentry.dev/self-hosted/) about enabling pre-release features and links out to the issue label listing for current pre-release features. We have some reference under [feature flags](https://develop.sentry.dev/feature-flags/) but nothing specific under [self-hosted](https://develop.sentry.dev/self-hosted/) afaict.

@ale Can you help us share https://github.com/getsentry/self-hosted/issues/2698 with customers who are interested trying out Metrics in self-hosted?

@ayirr7 I'm gonna close this PR. 🙏

Issue #2728, opened 2024-01-23
Python: ModuleNotFoundError: No module named 'memcache'


NickMous:
### Self-Hosted Version

24.1.0

### CPU Architecture

x86_64

### Docker Version

25.0.0

### Docker Compose Version

2.24.1

### Steps to Reproduce

1. Have sentry self-hosted previous version installed
2. do git pull
3. run ./install.sh

### Expected Result

A working sentry

### Actual Result

![Screenshot_20240123_095114](https://github.com/getsentry/self-hosted/assets/61657558/a7e93cd0-b84c-4737-b557-b2bfdc9c07d0)
![Screenshot_20240123_095302](https://github.com/getsentry/self-hosted/assets/61657558/4905fe9c-3ecb-4237-8774-31d5a6221ea1)
While upgrading it already stated that it couldn't do anything with redis, because of memcached. This was only a warning, so it went through. But now it also shows up in the logs of the sentry web runner, and I get a 502 bad gateway.

lukasleitsch:
A workaround until the issue is fixed in the nightly image.

```
SENTRY_IMAGE=getsentry/sentry:24.1.0 ./install.sh
```

NickMous:
Thanks! I switched to the 24.1.0 label for the time being :)

aldy505:
Hi @NickMous can you confirm that the solution on this works? https://github.com/getsentry/self-hosted/pull/2722

NickMous:
Hi @aldy505 ! I unfortunately was sick, but now I've taken the time for testing this, and this is what I now have:
When pulling the latest from master there are errors still remaining about memcache, but now mainly `AttributeError: module 'django.core.cache.backends.memcached' has no attribute `'MemcachedCache'`.
So far with replacing the image version from nightly to 24.1.0 I get another error that the app_bootstrap_initializeMain is not found. I'll try to rerun the install.sh, I'll get back in a sec

NickMous:
For some reason running without the tag to give y'all insight for errors and manually saying yes, the install worked now and it's up and running. So it has not been fixed yet on master/nightly, but master/SENTRY_IMAGE=getsentry/sentry:24.1.0 works

bobvandevijver:
I believe I hit this issue when upgrading from 24.1.0 to 24.1.1. 

```
[...]

PR #2778, opened 2024-02-08
Adapt to new memcached backend in Django

chadwhitacre:
Fixes https://github.com/getsentry/self-hosted/issues/2728.

### DIFF
--- a/install/check-memcached-backend.sh
+++ b/install/check-memcached-backend.sh
@@ -1,0 +1,16 @@
+echo "${_group}Checking memcached backend ..."
+
+if grep -q "\.PyMemcacheCache" "$SENTRY_CONFIG_PY"; then
+  echo "PyMemcacheCache found in $SENTRY_CONFIG_PY, gonna assume you're good."
+else
+  if grep -q "\.MemcachedCache" "$SENTRY_CONFIG_PY"; then
+    echo "MemcachedCache found in $SENTRY_CONFIG_PY, you should switch to PyMemcacheCache."
+    echo "See:"
+    echo "  https://develop.sentry.dev/self-hosted/releases/#breaking-changes"
+    exit 1
+  else
+    echo 'Your setup looks weird. Good luck.'
+  fi
+fi
+
+echo "${_endgroup}"
--- a/install.sh
+++ b/install.sh
@@ -23,6 +23,7 @@ source install/check-minimum-requirements.sh
 source install/turn-things-off.sh
 source install/create-docker-volumes.sh
 source install/ensure-files-from-examples.sh
+source install/check-memcached-backend.sh
 source install/ensure-relay-credentials.sh
 source install/generate-secret-key.sh
 source install/update-docker-images.sh


install/check-memcached-backend.sh
```diff
@@ -0,0 +1,17 @@
+echo "${_group}Checking memcached backend ..."
+
+if grep -q "\.PyMemcacheCache" "$SENTRY_CONFIG_PY"; then
+  echo "found PyMemcacheCache in $SENTRY_CONFIG_PY, looks good"
+else
+  if grep -q "\.MemcachedCache" "$SENTRY_CONFIG_PY"; then
```
azaslavsky, line 6: Wonder if in this case, we can't rename their existing `CACHES` variable to `CACHES_BAK` and print a message that we did so? This keeps from overwriting their settings wholesale, which might be quite different from what our recommended defaults were, and are not otherwise backed up. In particular, I'm worried about how any changes the user might have made to `OPTIONS` will interact with our change, since the Django Memcached and PyMemCached have very different options APIs.

install/check-memcached-backend.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "found MemcachedCache in $SENTRY_CONFIG_PY, switching to PyMemcacheCache"
+    sed -i '' -e 's/\.MemcachedCache/.PyMemcacheCache/' "$SENTRY_CONFIG_PY"
+    sed -i '' -e 's/^\( *\)"server_max_value_length.*$/\1"ignore_exc": True,/' "$SENTRY_CONFIG_PY"
```
azaslavsky, line 9: What if they have more configs on this line? This will erase those. How about:

```suggestion
    sed -i '' -e 's/^\( *\)"server_max_value_length.+?\),/\1"ignore_exc": True,/' "$SENTRY_CONFIG_PY"
```

But I think we could avoid all of this search and replace if we took the `CACHES_BAK` route above, and just wrote the new config wholesale.

install/check-memcached-backend.sh
chadwhitacre: That sounds great, but imo we're out of time. We need to get a security release out the door and not ship the memcached breakage with it. :(

chadwhitacre: @azaslavsky Do you think it would be better to simply hard-fail if they have `.MemcachedCache` in their config file, with some docs on how to manually address?

azaslavsky: Honestly, hard fail seems like the best way to go. If people are up in arms about it, we can always cut another release with one of these automatic config changes we're discussing. But using regex to try and guess at user's non-version-controlled configs is always risky imo.

I can write up the doc if you want to just change this PR to point the user to https://develop.sentry.dev/self-hosted/releases/#breaking-changes?

chadwhitacre: Deal.

chadwhitacre: Done in b8e8baa353bd2d8f4f2b0474154b9dc43adc4a14.

azaslavsky:
LGTM, ready to approve that future PR.

azaslavsky (approved):

PR #2821, opened 2024-02-22
fix(install): Manually exit for faster wrap up

azaslavsky:


BYK:
Can you explain why/how this is faster as a comment in the code?

PR #2829, opened 2024-02-26
Revert changes in 3067683f6c0e1c6dd9ceb72cb5155c1dbf3bf501

hubertdeng123:
This made it directly to master...perhaps we are missing some branch protection rules

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -254,35 +254,35 @@ services:
   # Kafka consumer responsible for feeding events into Clickhouse
   snuba-consumer:
     <<: *snuba_defaults
-    command: rust-consumer --storage errors --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
+    command: consumer --storage errors --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
   # Kafka consumer responsible for feeding outcomes into Clickhouse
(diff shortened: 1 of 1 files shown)


hubertdeng123:
> It's because the rules can be silently bypassed by repo admins. We had this issue before 😞

GitHub seems to have a rule now where there is a checkbox for admins to not bypass branch protection rules. That is now enabled so this shouldn't happen again! :)

BYK:
>GitHub seems to have a rule now where there is a checkbox for admins to not bypass branch protection rules. That is now enabled so this shouldn't happen again! :)

That existed for a very long while and there was/is a reason why it was not enabled. I recommend you to uncheck that box before the next release or you're gonna be in for some fun.

This was documented here: https://github.com/getsentry/publish/pull/1496/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L59 but looks like it was missed when the Craft documentation was created. You may wanna ping @asottile-sentry to update the docs accordingly as this is a critical part of the flow.

hubertdeng123:
Thanks for the context, checked with @asottile-sentry and it appears that the bot now has a separate role that bypasses branch protection. So, hopefully the release will go smoothly 😅...only one way to find out

BYK:
Ah, right. Now _that_ came quite a while later and it makes sense. Don't think you'd have any issues with the release then!

Issue #2836, opened 2024-02-27
After 24.1.1 upgrade: envelope buffer capacity exceeded


TimofeyMakoveychuk:
### Self-Hosted Version

24.1.1

### CPU Architecture

x86_64

### Docker Version

24.0.2

### Docker Compose Version

2.17.3

### Steps to Reproduce

1. Upgrade Sentry from 24.1.0 to 24.1.1

### Expected Result

Sentry should not drop any incoming errors

### Actual Result

All incoming error events are dropped. Go to /organizations/YOUR_ORGANISATION/stats/ and you will see that all error events are dropped.

We have about 100K events/hour and all are dropped in version 24.1.1. In version 24.1.0 all are either filtered or accepted by Sentry.

The log /docker/relay.log has bursts of the following errors: `The error log ERROR relay_server::endpoints::common: error handling request error=failed to queue envelope error.sources=[envelope buffer capacity exceeded]` and `ERROR relay_server::utils::managed_envelope: dropped envelope: internal error`

We can also observe the following errors in our frontend: `POST /api/* 503 (Service Unavailable)`

csvan:
We have hit the same issue as of 24.1.1, it's also on 24.2.0. Solution on our end is a cron job to restart Sentry every 24h but this is of course not sustainable.

Issue does not seem to be that the system actually overflows the buffer, rather it _just stops processing events altogether, quickly causing it to build up_.

csvan:
It _may_ be related to this, especially if you have high traffic: https://github.com/getsentry/self-hosted/issues/2825

hubertdeng123:
Thanks for filing this. For both your instances, are all errors being dropped? That definitely does not appear to be normal. What does CPU/RAM usage look like?

TimofeyMakoveychuk:
> Thanks for filing this. For both your instances, are all errors being dropped? That definitely does not appear to be normal. What does CPU/RAM usage look like?
Yes, all errors are being dropped.

csvan:
@hubertdeng123 in our case memory goes up by around 20%, and peaks once it starts dropping envelopes from the buffer.

CPU usage goes down to very low levels (around 10% usage), most likely because everything is being dropped and no processing happens.

TimofeyMakoveychuk:
@RexTim Thank you! Removing `proxy_buffering off;` in nginx.conf solved the issue for us :)

maxpain:
The same problem on 24.2.0
`proxy_buffering off;` didn't help

[...]

PR #2844, opened 2024-02-29
Enable proxy buffering in nginx

RexTim:
Setting the proxy_buffering directive to “off” is a common mistake because it can cause performance issues and unexpected behavior in NGINX. When proxy buffering is disabled, NGINX receives a response from the proxied server and immediately sends it to the client without storing it in a buffer. This can cause problems if the response is large or if the connection between NGINX and the client is slow, because it can result in increased memory usage and potentially lead to request timeouts.

To avoid this mistake, it is recommended to always enable proxy buffering in NGINX.

Resolves https://github.com/getsentry/self-hosted/issues/2836

### Legal Boilerplate

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -37,7 +37,6 @@ http {
 
 	proxy_http_version 1.1;
 	proxy_redirect off;
-	proxy_buffering off;
 	proxy_buffer_size          128k;
 	proxy_buffers              4 256k;
 	proxy_busy_buffers_size    256k;


hubertdeng123 (approved):

hubertdeng123 (commented):
seems fine, proxy buffering is enabled by default

aminvakil:
Also to add https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#proxy_buffering-off

But maybe we should think about changing proxy_buffer_size as well.

aminvakil:
I guess I was late about 5 seconds :)

RexTim:
@aminvakil proxy_buffer_size is set to 128k, I believe it should be sufficient for most cases?

https://github.com/getsentry/self-hosted/pull/2844/files#diff-4adaaefa7fc959e7dfe65e8dc8dadc9ee27ddd284eee3426463b1cdc000587d5R40

Issue #2832, opened 2024-02-26
Clickhouse corrupt after server failure


RoyvanEmpel:
### Environment

self-hosted (https://develop.sentry.dev/self-hosted/)

### What are you trying to accomplish?

Hi,

The server that our self hosted sentry runt on had a hardware failure causing it to shutdown. After getting everything back up and running sentry no longer starts up.

Running docker compose up -d i get the following output:
```
dependency failed to start: container sentry-self-hosted-clickhouse-1 is unhealthy
```

I tried reinstalling using ./install.sh and after that i tried updating it to the latest version but i again get this output:

> ▶ Bootstrapping and migrating Snuba ...
>  Container sentry-self-hosted-zookeeper-1  Creating
>  Container sentry-self-hosted-redis-1  Creating
>  Container sentry-self-hosted-clickhouse-1  Creating
>  Container sentry-self-hosted-redis-1  Created
>  Container sentry-self-hosted-zookeeper-1  Created
>  Container sentry-self-hosted-kafka-1  Creating
>  Container sentry-self-hosted-clickhouse-1  Created
>  Container sentry-self-hosted-kafka-1  Created
>  Container sentry-self-hosted-clickhouse-1  Starting
>  Container sentry-self-hosted-zookeeper-1  Starting
>  Container sentry-self-hosted-redis-1  Starting
>  Container sentry-self-hosted-redis-1  Started
>  Container sentry-self-hosted-clickhouse-1  Started
>  Container sentry-self-hosted-zookeeper-1  Started
>  Container sentry-self-hosted-zookeeper-1  Waiting
>  Container sentry-self-hosted-zookeeper-1  Healthy
>  Container sentry-self-hosted-kafka-1  Starting
>  Container sentry-self-hosted-kafka-1  Started
> dependency failed to start: container sentry-self-hosted-clickhouse-1 is unhealthy
> Error in install/bootstrap-snuba.sh:3.
> '$dcr snuba-api bootstrap --no-migrate --force' exited with status 1
> -> ./install.sh:main:32
> --> install/bootstrap-snuba.sh:source:3
> 
> Cleaning up...

### How are you getting stuck?

Because of this sentry isn't starting. It seems that the clickhouse DB is corrupted, can i just drop the files of it and will it recover? I am guessing that issues are stored in it.
Maybe there is a command to try a self-repair?

Oh and BTW:
this says that the latest version is 24.3.0 while it is 24.2.0
[...]

PR #2853, opened 2024-03-04
feat(clickhouse): Added max_suspicious_broken_parts to the config.xml

victorelec14:
This PR adds the "max_suspicious_broken_parts" setting which is necessary to imitate corrupted data in the Clickhouse database.

The default setting was 10, but it was changed to 100. In my case, setting it to 1000 solved the problems I had and allowed Clickhouse to start again normally.

Related Sentry Issue: #2832

Clickhouse References:
https://github.com/ClickHouse/ClickHouse/issues/41423
https://github.com/ClickHouse/ClickHouse/pull/41619
https://clickhouse.com/docs/en/operations/settings/merge-tree-settings

### Legal Boilerplate


azaslavsky:
How about we add the setting, but keep it at the default (10). Then above that, we can add a comment pointing to issue #2832 with a brief note about how you might adjust the setting to recover a corrupt install. That way future users will have a searchable way to fix this problem, but we won't mess up any active installs that may be relying on the old default.

victorelec14:
> How about we add the setting, but keep it at the default (10). Then above that, we can add a comment pointing to issue #2832 with a brief note about how you might adjust the setting to recover a corrupt install. That way future users will have a searchable way to fix this problem, but we won't mess up any active installs that may be relying on the old default.

LGTM, I just updated the code and added a reference to the Issue, thanks !. 👍

hubertdeng123 (approved):

PR #2858, opened 2024-03-05
Integration test improvements

hubertdeng123:
This PR adds a couple of improvements to our integration test pipeline.

1. Parallelizes some tests, runtime was 23+ minutes. Now this is decreased to <18 minutes
2. Reports failures to Sentry project

In order to report flakes, I wanted to cut down the time it takes to actually run these tests. Given that with 20 min tests, it could take an hour if the tests flake twice.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,22 +86,12 @@ jobs:
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
 
       - name: Install self-hosted
-        uses: nick-fields/retry@v3
-        with:
-          max_attempts: 3
-          timeout_minutes: 10
-          retry_on: error
-          command: |
-            export REPORT_SELF_HOSTED_ISSUES=1
-            ./install.sh
+        run: |
+          export REPORT_SELF_HOSTED_ISSUES=1
+          ./install.sh
 
       - name: Integration Test
-        uses: nick-fields/retry@v3
-        with:
-          max_attempts: 3
-          timeout_minutes: 25
-          retry_on: error
-          command: ./integration-test.sh --${{ matrix.test_type }}
+        run: ./integration-test.sh --${{ matrix.test_type }}
 
       - name: Inspect failure
         if: failure()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,7 @@ jobs:
             compose_path: "/usr/local/lib/docker/cli-plugins"
     env:
       COMPOSE_PROJECT_NAME: self-hosted-${{ strategy.job-index }}
+      SENTRY_DSN: https://5a620019b5124cbba230a9e62db9b825@o1.ingest.us.sentry.io/6627632
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -91,7 +92,7 @@ jobs:
           timeout_minutes: 10
           retry_on: error
           command: |
-            export REPORT_SELF_HOSTED_ISSUES=0
+            export REPORT_SELF_HOSTED_ISSUES=1
             ./install.sh
 
       - name: Integration Test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,7 @@ jobs:
     env:
       COMPOSE_PROJECT_NAME: self-hosted-${{ strategy.job-index }}
       SENTRY_DSN: https://5a620019b5124cbba230a9e62db9b825@o1.ingest.us.sentry.io/6627632
+      REPORT_SELF_HOSTED_ISSUES: 1
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -86,9 +87,7 @@ jobs:
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
 
       - name: Install self-hosted
-        run: |
-          export REPORT_SELF_HOSTED_ISSUES=1
-          ./install.sh
+        run: ./install.sh
 
       - name: Integration Test
         run: ./integration-test.sh --${{ matrix.test_type }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,12 +59,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        test_type: ["initial-install", "customizations"]
+        compose_version: ["v2.0.1", "v2.7.0"]
         include:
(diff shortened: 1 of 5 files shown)


BYK:
I don't understand how this is going to help. Can you share your thinking?

hubertdeng123:
> I don't understand how this is going to help. Can you share your thinking?

Please ignore things going on in this branch for now, I am just using this as a means to run CI while I am experimenting

_integration-test/run.sh
```diff
@@ -41,10 +45,10 @@ echo "${_endgroup}"
 echo "${_group}Starting Sentry for tests ..."
 # Disable beacon for e2e tests
 echo 'SENTRY_BEACON=False' >>$SENTRY_CONFIG_PY
-echo y | $dcr web createuser --force-update --superuser --email $TEST_USER --password $TEST_PASS
```
hubertdeng123, line 44: docker run is slower than just an exec into a running container

hubertdeng123:
`integration test (v2.0.1, /usr/local/lib/docker/cli-plugins)` and `integration test (v2.7.0, /usr/local/lib/docker/cli-plugins) ` are expected and pending because of branch protection rules, if this PR goes in I'll change those

.github/workflows/test.yml
```diff
@@ -82,8 +84,23 @@ jobs:
           sudo curl -L https://github.com/docker/compose/releases/download/${{ matrix.compose_version }}/docker-compose-`uname -s`-`uname -m` -o "${{ matrix.compose_path }}/docker-compose"
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
 
+      - name: Install self-hosted
+        uses: nick-fields/retry@v3
+        with:
+          max_attempts: 3
+          timeout_minutes: 10
```
azaslavsky, line 91: Any chance this could actually increase the flake rate, due to timeouts becoming more frequent? Or are flakes caused by the actions itself timing out?

.github/workflows/test.yml
```diff
@@ -82,8 +84,23 @@ jobs:
...
+          retry_on: error
+          command: |
+            export REPORT_SELF_HOSTED_ISSUES=0
+            ./install.sh
+
       - name: Integration Test
-        run: ./integration-test.sh
+        uses: nick-fields/retry@v3
+        with:
+          max_attempts: 3
```
azaslavsky, line 100: Just so I understand: what was the behavior before you added these settings? Unlimited retries? No timeout so it hung until the action crashed?

_integration-test/run.sh
```diff
@@ -28,7 +28,12 @@ teardown() {
   DID_TEAR_DOWN=1
 
   if [ "$1" != "EXIT" ]; then
-    echo "An error occurred, caught SIG$1 on line $2"
+    set +x
+    error_msg="An error occurred, caught SIG$1 on line $2"
+    echo "$error_msg"
+    dsn="https://5a620019b5124cbba230a9e62db9b825@o1.ingest.us.sentry.io/6627632"
+    local sentry_cli="docker run --rm -v $(pwd):/work -e SENTRY_DSN=$dsn getsentry/sentry-cli"
```
azaslavsky, line 35: Odd to me that we set REPORT_ISSUES=0 above, but then send an envelope here anyway? Maybe I am misunderstanding something, but why would we not just do `REPORT_ISSUES=1`, and maybe figure out some way to configure it to send to an "integration tests" or similar project on our actual dogfood instance?

_integration-test/run.sh
```diff
@@ -41,10 +46,10 @@ echo "${_endgroup}"
 echo "${_group}Starting Sentry for tests ..."
 # Disable beacon for e2e tests
 echo 'SENTRY_BEACON=False' >>$SENTRY_CONFIG_PY
-echo y | $dcr web createuser --force-update --superuser --email $TEST_USER --password $TEST_PASS
 $dc up -d
-printf "Waiting for Sentry to be up"
 timeout 90 bash -c 'until $(curl -Isf -o /dev/null $SENTRY_TEST_HOST); do printf '.'; sleep 0.5; done'
+echo y | $dc exec web sentry createuser --force-update --superuser --email $TEST_USER --password $TEST_PASS
```
azaslavsky, line 51: What happens if we exec into the web container but it isn't ready yet? Does docker wait until it is ready, or does this fail? If it's the former, we should `echo` "Waiting for Sentry..." before this runs, otherwise the user may be waiting a while. If it fails, we should add some sort of sync point to wait for the container to be up before trying this.

_integration-test/run.sh
hubertdeng123: Sure, I can incorporate this into the existing logic for `REPORT_ISSUES` I believe.

_integration-test/run.sh
hubertdeng123: Docker compose up will only succeed if the container healthcheck for web passes, so I don't think this will be a problem. That is performed on a previous line, so when the tests get to the createuser logic the web container will always be ready

.github/workflows/test.yml
hubertdeng123: The typical install logic takes around 4-5 minutes. I doubled that time for the timeout so I don't think this should ever increase the flake rate.

.github/workflows/test.yml
hubertdeng123: Yep, that's correct. The `max_attempts` really is just the first step into adding flaky test detection. If a job fails, but then is retried and succeed, it can be marked as flaky. The `timeout_minutes` is a required parameter here. I can remove this and readd in a follow-up PR if that is more clear

azaslavsky: No, that's fine. I just wanted to understand the change. LGTM!

_integration-test/run.sh
BYK: It is but it is also intentional to keep the run separate from the main web process.

That said for a one-off thing like this, I think using `exec` is a good compromise if it saves a notable amount of time. I'd just want it documented with a brief comment.

BYK (approved):
Love the changes. BTW not sure how much time it costs us but maybe we can also use system-installed versions of `jq` and `curl` if we detect them and use the Docker-based ones as a fallback?

azaslavsky (approved):

hubertdeng123:
> BTW not sure how much time it costs us but maybe we can also use system-installed versions of jq and curl if we detect them and use the Docker-based ones as a fallback?

Good point, I can investigate that in a follow up

PR #2866, opened 2024-03-07
feat(vroom): Allow to customize vroom's environment variables in a different file

phacops:
We've had a request to be able to customize `vroom`'s configuration (which is using environment variables) via a file. 

Instead of implementing a file based configuration, we can simplify this by loading from a different environment file dedicated to `vroom`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -440,13 +440,8 @@ services:
     <<: *restart_policy
     image: "$VROOM_IMAGE"
     environment:
-      SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
-      SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
-      SENTRY_BUCKET_PROFILES: file://localhost//var/lib/sentry-profiles
-      SENTRY_SNUBA_HOST: "http://snuba-api:1218"
     env_file:
-      - path: ./vroom.env
-        required: false
+      - ./vroom.env
     volumes:
       - sentry-vroom:/var/lib/sentry-profiles
     depends_on:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -439,11 +439,8 @@ services:
   vroom:
     <<: *restart_policy
     image: "$VROOM_IMAGE"
-    environment:
-      SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
-      SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
-      SENTRY_BUCKET_PROFILES: file://localhost//var/lib/sentry-profiles
-      SENTRY_SNUBA_HOST: "http://snuba-api:1218"
+    env_file:
```
hubertdeng123, line 442: This will require docker compose v2.24.0+. Which, I think we do want to get to due to how we can use this for docker compose env override files. For now though, for visibility purposes if we're going on this route it seems ok to include these vroom specific env variables into the `.env` file. I know it's not what you originally envisioned @aldy505, but will that do for now?

aldy505: I still don't prefer this route, because that would mean there are no unified way to configure sentry services, as people know currently it's been done through config files and it works for most of the deployment models (docker compose, k8s, docker swarm, separate compose file, etc -- it's been told on the community) 😞. But since I'm not 100% of the community's voice here, I'd like to hear about what others (who self-hosted their sentry) )would say about this.

vroom.env
```diff
@@ -0,0 +1,4 @@
+SENTRY_KAFKA_BROKERS_PROFILING="kafka:9092"
+SENTRY_KAFKA_BROKERS_OCCURRENCES="kafka:9092"
+SENTRY_BUCKET_PROFILES=file://localhost//var/lib/sentry-profiles
```
williamdes, line 3: awesome !
Looking forward to store profiles on something else than my drive.
it runs out of inodes too quickly.

phacops:
It seems we'll need more consensus on the way to have a file to configure `vroom`. This PR doesn't seem like the right approach.

PR #2874, opened 2024-03-08
Add flaky test detector

hubertdeng123:


hubertdeng123:
going to put this on hold since it's causing me a lot of pain, going to look into converting the `run.sh` file to python

BYK:
The reason we chose bash was that we could rely on it being available on most of the target systems and without any dependencies. 

If we're going to revisit that, may I humbly suggest considering https://xon.sh through a docker image?

That said it may require mounting docker.sock which might be a trouble on certain systems.

hubertdeng123:
Just to clarify, I wasn't planning on rewriting the installer. For integration tests, it doesn't seem like using bash would be a clear advantage given that we're controlling the testing environment and the dependencies that come with it. Looks like you are suggesting xonsh to be used for the installer?

BYK:
Oh, right sorry. You may still benefit from using xonsh which is essentially a marriage between Python and bash. But yes, I was thinking about that for the installer.

hubertdeng123:
Closing this, as https://github.com/getsentry/self-hosted/pull/2918 addresses flaky test reporting

PR #2892, opened 2024-03-13
Integration tests in python

hubertdeng123:
This adds a couple of exciting things and will be the first step into modernizing the test suite for self-hosted. 

1. integration tests in python
2. pytest-sentry so that pytest failures will automatically be reported to a Sentry project
3. retries for pytest test suite


integration-test.sh
```diff
@@ -15,12 +15,11 @@ export MINIMIZE_DOWNTIME=0
 
 if [[ "$test_option" == "--initial-install" ]]; then
```
hubertdeng123, line 16: The goal is to rewrite this file entirely in python using pytest, but baby steps!

_integration-test/run.py
```diff
@@ -0,0 +1,159 @@
...
+    return subprocess.run(cmd, shell=True, check=True, **kwargs)
+
+
+class IntegrationTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.session = requests.Session()
+        response = None
+        run_shell_command("docker compose --ansi never up -d")
+        for i in range(TIMEOUT_SECONDS):
```
hubertdeng123, line 27: this is pretty ugly, open to suggestions on how to make polling for a response cleaner

.github/workflows/test.yml
```diff
@@ -74,6 +74,21 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Setup dev environment
+        run: |
+          pip install -r requirements-dev.txt
+          ### pytest-sentry configuration ###
+          if [ "$GITHUB_REPOSITORY" = "getsentry/self-hosted" ]; then
+            echo "PYTEST_SENTRY_DSN=https://5a620019b5124cbba230a9e62db9b825@o1.ingest.us.sentry.io/6627632" >> $GITHUB_ENV
+            echo "PYTEST_SENTRY_TRACES_SAMPLE_RATE=0" >> $GITHUB_ENV
```
asottile-sentry, line 83: I would use a different dsn if you haven't already -- and ~slightly hide it in github variables (doesn't have to be a secret, just a repository variable) -- should help with forks sending garbage into this

_integration-test/custom-ca-roots/setup.sh
```diff
@@ -25,8 +25,11 @@ openssl req -new -nodes -newkey rsa:2048 -keyout $TEST_NGINX_CONF_PATH/self.test
 
 # openssl req -in nginx/self.test.req -text -noout
 
+# Store extension details in a temporary file since python subprocess can't use process substitution
+printf "subjectAltName=DNS:self.test" >/tmp/extension_details
```
asottile-sentry, line 29: you should be able to do this inline in python just fine -- could you share what you're running into here?

as written this is susceptible to tempfile injection since it's using an easily guessable (static) path

_integration-test/run.py
```diff
@@ -0,0 +1,159 @@
...
+TEST_USER = "test@example.com"
+TEST_PASS = "test123TEST"
+TIMEOUT_SECONDS = 60
+
+
+def run_shell_command(cmd, **kwargs) -> subprocess.CompletedProcess:
+    return subprocess.run(cmd, shell=True, check=True, **kwargs)
+
+
+class IntegrationTests(unittest.TestCase):
```
asottile-sentry, line 21: I would not recommend using `unittest.TestCase` -- especially for new projects

you should use [pytest](https://pytest.org) to write tests (module-level `def test_...`) and fixtures (`@pytest.fixture`)

_integration-test/run.py
```diff
@@ -0,0 +1,159 @@
...
+    @classmethod
+    def setUpClass(cls):
```
asottile-sentry, line 23: this would probably be a `session` fixture -- essentially waiting until the app has started

asottile-sentry (commented):
just a start at reviewing -- I think this is going in the right direction though!  I suspect a lot will change from my initial comments so I didn't want to dig in too much into the rest of it

.github/workflows/test.yml
hubertdeng123: This DSN is specific for self-hosted tests so it won't be shared with sentry. I can put it as a variable in GH

_integration-test/custom-ca-roots/setup.sh
hubertdeng123: Ah, good point. This is the error I'm getting:
```
_integration-test/custom-ca-roots/setup.sh: line 29: syntax error near unexpected token `('
```

Noting that I do get the same thing running locally, but in CI it does not fail

asottile-sentry: can you show what code led to that?

hubertdeng123: Specifically this line:
```
openssl x509 -req -in $TEST_NGINX_CONF_PATH/self.test.req -CA $TEST_NGINX_CONF_PATH/ca.crt -CAkey $TEST_NGINX_CONF_PATH/ca.key \
  -extfile <(printf "subjectAltName=DNS:self.test") \
  -CAcreateserial -out $TEST_NGINX_CONF_PATH/self.test.crt -days 1 -sha256
```

the `<(printf "subjectAltName=DNS:self.test")`

asottile-sentry: ah yeah so what's happening in bash is it's making a process and file descriptor and substituting that into the command

you can pretty easily do that in python as well -- but you need to manage the file descriptor yourself

it's probably easier to use `NamedTemporaryFile` rather than an anonymous descriptor

hubertdeng123: Makes sense. I would like to eventually rewrite the entire file in python, but would like to do that in a follow up PR. For now, I decided to use `mktmp` so it's no longer a static path.

_integration-test/custom-ca-roots/setup.sh
```diff
@@ -25,8 +25,12 @@ openssl req -new -nodes -newkey rsa:2048 -keyout $TEST_NGINX_CONF_PATH/self.test
 
 # openssl req -in nginx/self.test.req -text -noout
 
+# Store extension details in a temporary file since python subprocess can't use process substitution
+temp_file=$(mktemp)
+printf "subjectAltName=DNS:self.test" >"$temp_file"
+
```
asottile-sentry, line 31: oh this shouldn't change at all -- this is failing because the file is missing a shebang `#!/usr/bin/env bash` and is using bash features.  the os will default to `sh` which doesn't understand the original syntax

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+
+SENTRY_CONFIG_PY = "sentry/sentry.conf.py"
+SENTRY_TEST_HOST = os.getenv("SENTRY_TEST_HOST", "http://localhost:9000")
+TEST_USER = "test@example.com"
+TEST_PASS = "test123TEST"
+TIMEOUT_SECONDS = 60
+
+
+def run_shell_command(cmd, **kwargs) -> subprocess.CompletedProcess:
+    return subprocess.run(cmd, shell=True, check=True, **kwargs)
```
asottile-sentry, line 18: shell=True is an absolute no-go -- this is how you create security issues

this function shouldn't exist at all imo

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
+import subprocess
+import os
+from functools import lru_cache
+import requests
```
asottile-sentry, line 4: I would not recommend using requests in new code

if you can get away with it, stdlib `urllib.request` does not require additional dependencies.  httpx is a modern replacement if you find urllib.request too cumbersome

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+    return subprocess.run(cmd, shell=True, check=True, **kwargs)
+
+
+def poll_for_response(request: str, session: requests.Session) -> requests.Response:
+    for i in range(TIMEOUT_SECONDS):
+        response = session.get(request, headers={"Referer": SENTRY_TEST_HOST})
+        if response.status_code == 200:
+            break
+        time.sleep(1)
+    assert response.status_code == 200
```
asottile-sentry, line 27: better would be to use `for: ... else: raise AssertionError('timeout waiting for ...')`

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+    for i in range(TIMEOUT_SECONDS):
+        response = session.get(
+            f"{SENTRY_TEST_HOST}/api/0/projects/sentry/internal/keys/",
+            headers={"Referer": SENTRY_TEST_HOST},
+        )
+        if response.status_code == 200:
+            sentry_dsn = json.loads(response.text)[0]["dsn"]["public"]
+            if len(sentry_dsn) > 0:
+                break
+        time.sleep(1)
```
asottile-sentry, line 42: this looks suspiciously like the function above

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+            response = requests.get(SENTRY_TEST_HOST)
+        except requests.ConnectionError:
+            time.sleep(1)
+        if response is not None and response.status_code == 200:
+            break
+    assert response.status_code == 200
+
+    # Create test user
+    run_shell_command(
+        f"echo y | docker compose exec web sentry createuser --force-update --superuser --email {TEST_USER} --password {TEST_PASS}"
```
asottile-sentry, line 63: there's a `--noinput` option to this I believe

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+@pytest.fixture(scope="function")
+def session_login():
+    session = requests.Session()
+    login_csrf_token = (
+        session.get(SENTRY_TEST_HOST)
+        .text.split('"csrfmiddlewaretoken" value="')[1]
+        .split('"')[0]
+    )
```
asottile-sentry, line 74: this seems really fragile -- is there not a standardized format for this? (what does the response look like?)

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+            break
+    assert response.status_code == 200
+
+    # Create test user
+    run_shell_command(
+        f"echo y | docker compose exec web sentry createuser --force-update --superuser --email {TEST_USER} --password {TEST_PASS}"
+    )
+
+
+@pytest.fixture(scope="function")
```
asottile-sentry, line 67: `scope="function"` is the default
```suggestion
@pytest.fixture
```

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+    # Set up initial/required settings (InstallWizard request)
+    session.headers.update({"X-CSRFToken": login_response.cookies["sc"]})
+    response = session.put(
+        f"{SENTRY_TEST_HOST}/api/0/internal/options/?query=is:required",
+        headers={"Referer": SENTRY_TEST_HOST},
+        data={
+            "mail.use-tls": False,
+            "mail.username": "",
+            "mail.port": 25,
+            "system.admin-email": "ben@byk.im",
```
asottile-sentry, line 109: tests should use `example.com` as a placeholder domain

_integration-test/run.py
```diff
@@ -0,0 +1,176 @@
...
+    run_shell_command(". _integration-test/custom-ca-roots/teardown.sh")
+
+
+def test_receive_transaction_events(session_login):
+    session, _ = session_login
+    with sentry_sdk.init(
+        dsn=get_sentry_dsn(session), profiles_sample_rate=1.0, traces_sample_rate=1.0
+    ):
+
+        def dummy_func():
```
asottile-sentry, line 155: ableist language
```suggestion
        def placeholder_func():
```

asottile-sentry (changes requested):
the `shell=True` is a blocker

_integration-test/custom-ca-roots/setup.sh
hubertdeng123: I did try this locally, but it still didn't work. Let me try it in CI to see if it makes a difference

hubertdeng123: Looks like it did not work https://github.com/getsentry/self-hosted/actions/runs/8330451282/job/22795132410?pr=2892

asottile-sentry: the calling code is also wrong -- it's using `.` which is an alias to `source` (instead of executing it) which isn't going to do what you expect anyway:  https://youtu.be/YE6ZDpRcdQU

_integration-test/run.py
hubertdeng123: I agree with you, seems pretty fragile here but don't think there's a good way to go about this. The response here is returned in html, and the value of the token is hidden under  a hidden input 
```<input type="hidden" name="csrfmiddlewaretoken" value="token">```
in a form. Perhaps I can use something like BeautifulSoup to make this cleaner though

Comment (asottile-sentry) [reply _integration-test/run.py:74]: the stdlib has `html.parser` that you might be able to use without needing a dependency

File: _integration-test/run.py (hunk 9 above)
Comment (hubertdeng123) [reply _integration-test/run.py]: switched to using httpx, since I wanted a good way to save cookies

File: _integration-test/run.py (hunk 8 above)
Comment (hubertdeng123) [reply _integration-test/run.py]: removed all instances of shell=True

File: _integration-test/run.py (hunk 10 above)
Comment (hubertdeng123) [reply _integration-test/run.py]: did not know that existed! 👍

File: _integration-test/run.py (hunk 12 above)
Comment (hubertdeng123) [reply _integration-test/run.py]: ended up using the `input` option in subprocess.run which

File: _integration-test/run.py (hunk 9 above)
Comment (hubertdeng123) [reply _integration-test/run.py]: What's the reason why requests is not the preferred choice?

File: _integration-test/custom-ca-roots/setup.sh (hunk 7 above)
Comment (hubertdeng123) [reply _integration-test/custom-ca-roots/setup.sh]: calling code corrected, now it's working :)

File: _integration-test/run.py (hunk 8 above)
Comment (hubertdeng123) [reply _integration-test/run.py]: shell injection...ahh gotcha

File: _integration-test/run.py [hunk 17]
```diff
@@ -0,0 +1,197 @@
...
+TEST_PASS = "test123TEST"
+TIMEOUT_SECONDS = 60
+
+
+def poll_for_response(request: str, client: httpx.Client) -> httpx.Response:
+    for i in range(TIMEOUT_SECONDS):
+        response = client.get(
+            request, follow_redirects=True, headers={"Referer": SENTRY_TEST_HOST}
+        )
+        if response.status_code == 200:
```
phacops, line 24: This function won't work for spans as it's waiting for a span to be ingested to succeed, so it's receiving a 200 no matter what and looking for the `data` field to not be empty.

We risk flaky tests if we leave it at that. Maybe we can have a custom behavior via a function to return `True`/`False` if it's considered a success of if we need to retry?

integration-test.sh
```diff
@@ -15,7 +15,7 @@ export MINIMIZE_DOWNTIME=0
...
   echo "Testing initial install"
-  source _integration-test/run.sh
+  pytest --reruns 5 _integration-test/run.py
```
phacops, line 18: Is that even needed? If we have proper retries when polling data, we likely don't need to retry the tests up to 5 times, do we?

_integration-test/run.py
```diff
@@ -0,0 +1,197 @@
...
+            "--force-update",
+            "--superuser",
+            "--email",
+            TEST_USER,
+            "--password",
+            TEST_PASS,
+        ],
+        check=True,
+        text=True,
+        input="y",
```
asottile-sentry, line 75: this is what sentry does to avoid needing to pass input: https://github.com/getsentry/sentry/blob/d84a4ea18bae97b933393261a7f1554bb5537e58/scripts/lib.sh#L198

_integration-test/run.py
```diff
@@ -0,0 +1,197 @@
...
+def test_initial_redirect():
+    initial_auth_redirect = httpx.get(SENTRY_TEST_HOST, follow_redirects=True)
+    assert initial_auth_redirect.url == f"{SENTRY_TEST_HOST}/auth/login/sentry/"
+
+
+def test_login(client_login):
+    client, login_response = client_login
+    assert '"isAuthenticated":true' in login_response.text
+    assert '"username":"test@example.com"' in login_response.text
+    assert '"isSuperuser":true' in login_response.text
```
asottile-sentry, line 109: this is fragile -- this should use some sort of parser to acquire this information (is it json?)

_integration-test/run.py
```diff
@@ -0,0 +1,197 @@
...
+    cleanup_crons = pattern.findall(docker_services)
+    assert len(cleanup_crons) > 0
+
+
+def test_custom_cas():
+    subprocess.run(["./_integration-test/custom-ca-roots/setup.sh"], check=True)
+    subprocess.run(
+        ["docker", "compose", "--ansi", "never", "run", "--no-deps", "web", "python3", "/etc/sentry/test-custom-ca-roots.py"], check=True
+    )
+    subprocess.run(["./_integration-test/custom-ca-roots/teardown.sh"], check=True)
```
asottile-sentry, line 167: should probably be `try:` / `finally:` so the teardown runs?  or move the setup/teardown to a `yield`ing fixture

integration-test.sh
asottile-sentry, line 18: I would recommend *not* having reruns -- they're a crutch that allows flakiness to creep in and ideally we wouldn't have them in `sentry`

asottile-sentry (commented):
neat!  probably close to shippable :D

integration-test.sh
hubertdeng123: Primary reason behind this is because docker compose itself is quite flaky in CI, and I wanted to have a way to rerun  the fixture that brings self-hosted up. 

https://github.com/getsentry/team-ospo/issues/242

_integration-test/run.py
```diff
@@ -0,0 +1,197 @@
...
+def configure_self_hosted_environment():
+    response = None
+    subprocess.run(["docker", "compose", "--ansi", "never", "up", "-d"], check=True)
+    for i in range(TIMEOUT_SECONDS):
+        try:
+            response = httpx.get(SENTRY_TEST_HOST, follow_redirects=True)
+        except httpx.ConnectionError:
+            time.sleep(1)
+        if response is not None and response.status_code == 200:
+            break
```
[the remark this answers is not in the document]
hubertdeng123: done

_integration-test/run.py
hubertdeng123: it is json inside of an html body, added a html parser and json parser to get this information so it's less fragile

_integration-test/run.py
hubertdeng123: good point, added

asottile-sentry (approved):

PR #2908, opened 2024-03-19
Use transactions python consumer

hubertdeng123:
Seems like this is not used in prod, and not extensively tested and may be the reason why the Stats page is not properly showing events that are accepted/filtered/dropped.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -266,7 +266,7 @@ services:
   # Kafka consumer responsible for feeding transactions data into Clickhouse
   snuba-transactions-consumer:
     <<: *snuba_defaults
-    command: rust-consumer --storage transactions --consumer-group transactions_group --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset --no-skip-write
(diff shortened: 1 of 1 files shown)


lynnagara (approved):

aldy505 (approved):
I'm gonna give it a try in a few hours.

hubertdeng123:
nvm, stats page fix should be here:
https://github.com/getsentry/self-hosted/pull/2909

hubertdeng123:
However, this should still be merged

zKoz210:
I noticed a significant degradation in processing speed. Now events are displayed after an hour. At a measly 400-500 rps

hubertdeng123:
@zKoz210 is this after switching to rust python transactions consumer? Or using the python transactions python consumer?

zKoz210:
After changing to rust, python works for me without any problems.

zKoz210:
@hubertdeng123 I'm still experiencing performance issues with event processing. Every new release I have to roll back Rust and return old workers. Can I somehow help find and solve this problem?

zKoz210:
image

Here you can see how workers worked in Python, then I turned on Rust again and returned Python back

hubertdeng123:
Closing as this PR never resolved the underlying issue it was supposed to, and switching from rust consumers to python consumers is a separate issue

PR #2909, opened 2024-03-19
feat: run outcomes-billing consumer

lynnagara:


hubertdeng123 (approved):

aldy505:
Question @lynnagara, what's the outcomes-billing is used for? I'm just curious to know what this container does.

lynnagara:
> Question @lynnagara, what's the outcomes-billing used for? I'm just curious to know what this container does.

Statistics that are used to calculate usage for billing purposes are produced on that topic. It's less relevant in self-hosted enviroments since quotas and billing work very differently but I think it's also used to power some other stats pages that may not be working otherwise.

PR #2918, opened 2024-03-21
Write Customization tests in python

hubertdeng123:
This PR adds a few things

1. Codecov reporting for self-hosted
2. Proper flaky test detection for all of integration tests in self-hosted with pytest-sentry
3. All integration tests are now running through pytest


hubertdeng123:
Flaky test reporting works!
https://sentry.sentry.io/issues/5093043103/?project=6627632&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=0&utc=true

.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ defaults:
 jobs:
   e2e-test:
     if: github.repository_owner == 'getsentry'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
hubertdeng123, line 19: bumped this while at it

.github/workflows/test.yml
```diff
@@ -68,18 +68,19 @@ jobs:
...
         uses: actions/checkout@v4
 
       - name: Setup dev environment
         run: |
           pip install -r requirements-dev.txt
+          echo "PY_COLORS=1" >> "$GITHUB_ENV"
           ### pytest-sentry configuration ###
           if [ "$GITHUB_REPOSITORY" = "getsentry/self-hosted" ]; then
-            echo "PYTEST_SENTRY_DSN=${{ env.SELF_HOSTED_TESTING_DSN }}" >> $GITHUB_ENV
+            echo "PYTEST_SENTRY_DSN=$SELF_HOSTED_TESTING_DSN" >> $GITHUB_ENV
```
hubertdeng123, line 83: This was previously not being picked up correctly

_integration-test/conftest.py
```diff
@@ -10,27 +10,49 @@
...
             time.sleep(1)
         else:
             if response.status_code == 200:
                 break
     else:
         raise AssertionError("timeout waiting for self-hosted to come up")
 
+    if request.config.getoption("--customizations") == "True":
+        os.environ['TEST_CUSTOMIZATIONS'] = "True"
+        script_content = '''\
```
hubertdeng123, line 32: These lines are definitely covered by testing, not sure why Codecov is complaining here.

hubertdeng123: That's the reason why `Test / integration test (True, v2.0.1)` and `Test / integration test (True, v2.7.0)` are taking 8 minutes longer

.github/workflows/test.yml
azaslavsky: nit: Imo we should split this into a separate PR, since I could imagine cases where we want to only roll this bump back.

_integration-test/conftest.py
azaslavsky: nit: It's pretty hard for a naive observer to figure out what the "True" in these variants means. Maybe we could have values be `Customization Enabled` vs `Customization Disabled` rather than `True`/`False`?

_integration-test/test_run.py
```diff
@@ -171,3 +182,64 @@ def placeholder_fn():
...
+            "sentry-cleanup",
+            "python",
+            "-c",
+            "import ldap",
+        ]
+    ]
+    for command in commands:
+        result = subprocess.run(command, check=False)
+        if os.getenv("TEST_CUSTOMIZATIONS", "False") == "True":
+            assert result.returncode == 0
```
azaslavsky, line 243: Does `test_customization` run in both matrix variants?

azaslavsky (approved):

_integration-test/test_run.py
hubertdeng123: Yes, it does. If customizations enabled, it checks to ensure that they are truly enabled by making sure a dependency can be imported that is included in the custom image (ldap). If not, it checks to ensure that an error is thrown when trying to import a dependency that comes from the custom image.

_integration-test/conftest.py
hubertdeng123: Apparently, you can create a custom name for the job. Now it looks like
`Test / integration test v2.0.1 - customizations disabled`

PR #2919, opened 2024-03-22
bump zookeeper and kafka

### LINKED ISSUE #2304: Outdated images for Zookeeper and Kafka
### Problem Statement

In the docker-compose.yaml you use two images that are very old.
- confluentinc/cp-zookeeper:5.5.7
- confluentinc/cp-kafka:5.5.7

Both are based on Debian 8, which has been EOL for three years.

### Solution Brainstorm

Please update to a newer Version such as the tags 6.0.14 or 7.4.1.

thuvh:
Change zookeeper and kafka to 7.6.0
fix #2304


aldy505 (commented):
Can the upgrade process be done directly, or does it need some kind of data migration script that we have to execute?

thuvh:
I start with new install, but next week I will check with exist data.

thuvh:
I test by start with branch 24.3.0, then I upgrade kafka and zookeeper, I have problem like [here](https://github.com/confluentinc/kafka-images/issues/131). After adding "user:root" two zookeeper and kafka, I can start with no error. I think we can upgrade directly.

docker-compose.yml
```diff
@@ -171,6 +171,7 @@ services:
       nofile:
         soft: 4096
         hard: 4096
+    user: root
```
hubertdeng123, line 174: I am a bit wary of doing this, wondering if there is a better approach here

aldy505:
@hubertdeng123 can you confirm about data loss after upgrade (if there are any data loss)?

docker-compose.yml
thuvh: I think this problem of kafka image from confluence, this line likes a work around for upgrading

thuvh: have you got any idea @hubertdeng123

hubertdeng123:
@aldy505 Existing data will not be lost. If there are events that are coming in during the downtime of your instance, that will be lost since we operate on a single node but that is not any different than upgrading your self-hosted instance as of today.

docker-compose.yml
hubertdeng123: I was able to get around this with something like this:

```
chown -R 1000:1000 /var/lib/docker/volumes/sentry-zookeeper/_data
chown -R 1000:1000 /var/lib/docker/volumes/sentry-self-hosted_sentry-zookeeper-log/_data
```

I also ran `chown -R 1000:1000 /var/lib/docker/volumes/sentry-kafka/_data` but I am not sure if that is needed

hubertdeng123: Can you try that?

hubertdeng123: Given that an approach for this may require us to write a script of some sort to properly set the permissions of docker volumes, I think it is more reasonable for us to take on the brunt of this work in https://github.com/getsentry/self-hosted/pull/2988

thuvh: do you need a test?

thuvh: and if you think your commit is better, let me know to close this pr, thank!

PR #2927, opened 2024-04-01
feat: adds group attributes consumer

scefali:
We need to enable the group attributes consumer so we can write to this dataset in self-hosted


azaslavsky (approved):

scefali:
@aldy505 good question, the group attributes consumer is used to capture state changes on Sentry issues in Postgres and send them to ClickHouse. We can then query issue attributes (ex: the first time an issue was seen) in ClickHouse when users search for issues rather than having to query Postgres first then ClickHouse.

PR #2953, opened 2024-04-11
Use python for e2e tests

hubertdeng123:
Picks up https://github.com/getsentry/action-self-hosted-e2e-tests/pull/14


azaslavsky (commented):
Love to see that code deletion! 🎉

azaslavsky (approved):

PR #2955, opened 2024-04-12
Capture output for CI flakes

hubertdeng123:


hubertdeng123:
hm oops, I must've rebased things too far

PR #2959, opened 2024-04-15
Add example to docker compose version in problem report

edgariscoding:
Adds an example to get the version of docker compose. 

Since the command to get the version of **Docker** is `docker --version` you would expect the command for **Docker Compose** to be `docker compose --version` but for whatever reason it's not, so people probably waste time trying to figure out what the proper command is.

### Legal Boilerplate


BYK (approved):
Thanks for this!

hubertdeng123 (approved):

aldy505 (approved):

PR #2968, opened 2024-04-17
Add workstation configuration

azaslavsky:
These are prebuilt docker images for spinning up a local self-hosted image on the Google Cloud Workstation project. While primarily intended for internal development at Sentry, in theory these can be used by anyone with GCWS project to create a fresh workstation for developing self-hosted via a remote VSCode connection.

Users who have GCWS properly configured will be able to use the forthcoming `workstations ...` command in the `sentry` dev CLI to create, manage, and destroy one-off or long-lived workstations in either the pre-install or post-install configuration.

Note that the `sentry workstations ...` CLI has not yet landed in the `sentry` repo - those changes are coming soon!

Issue: getsentry/team-ospo#240


workstation/299_setup-completed.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+
+# Add a dot file to the home directory indicating that the setup has been completed successfully.
+# The host-side of the connection will look for this file when polling for completion to indicate to
```
hubertdeng123, line 5: For my understanding, where is this defined upstream? Is it some setting in GC workstation config?

workstation/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+## Creating and connecting to a workstation
+
+Once the Google Cloud services are configured and the docker images uploaded per the instructions
+above, end users should be able to list the configurations available to them using `sentry
+workstations config`:
+
+```shell
+$> sentry workstations configs --project=$GCP_PROJECT_ID
```
hubertdeng123, line 145: Is this in sentry dev? Or needs to be exec'd in the web container?

workstation/299_setup-completed.sh
azaslavsky: The `sentry workstations create` tool knows to look for this file specifically. That PR is incoming in the `sentry` repo.

workstation/README.md
hubertdeng123: Gotcha, so this is predominantly used as documentation

PR #2975, opened 2024-04-19
Mdtro/bump postgres

mdtro:
The `postgres:14.5` image has known vulnerabilities. This bumps us to `14.11-alpine` where the vulnerabilities are patched and moves us to a lightweight container. 

Sentry developer environments are currently on the alpine version of 14.9, but [should be on 14.11 soon.](https://github.com/getsentry/image-mirror/pull/20)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -132,7 +132,7 @@ services:
   postgres:
     <<: *restart_policy
     # Using the same postgres version as Sentry dev for consistency purposes
-    image: "postgres:14.5"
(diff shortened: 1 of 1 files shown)


aldy505 (approved):
But when will we upgrade to a higher major version of Postgres?

azaslavsky:
The postgres version is closely tied to what we use on prod, and we try to be quite conservative about major version dependency bumps.

azaslavsky (approved):

PR #2980, opened 2024-04-19
Bump docker compose version in CI

hubertdeng123:
Played around trying to find a straightforward way to resolve the flaky errors we were seeing in testing. For docker compose versions v2.7.0 and v2.0.1 wasn't able to find anything substantial. Turns out the flaky errors we were seeing are resolved in later versions of docker, so for now it seems reasonable to only rerun tests for self-hosted on docker compose v2.0.1.


.github/workflows/test.yml
```diff
@@ -110,7 +110,12 @@ jobs:
           command: ./install.sh
 
       - name: Integration Test
-        run: pytest --cov --junitxml=junit.xml --reruns 3 _integration-test/ --customizations=${{ matrix.customizations }}
+        run: |
+          if [ "${{ matrix.compose_version }}" = "v2.0.1" ]; then
+            pytest --reruns 5 --cov --junitxml=junit.xml _integration-test/ --customizations=${{ matrix.customizations }}
```
azaslavsky, line 115: Let's change this back to 3, otherwise looks good.

hubertdeng123: done

azaslavsky (approved):

PR #2988, opened 2024-04-22
Bump kafka and zookeeper versions

hubertdeng123:
let's try this getting this to 7.6.1 so we can get rid of zookeeper

more details: https://docs.confluent.io/platform/current/installation/migrate-zk-kraft.html


hubertdeng123:
tested it on our dogfood instance, seems to work

aldy505:
Dupe of #2919?

hubertdeng123:
> Dupe of https://github.com/getsentry/self-hosted/pull/2919?

Yes, but wanted an easier way of working locally with the changes. I'll merge the dupe before this one if we decide to go to 7.6.1, since that has the capability of getting rid of zookeeper

.github/workflows/test.yml
```diff
@@ -102,6 +102,9 @@ jobs:
           sudo curl -L https://github.com/docker/compose/releases/download/${{ matrix.compose_version }}/docker-compose-`uname -s`-`uname -m` -o "${{ matrix.compose_path }}/docker-compose"
           sudo chmod +x "${{ matrix.compose_path }}/docker-compose"
 
+      - name: Ensure runner is owner of /var/lib/docker to set permissions of docker volumes
```
hubertdeng123, line 105: Unfortunately, this is needed since the github runner does not have access to `/var/lib/docker/volumes` and I'd like to avoid invoking sudo in the install script

hubertdeng123: this is specific since we're performing a chmod now in the install script

aldy505:
_just use Redpanda already, lol_

_integration-test/conftest.py
```diff
@@ -52,7 +52,15 @@ def configure_self_hosted_environment(request):
         with open("sentry/requirements.txt", "w") as req_file:
             req_file.write("python-ldap\n")
         os.environ["MINIMIZE_DOWNTIME"] = "1"
-        subprocess.run(["./install.sh"], check=True, capture_output=True)
+        subprocess.run(
+            ["sudo", "chown", "-R", "$(whoami):$(whoami)", "/var/lib/docker"],
```
aldy505, line 56: Nooooo don't do this. Try add current `$(whoami)` to docker group instead

aldy505: Should be `usermod -aG docker $(whoami)`

hubertdeng123: giving this a shot

thuvh:
@hubertdeng123 well done!

install/update-docker-volume-permissions.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Only supporting platforms on linux x86 platforms and not apple silicon. I'm assuming that folks using apple silicon are doing it for dev purposes and it's difficult
+# to change permissions of docker volumes since it is run in a VM.
+if [[ "$DOCKER_PLATFORM" = "linux/amd64" && -n "$(docker volume ls -q -f name=sentry-zookeeper)" && -n "$(docker volume ls -q -f name=sentry-kafka)" ]]; then
+  zookeeper_data_dir="/var/lib/docker/volumes/sentry-zookeeper/_data"
+  kafka_data_dir="/var/lib/docker/volumes/sentry-kafka/_data"
+  zookeeper_log_data_dir="/var/lib/docker/volumes/sentry-self-hosted_sentry-zookeeper-log/_data"
+  chmod -R a+w $zookeeper_data_dir $kafka_data_dir $zookeeper_log_data_dir && returncode=$? || returncode=$?
+  if [[ $returncode == "1" ]]; then
+    echo "WARNING: Error when setting appropriate permissions for zookeeper, kafka, and zookeeper log docker volumes. This may corrupt your self-hosted install."
```
azaslavsky, line 11: Should we point to a bug or doc with more info? It seems like users could recover from this manually if this script fails for some reason.

hubertdeng123: linked to a bug to give users more context

azaslavsky (approved):

Issue #2970, opened 2024-04-18
INP not showing on self-hosted


amenk:
I am currently testing the new INP integration locally.

Versions:

* Version 122.0.6261.128 (Official Build) (64-bit)
* Sentry Browser Integration https://browser.sentry-cdn.com/7.110.0/bundle.tracing.min.js
* Sentry 24.3.0 self hosted, with the default features from the config example

Tracing sample rate is set to 1

I used

```
    Sentry.browserTracingIntegration({
        enableInp: true,
    }),
```

The web vitals view looks like this

![image](https://github.com/getsentry/sentry-javascript/assets/1087128/e77a0546-3192-42f2-b232-bc7f56fe4d36)

I did not deploy the enableInp yet to production, maybe it would work there with a greater user base?

I am mostly interested in the INP element - how would this show up? Can someone post an example how it should look like when it is working?

When exactly is the INP transmitted? I saw in https://github.com/getsentry/sentry-javascript/discussions/7765 that some caching might be in place?

_Originally posted by @amenk in https://github.com/getsentry/sentry-javascript/discussions/11591_

aldy505:
Can you try to add `"organizations:performance-vitals-inp"` to the feature flag on to the list of feature flags that's defined here? https://github.com/getsentry/self-hosted/blob/d59c0aa57c1432af9c5000007c17a87c07935a55/sentry/sentry.conf.example.py#L266-L312

Then re-run `./install.sh` script.

amenk:
@aldy505 Added ist and re-ran install.sh - still "no data" after browsing a bit on the page.
How to sort out if the browser is sending INP? I have the debug javascript package already included.

```
<script src="https://browser.sentry-cdn.com/7.110.0/debug.min.js" crossorigin="anonymous"></script>
```

aldy505:
Okay I've tested it on my instance. All you need to do:

1. Add another feature flag on `sentry/sentry.conf.py`
    ```py
    SENTRY_FEATURES.update(
      # ...
      "organizations:performance-vitals-inp",
      # ...
    )
    ```
2. Re-run `sudo ./install.sh && sudo docker compose up -d`
3. I'm using this bundle: `https://browser.sentry-cdn.com/7.111.0/bundle.tracing.replay.min.js`
4. On `Sentry.init`, add `enableInp: true`
    ```js
    Sentry.init({
      // ...
      integrations: [
        Sentry.browserTracingIntegration({
          enableInp: true
        })
      ]
    })
[...]

PR #2995, opened 2024-04-25
feat(performance): add INP and Request feature flag

aldy505:
Solves INP result not showing up on self-hosted. Also add the "Requests" page under "Performance" tab.

Linked issues: #2970

### Legal Boilerplate


hubertdeng123 (commented):
This may have been a problem in past versions of self-hosted, but the flags were just removed recently in Sentry. These features shouldn't be gated for the next release

PR #3009, opened 2024-05-01
Upgrade clickhouse to 23.8

hubertdeng123:
This PR includes a script that upgrades our users to CH 23.8, in this order:
21.8 -> 22.8 -> 23.3 -> 23.8


hubertdeng123:
Seems like I am encountering an issue where clickhouse immediately is booted with v23.8 instead of incrementally upgrading, which does not seem to be happening locally

install.sh
```diff
@@ -20,6 +20,7 @@ source install/check-latest-commit.sh
 source install/check-minimum-requirements.sh
 
 # Let's go! Start impacting things.
+source install/upgrade-clickhouse.sh
```
hubertdeng123, line 23: The reason this is placed before turn-things-off is because turn things off script will remove the clickhouse image, which we need to verify the version of clickhouse before proceeding

install/detect-platform.sh
```diff
@@ -20,10 +20,8 @@ fi
 export DOCKER_ARCH=$(docker info --format '{{.Architecture}}')
 if [[ "$DOCKER_ARCH" = "x86_64" ]]; then
   export DOCKER_PLATFORM="linux/amd64"
-  export CLICKHOUSE_IMAGE="altinity/clickhouse-server:21.8.13.1.altinitystable"
```
hubertdeng123, line 23: not needed, since more recent clickhouse images are multi platform

clickhouse/config.xml
```diff
@@ -1,10 +1,6 @@
 <yandex>
-    <max_server_memory_usage_to_ram_ratio>
-        <!-- This include is important!
-         It is required for the version of Clickhouse
-         used on ARM to read the environment variable. -->
-        <include from_env="MAX_MEMORY_USAGE_RATIO"/>
-    </max_server_memory_usage_to_ram_ratio>
+ <!-- This include is important! It is required for the version of Clickhouse used on ARM to read the environment variable. -->
```
hubertdeng123, line 2: For some reason, in more recent versions of clickhouse the newlines here are causing parsing issues, so I just removed them

install/upgrade-clickhouse.sh
```diff
@@ -0,0 +1,28 @@
+echo "${_group}Upgrading Clickhouse ..."
+
+# First check to see if user is upgrading by checking for existing clickhouse volume
+if [[ -n "$(docker volume ls -q --filter name=sentry-clickhouse)" ]]; then
```
azaslavsky, line 4: Wonder if it's worth adding a test for this? Should (hopefully) be pretty straightforward with the new python-based testing regime?

install.sh
azaslavsky: Let's add this as a comment in the code, since whoever edits this in the future will probably want to know this fact as well.

install/upgrade-clickhouse.sh
hubertdeng123: Editing the install script is not easily tested using the new python tests. Those are effective after self-hosted sentry is installed and up and running. The new self-hosted Sentry upgrade tests I added sufficiently covers this IMO, it skips over this portion of the code on initial install and hits this on the upgrade

azaslavsky (approved):

Issue #3030, opened 2024-05-07
chmod: cannot access '/var/lib/docker/volumes/sentry-self-hosted_sentry-zookeeper-log/_data


etlam:
### Self-Hosted Version

24.4.2

### CPU Architecture

x86_64

### Docker Version

24.0.7

### Docker Compose Version

2.21.0

### Steps to Reproduce

1. Create `.env.custom`
2. Change `COMPOSE_PROJECT_NAME` to any other value than the default value `sentry-self-hosted`
3. Run `sudo ./install.sh --skip-user-creation  --no-report-self-hosted-issues`

### Expected Result

Checking volume permissions for kafka and zookepper does not fail.  
Located in this file:  https://github.com/getsentry/self-hosted/blame/1c72fbe6120444ac232df0dae242cd5af40b38df/install/update-docker-volume-permissions.sh  

### Actual Result

```
Ensuring Kafka and Zookeeper volumes have correct permissions ...
chmod: cannot access '/var/lib/docker/volumes/sentry-zookeeper/_data': Permission denied
chmod: cannot access '/var/lib/docker/volumes/sentry-kafka/_data': Permission denied
chmod: cannot access '/var/lib/docker/volumes/sentry-self-hosted_sentry-zookeeper-log/_data': Permission denied
WARNING: Error when setting appropriate permissions for zookeeper, kafka, and zookeeper log docker volumes. This may corrupt your self-hosted install. See https://github.com/confluentinc/kafka-images/issues/127 for context on why this was added.
```  

Problem:
- Docker compose project name was changed
- This leads to other volume names, but they are hardcoded here:  
  -  https://github.com/getsentry/self-hosted/blob/1c72fbe6120444ac232df0dae242cd5af40b38df/install/update-docker-volume-permissions.sh#L6
  - https://github.com/getsentry/self-hosted/blob/1c72fbe6120444ac232df0dae242cd5af40b38df/install/update-docker-volume-permissions.sh#L7
  - https://github.com/getsentry/self-hosted/blob/1c72fbe6120444ac232df0dae242cd5af40b38df/install/update-docker-volume-permissions.sh#L8

azaslavsky:
Just out of curiosity: what is the use case for doing this? I agree that we've been a bit sloppy about hard coding vs using this variable appropriately, it'd be nice to know what sorts of applications folks are using this for.

etlam:
[...]

PR #3039, opened 2024-05-10
fix: Make docker volume script respect compose project name

hubertdeng123:
Close https://github.com/getsentry/self-hosted/issues/3030


azaslavsky (approved):

aldy505 (approved):
But wouldn't it be better if we check whether the directory exists on the first place?

hubertdeng123:
If the directory doesn't exist in the first place, we still want to throw the same error so behavior wise I think this should be fine

PR #3040, opened 2024-05-10
Remove no skip write flag for snuba consumers

hubertdeng123:
follows after https://github.com/getsentry/snuba/pull/5846


hubertdeng123:
closing since duplicate of https://github.com/getsentry/self-hosted/pull/3041

PR #3060, opened 2024-05-16
feat(feedback): create kafka topic and consumer for ingest-feedback

aliu39:
I recently migrated feedback off the ingest-events pipeline in all prod regions: see https://github.com/getsentry/sentry/issues/66100.

This isn't a pressing change for self-hosted, but without it, we can't clean up [this option](https://github.com/getsentry/sentry/blob/5a4e113c83b0a4581a7bb0fde67115de8e2f07df/src/sentry/options/defaults.py#L480) which we use like a FF [in relay](https://github.com/getsentry/relay/pull/3344).

Needed for https://github.com/getsentry/relay/pull/3604

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -177,8 +177,7 @@ services:
       - "sentry-secrets:/etc/zookeeper/secrets"
     healthcheck:
       <<: *healthcheck_defaults
-      test:
-        ["CMD-SHELL", 'echo "ruok" | nc -w 2 localhost 2181 | grep imok']
+      test: ["CMD-SHELL", 'echo "ruok" | nc -w 2 localhost 2181 | grep imok']
   kafka:
     <<: *restart_policy
     depends_on:
@@ -250,7 +249,8 @@ services:
     # Override the entrypoint in order to avoid using envvars for config.
     # Futz with settings so we can keep mmdb and conf in same dir on host
     # (image looks for them in separate dirs by default).
-    entrypoint: ["/usr/bin/geoipupdate", "-d", "/sentry", "-f", "/sentry/GeoIP.conf"]
+    entrypoint:
+      ["/usr/bin/geoipupdate", "-d", "/sentry", "-f", "/sentry/GeoIP.conf"]
     volumes:
       - "./geoip:/sentry"
   snuba-api:
@@ -363,6 +363,9 @@ services:
   transactions-consumer:
     <<: *sentry_defaults
     command: run consumer ingest-transactions --consumer-group ingest-consumer
+  feedback-events-consumer:
+    <<: *sentry_defaults
+    command: run consumer ingest-feedback-events --consumer-group ingest-feedback-events
   metrics-consumer:
     <<: *sentry_defaults
     command: run consumer ingest-metrics --consumer-group metrics-consumer
--- a/install/create-kafka-topics.sh
+++ b/install/create-kafka-topics.sh
@@ -14,7 +14,7 @@ done
 
 # XXX(BYK): We cannot use auto.create.topics as Confluence and Apache hates it now (and makes it very hard to enable)
 EXISTING_KAFKA_TOPICS=$($dc exec -T kafka kafka-topics --list --bootstrap-server kafka:9092 2>/dev/null)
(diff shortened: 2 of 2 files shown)


aliu3ntry:
Code refs for the consumer:
https://github.com/getsentry/sentry/blob/3e8633166fed0340a2d007399d156232d023ca5e/src/sentry/consumers/__init__.py#L284
https://github.com/getsentry/ops/blob/3323d813b6c89ce631f294dad0a9beb14f7ca1a5/single-tenant/k8s/sentry/_values.yaml#L613

docker-compose.yml
```diff
@@ -250,7 +249,8 @@ services:
     # Override the entrypoint in order to avoid using envvars for config.
     # Futz with settings so we can keep mmdb and conf in same dir on host
     # (image looks for them in separate dirs by default).
-    entrypoint: ["/usr/bin/geoipupdate", "-d", "/sentry", "-f", "/sentry/GeoIP.conf"]
```
aliu3ntry, line 253: These were from my vscode formatter..

docker-compose.yml
```diff
@@ -363,6 +363,9 @@ services:
   transactions-consumer:
     <<: *sentry_defaults
     command: run consumer ingest-transactions --consumer-group ingest-consumer
+  ingest-feedback-events:
```
hubertdeng123, line 366: nit: could you change this to `feedback-events-consumer`? It matches the naming convention of the other ingest consumers

docker-compose.yml
```diff
@@ -363,6 +363,9 @@ services:
...
+    <<: *sentry_defaults
+    command: run consumer ingest-feedback-events --consumer-group ingest-consumer
```
hubertdeng123, line 368: Should we use `ingest-feedback-events` as the consumer group instead?

docker-compose.yml
aliu3ntry: Sure, should it be changed in create-kafka-topics.sh too? It's named ingest-feedback-events in prod so I was worried bout keeping consistent with that

docker-compose.yml
aliu3ntry: Does this consumer group exist for self-hosted? I was following the convention of the other ingest consumers in this file

aliu3ntry: Will the command create the group?

docker-compose.yml
hubertdeng123: Nope, this is just the name of the service. We need the topic to stay as `ingest-feedback-events` since that is the topic being published/consumed from in prod

aliu3ntry: Ok so only change line 366 to feedback-events-consumer?

docker-compose.yml
hubertdeng123: Other ingest consumers in this file also have different consumer groups. The reason why transactions, attachments, and events all have the same consumer group is because they used to be combined, but were split in https://github.com/getsentry/self-hosted/pull/2193/files

docker-compose.yml
hubertdeng123: yep!

aliu3ntry:
My bad, lost the commit history reverting some stuff, I thought I was on another branch..

hubertdeng123 (approved):
This looks good to me. Have you verified that a user feedback event is ingested properly yet in self-hosted? I can help out with that if not.

aliu3ntry:
> This looks good to me. Have you verified that a user feedback event is ingested properly yet in self-hosted? I can help out with that if not.

No, think we'll need https://github.com/getsentry/self-hosted/pull/3061 for that. I'll merge and test it tomorrow!

hubertdeng123:
@aliu3ntry You should be able to test your changes out before merging

aliu3ntry:
Had trouble ingesting feedback in self-hosted instance due to CORS error / msg being dropped by proxy before it could reach the relay processor. Tried debugging and discussed with @hubertdeng123. It'd be nice to get this working so we can enable feedback by default. But due to other commitments, we'll come back to it later.

aliu39:
Moving to https://github.com/getsentry/self-hosted/pull/3193

PR #3061, opened 2024-05-16
feat(feedback): toggle option for relay to produce to ingest-feedback

aliu3ntry:
Follow-up to (merge after) https://github.com/getsentry/self-hosted/pull/3060

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -134,6 +134,9 @@ SENTRY_EVENTSTREAM_OPTIONS = {"producer_configuration": DEFAULT_KAFKA_OPTIONS}
 
 KAFKA_CLUSTERS["default"] = DEFAULT_KAFKA_OPTIONS
 
+# Temporary flag to test a new ingest consumer for user feedback.
+SENTRY_OPTIONS["feedback.ingest-topic.rollout-rate"] = 1.0
+
 ###############
 # Rate Limits #
 ###############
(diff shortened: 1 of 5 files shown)


sentry/sentry.conf.example.py
```diff
@@ -134,6 +134,9 @@ def get_internal_network():
 
 KAFKA_CLUSTERS["default"] = DEFAULT_KAFKA_OPTIONS
 
+# Temporary flag to test a new ingest consumer for user feedback.
+SENTRY_OPTIONS["feedback.ingest-topic.rollout-rate"] = 1.0
```
hubertdeng123, line 138: Will the default value of this option be changed in sentry? How will we know it is time to remove this temporary flag?

aliu3ntry: The default value is 0.0 and will not be changed

aliu3ntry: As soon as this ingest topic is tested and confirmed to work in self-hosted, I will remove references to the flag in all repos.

hubertdeng123: Got it, so essentially the option will not be needed in the future after the references to the flag is removed

aliu3ntry: Yup!

aliu3ntry:
circling back later, see comment on https://github.com/getsentry/self-hosted/pull/3060

PR #3069, opened 2024-05-21
task: add check if install.sh was run with root privileges

LvckyAPI:
## Story

Currently, it's possible to run the `install.sh` without root privileges.  
This could cause misunderstandings if errors are thrown while running the script without root.

I added a check, if the script was run with root privileges.

## Legal Boilerplate

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -8,10 +8,19 @@ if [[ -n "$MSYSTEM" ]]; then
 fi
 
 # Check if script was executed with root/sudo
-if [ `id -u` -ne 0 ]
-  then echo Please run this script as root or using sudo!
-  exit
+if [ "$(id -u)" -ne 0 ]; then
+    if [ "$CI" != "true" ]; then
+        echo "You should be using sudo, are you sure you want to continue? [y/n]"
+        read -r answer
+        if [ "$answer" != "${answer#[Yy]}" ] ;then
+            echo "Continuing without sudo. This might cause issues later on."
+        else
+            echo "The script execution was stopped. Please run it again with sudo."
+            exit
+        fi
+    fi
 fi
+
 
 source install/_lib.sh
 
--- a/install.sh
+++ b/install.sh
@@ -7,6 +7,12 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+# Check if script was executed with root/sudo
+if [ `id -u` -ne 0 ]
+  then echo Please run this script as root or using sudo!
+  exit
+fi
+
 source install/_lib.sh
 
 # Pre-flight. No impact yet.
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ fi
 # Check if script was executed with root/sudo
 if [ "$(id -u)" -ne 0 ]; then
     if [ "$CI" != "true" ]; then
-        echo "You should be using sudo, are you sure you want to continue? [y/n]"
+        echo "It is recommended to run the script as the root user (by using sudo or as a direct root user) to avoid permission issues. Are you sure you want to continue? [y/n]"
         read -r answer
         if [ "$answer" != "${answer#[Yy]}" ] ;then
             echo "Continuing without sudo. This might cause issues later on."


LvckyAPI:
@hubertdeng123

install.sh
```diff
@@ -7,6 +7,12 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+# Check if script was executed with root/sudo
+if [ `id -u` -ne 0 ]
+  then echo Please run this script as root or using sudo!
+  exit
+fi
+
```
aldy505, line 15: You should add another condition if there is no `CI=true` environment variable, then this check would be valid.

aldy505 (commented):
I'm not against this. In fact, I would recommend this since regular Docker installation requires higher privileges than regular Linux user. But, what should we do with people that don't have root access and currently are running sentry self-hosted normally? Should we add additional environment variable or a CLI flag that states they're aware about what they're doing?

So, it would be something like:
```bash
./install.sh --explicit-no-root-access
# or
EXPLICIT_NO_ROOT_ACCESS=true ./install.sh
```

BYK:
I'd just print a warning. I don't recall root being required from the old times 😅 If it is required we should also note in the docs too.

install.sh
LvckyAPI: do you want to have it ac environment variable or as additional command parameter?

LvckyAPI:
> I'd just print a warning. I don't recall root being required from the old times 😅 If it is required we should also note in the docs too.

but in the most cases the installation fails if you don't run the script with root
For example today in Discord: https://discord.com/channels/621778831602221064/1242216895327899808

BYK:
Okay, turns out we do say you should use sudo here: https://github.com/getsentry/self-hosted?tab=readme-ov-file#using-linux

I'm +1 to the idea where we stop unless specifically overriden by a flag or env variable. The stopping can be a prompt too btw (you should be using sudo, are you sure you wanna continue?). Finally don't forget the CI ejv variable handling as mentioned earlier.

install.sh
aldy505: `CI=true` should be from environment variable. It's for the CI to be able to pass.

LvckyAPI:
> Okay, turns out we do say you should use sudo here: https://github.com/getsentry/self-hosted?tab=readme-ov-file#using-linux
> 
> I'm +1 to the idea where we stop unless specifically overriden by a flag or env variable. The stopping can be a prompt too btw (you should be using sudo, are you sure you wanna continue?). Finally don't forget the CI env variable handling as mentioned earlier.

the idea with the prompt is great too

install.sh
LvckyAPI: okay

LvckyAPI:
I've added a check, if the environment variable `CI` is set to true, and I added a user prompt to ignor this warning

install.sh
```diff
@@ -7,6 +7,20 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+# Check if script was executed with root/sudo
+if [ "$CI" != "true" ]; then
+  if [ "$(id -u)" -ne 0 ]; then
+    echo "You should be using sudo, are you sure you want to continue? [y/n]"
+    read -r answer
+    if [ "$answer" != "${answer#[Yy]}" ] ;then
```
BYK, line 15: Not sure if in misinterpreting this but is the logic flipped here?

aminvakil (commented):
We are running self-hosted without `sudo` without any problem, also `git grep sudo` does not show any command which needs `sudo`, although once it has run `install.sh` with `sudo`, it probably needs to get run with `sudo` next times as well, but `sudo` is not needed AFAICS.

> Okay, turns out we do say you should use sudo here: https://github.com/getsentry/self-hosted?tab=readme-ov-file#using-linux

@BYK This just states that **if** you are using an environment variable and `sudo` together, put `sudo` before environment setting.

install.sh
LvckyAPI: i flipped it

LvckyAPI:
> We are running self-hosted without `sudo` without any problem

There is a new script https://github.com/getsentry/self-hosted/blob/master/install/update-docker-volume-permissions.sh#L9 which is updating permission, and this won't work without root privileges or Permissions to this Folder.

hubertdeng123:
Root is not required to run the install script.

> but in the most cases the installation fails if you don't run the script with root
For example today in Discord: https://discord.com/channels/621778831602221064/1242216895327899808

There should've been a warning printed out during the install script that the docker volume permissions were not updated properly. I believe that is the only part of the install script that may fail if permissions are not granted to the user. Is that what your install was also failing on @LvckyAPI?

LvckyAPI:
> Is that what your install was also failing on @LvckyAPI?

Yeah, on every three instances that we run.

Before that we could run the install script relatively easily even without root, but not since then. And since it's documented everywhere that you should use sudo I thought, so that you don't spend ages looking for the problem and opening a ticket again and again, just force sudo and everything will work.

install.sh
```diff
@@ -7,6 +7,21 @@ if [[ -n "$MSYSTEM" ]]; then
   exit 1
 fi
 
+# Check if script was executed with root/sudo
+if [ "$(id -u)" -ne 0 ]; then
+    if [ "$CI" != "true" ]; then
+        echo "You should be using sudo, are you sure you want to continue? [y/n]"
```
aldy505, line 13: I would prefer to use "It's recommended to use.." rather than "You should be using.."

```suggestion
        echo "It is recommended to run the script as the root user (by using sudo or as a direct root user) to avoid permission issues. Are you sure you want to continue? [y/n]"
```

install.sh
```diff
@@ -7,6 +7,21 @@ if [[ -n "$MSYSTEM" ]]; then
...
+        read -r answer
+        if [ "$answer" != "${answer#[Yy]}" ] ;then
+            echo "Continuing without sudo. This might cause issues later on."
```
aldy505, line 16: I'm thinking of adding something along the lines of "You should know what you're doing as this might cause issues later on". But that might sound a bit demeaning for some people.

LvckyAPI: hmm, into line 16?

aldy505: Yeah but I don't think that's proper 😂

LvckyAPI: `You need to know what you're doing, because this could cause issues later on.` ?

aldy505:
> > I would prefer to use "It's recommended to use.." rather than "You should be using.."
> 
> @aldy505 On the contrary we should not recommend to use root IMO. you should be in docker group to use docker anyway, and you will be fine with that to make all changes in docker containers, volumes, and so on.

I'm thinking of having self-hosted as something that would be easier to set up for new people coming in. That's the reason on why I prefer that way. Stating to run `sudo ./install.sh` on the frontpage would also (hopefully) says that "you shouldn't really use rootless Docker on this one".

> > Root is not required to run the install script.
> > > but in the most cases the installation fails if you don't run the script with root
> > > For example today in Discord: https://discord.com/channels/621778831602221064/1242216895327899808
> > 
> > 
> > There should've been a warning printed out during the install script that the docker volume permissions were not updated properly. I believe that is the only part of the install script that may fail if permissions are not granted to the user. Is that what your install was also failing on @LvckyAPI?
> 
> @hubertdeng123 I think we should run a container with root user, mounting this volumes, and change permissions with that container, instead of using root user on host.

+1 I'd like to see this too.

BYK:
I'm with @aminvakil here:

>@hubertdeng123 I think we should run a container with root user, mounting this volumes, and change permissions with that container, instead of using root user on host

Much better solution and avoid the need for sudo, which is always a good thing.

hubertdeng123:
> @hubertdeng123 I think we should run a container with root user, mounting this volumes, and change permissions with that container, instead of using root user on host

This does sound like a better approach, I can make a PR for this

hubertdeng123:
https://github.com/getsentry/self-hosted/pull/3084 should fix this, so we don't need to run install script with root privileges

PR #3087, opened 2024-05-28
CI: Add upgrade test for last hard stop

hubertdeng123:
This will help catch issues that happen during the upgrade from the latest hard stop to the most current Sentry

### DIFF
--- a/install/upgrade-clickhouse.sh
+++ b/install/upgrade-clickhouse.sh
@@ -17,8 +17,21 @@ if [[ -n "$(docker volume ls -q --filter name=sentry-clickhouse)" ]]; then
   # Wait for clickhouse
   wait_for_clickhouse
 
-  # In order to get to 23.8, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8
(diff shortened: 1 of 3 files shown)


install/upgrade-clickhouse.sh
```diff
@@ -17,8 +17,21 @@ if [[ -n "$(docker volume ls -q --filter name=sentry-clickhouse)" ]]; then
   # Wait for clickhouse
   wait_for_clickhouse
 
-  # In order to get to 23.8, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8
+  # In order to get to 23.8, we need to first upgrade go from 20.3 -> 21.8 -> 22.8 -> 23.3 -> 23.8
```
BYK, line 20: Feels like it might be time to codify this information in a machine-readable format which would allow doing some nice things programmatically. WDYT?

aldy505:
@hubertdeng123 do you still wanna work on this one? I'd prefer to make a CI test for upgrades from the latest hard stop to the current release.

PR #3092, opened 2024-05-29
ref(spans): Add new feature flags needed

phacops:
We split the feature flags for span ingestion in 3 different buckets:
- indexed spans
- metrics extraction for some modules
- metrics extraction for add-on modules

This PR will make sure all 3 features are now enabled in self-hosted.

PR #3094, opened 2024-05-29
fix: Use env custom file if it exists for tests

hubertdeng123:
Tests in relay/snuba likely weren't being properly run with the pre-submit images that were built. This led to an event today where a change led to self-hosted e2e test breakage in relay on master that was not caught on the [PR](https://github.com/getsentry/relay/pull/3633). I believe this should fix that issue. I suspect the custom env file for every commit on a PR wasn't being used for e2e tests that is built here: https://github.com/getsentry/action-self-hosted-e2e-tests/blob/main/action.yml#L64


hubertdeng123:
CI failing as it is waiting for https://github.com/getsentry/self-hosted/pull/3092

BYK:
Just checking, what's the state of this PR? The changes don't look related to the env at all at this point?

Issue #3050, opened 2024-05-13
No Sentry DSN link - link invisible / non present


lcsvcn:
### Self-Hosted Version

Sentry 24.4.2

### CPU Architecture

x86_64

### Docker Version

Docker version 26.1.2, build 211e74b

### Docker Compose Version

Docker Compose version v2.27.0

### Steps to Reproduce

1. Install application
2. Create project

### Expected Result

DSN would appear.

### Actual Result

NO DSN, see image below:
<img width="912" alt="Screenshot 2024-05-13 at 17 30 03" src="https://github.com/getsentry/self-hosted/assets/6011385/389e3ddb-e512-495b-83d6-e83eb338452b">

Delete and recreate key doesnt work

lcsvcn:
Security Header Endpoint and other below appear like:

/api/5/security/?sentry_key=SECRET_KEY

lcsvcn:
creating any new project looks like this:
<img width="1108" alt="Screenshot 2024-05-13 at 17 37 33" src="https://github.com/getsentry/self-hosted/assets/6011385/02131b4b-fc79-458e-a14a-31f21b1f90e7">

aldy505:
Hello, this is a common issue that probably we should address this more clearly.

You should set the `system.url-prefix` to the domain that your Sentry instance is most accessible from:

https://github.com/getsentry/self-hosted/blob/9de4b70ece8b28d83d4b2e4a4b642ff14af52ae0/sentry/config.example.yml#L81-L82

lcsvcn:
ok, thanks that worked!

Issue #2621, opened 2023-12-02
New Project, DSN Empty


zystudios:
### Environment

self-hosted (https://develop.sentry.dev/self-hosted/)

### Steps to Reproduce

![image](https://github.com/getsentry/sentry/assets/29728228/a8f8b299-7ab2-4953-b815-4e017d2ad1ef)

![image](https://github.com/getsentry/sentry/assets/29728228/01cfed8d-0703-48a8-92a9-4c1ba5231f05)

I deployed sentry on my own intranet, I create a new project and the dsn is empty no matter what type of project it is, I don't know how to fix it, thanks!

version Sentry 23.8.0.dev0 f866963

### Expected Result

DSN not empty

### Actual Result

DSN empty

### Product Area

Settings - Projects

aldy505:
You'd need to uncomment this and fill it with your base URL.

https://github.com/getsentry/self-hosted/blob/76ab9a596e7dc32c468f387dd45ba637e8cfad53/sentry/config.example.yml#L82

hubertdeng123:
Technically, `system.url-prefix` should be set on the initialization of a Sentry instance on the setup page, but if not you'll need to do what @aldy505 has mentioned. Does this error happen for every project you create?

zystudios:
Thanks for the replies, it's true that   `system.url-prefix` is empty

Here's how I solved it

1 Go to <sentry_url>/manage/settings/

2 Change the Root Url, I have set it before, change it to a different one and change it back.

zystudios:
I have a suggestion, if the system detects that the  `system.url-prefix`  is empty, it should give a hint instead of letting the DSN display empty

This is very important, and many people will not know what's going on

PR #3100, opened 2024-06-01
chore: clearly state system.url-prefix value should be set

aldy505:
There are more than one cases of people having their DSN or email links empty, most of it because `system.url-prefix` is not set on the config. Hope this will solves that (by providing a clearer message). But, I guess we should add this one to the docs site too, regarding "which config file you should be modifying before you run your sentry".

### Legal Boilerplate


BYK:
This should be handled by the initial set up screen, that's why we didn't ask it to be set in the config.

Agree with docs update but disagree with this specific change. If there's an issue with the set-up flow we should address that instead (on get sentry/sentry side)

aldy505:
> This should be handled by the initial set up screen, that's why we didn't ask it to be set in the config.
> 
> Agree with docs update but disagree with this specific change. If there's an issue with the set-up flow we should address that instead (on get sentry/sentry side)

I'm quite pessimistic on having that changed on the sentry/sentry side, since this is a common issue that most people encountered during their first time self-hosting sentry.

BYK:
I'm mostly trying to understand how this is not being set in the initial setup. Do we have any links to people asking about this so I can check?

aldy505:
> Do we have any links to people asking about this so I can check?

A few that I remembered:
- https://github.com/getsentry/self-hosted/issues/3050
- https://github.com/getsentry/self-hosted/issues/2621

aldy505:
> Sorry for the late turnaround here. How about we try to do //github.com/[/issues/2621](https://github.com/getsentry/self-hosted/issues/2621)#issuecomment-1842014431 and the subsequent comment instead? Hard-coding makes changing this a bit more tricky (although I agree that it should not be something that is changed regularly)

I don't really get what you're trying to do here.. since I believe modifying through the web app is something that's not persisted on disk (to postgres or file). But again, I'm not so sure about that, hence I prefer it to be done via configuration file.

BYK:
> since I believe modifying through the web app is something that's not persisted on disk (to postgres or file). 

It should persist on Postgres, or somewhere persistent as that's how the options system works. I'll list the relevant code for reference:

1. https://github.com/getsentry/sentry/blob/accca8f98c4231dae9a484c8985da43e7a6c4645/src/sentry/options/defaults.py#L63
2. https://github.com/getsentry/sentry/blob/accca8f98c4231dae9a484c8985da43e7a6c4645/src/sentry/options/manager.py#L304
3. https://github.com/getsentry/sentry/blob/accca8f98c4231dae9a484c8985da43e7a6c4645/static/app/views/admin/adminSettings.spec.tsx#L15

Now in number 1 and 3, you can see that it will be disabled on the UI when it was set from the disk (options file). That said I cought something quite interesting: looks like we fetch the default value from `SENTRY_SYSTEM_URL_PREFIX` so I'd now encourage setting that (in `.env` maybe) and documenting it.

WDYT?

aldy505:
> > since I believe modifying through the web app is something that's not persisted on disk (to postgres or file).
> 
> It should persist on Postgres, or somewhere persistent as that's how the options system works. I'll list the relevant code for reference:
> 
> 1. https://github.com/getsentry/sentry/blob/accca8f98c4231dae9a484c8985da43e7a6c4645/src/sentry/options/defaults.py#L63
> 2. https://github.com/getsentry/sentry/blob/accca8f98c4231dae9a484c8985da43e7a6c4645/src/sentry/options/manager.py#L304
> 3. https://github.com/getsentry/sentry/blob/accca8f98c4231dae9a484c8985da43e7a6c4645/static/app/views/admin/adminSettings.spec.tsx#L15
> 
> Now in number 1 and 3, you can see that it will be disabled on the UI when it was set from the disk (options file). That said I cought something quite interesting: looks like we fetch the default value from `SENTRY_SYSTEM_URL_PREFIX` so I'd now encourage setting that (in `.env` maybe) and documenting it.
> 
> WDYT?

Putting on the `.env` file might work. Seeing that, do you still prefer setting it on the `.env` file for visibility, or we'll just leave the additional comment for the configuration file?

@hubertdeng123 @azaslavsky what's your thoughts about this?

BYK (approved):
LGTM

hubertdeng123 (commented):
Have you tested the setting `SENTRY_SYSTEM_URL_PREFIX` will work after registering the option `system.url-prefix`? My concern here is someone manages to register the option for `system.url-prefix` changing the `SENTRY_SYSTEM_URL_PREFIX` will not do anything and it will instead cause confusion.

hubertdeng123:
I too echo what @BYK  has mentioned above. The option should be set on the initial set up screen. Personally I have not seen DSN appearing as empty as a reproducible issue and setting the `system.url-prefix` is sort of a hacky workaround. If we are to fix this I'd prefer to understand the root cause instead of covering this up even more

PR #3106, opened 2024-06-03
feat: Add crons task consumers 

wedamija:
We now process tasks via Kafka consumers instead of celerybeat. This needs to be added to self-hosted as well

### DIFF
--- a/install/create-kafka-topics.sh
+++ b/install/create-kafka-topics.sh
@@ -14,12 +14,16 @@ done
 
 # XXX(BYK): We cannot use auto.create.topics as Confluence and Apache hates it now (and makes it very hard to enable)
 EXISTING_KAFKA_TOPICS=$($dc exec -T kafka kafka-topics --list --bootstrap-server kafka:9092 2>/dev/null)
-NEEDED_KAFKA_TOPICS="ingest-attachments ingest-transactions ingest-events ingest-replay-recordings profiles ingest-occurrences ingest-metrics ingest-performance-metrics ingest-monitors"
+NEEDED_KAFKA_TOPICS="ingest-attachments ingest-transactions ingest-events ingest-replay-recordings profiles ingest-occurrences ingest-metrics ingest-performance-metrics ingest-monitors monitors-clock-tasks"
 for topic in $NEEDED_KAFKA_TOPICS; do
   if ! echo "$EXISTING_KAFKA_TOPICS" | grep -qE "(^| )$topic( |$)"; then
     $dc exec kafka kafka-topics --create --topic $topic --bootstrap-server kafka:9092
     echo ""
   fi
 done
+
+# This topic must have only a single partition for the consumer to work correctly
+# https://github.com/getsentry/ops/blob/7dbc26f39c584ec924c8fef2ad5c532d6a158be3/k8s/clusters/us/_topicctl.yaml#L288-L295
+$dc exec kafka kafka-topics --create --topic monitors-clock-tick --bootstrap-server kafka:9092 --partitions 1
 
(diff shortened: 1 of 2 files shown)


evanpurkhiser (commented):
Looks good to me. Will let oss team look to approve

install/create-kafka-topics.sh
```diff
@@ -14,12 +14,16 @@ done
...
 for topic in $NEEDED_KAFKA_TOPICS; do
   if ! echo "$EXISTING_KAFKA_TOPICS" | grep -qE "(^| )$topic( |$)"; then
     $dc exec kafka kafka-topics --create --topic $topic --bootstrap-server kafka:9092
     echo ""
   fi
 done
 
+# This topic must have only a single partition for the consumer to work correctly
+# https://github.com/getsentry/ops/blob/7dbc26f39c584ec924c8fef2ad5c532d6a158be3/k8s/clusters/us/_topicctl.yaml#L288-L295
+$dc exec kafka kafka-topics --create --topic monitors-clock-tick --bootstrap-server kafka:9092 --partitions 1
```
hubertdeng123, line 27: Should we check to see if topic already exists here before trying to create?

hubertdeng123 (commented):
Recently we've rewritten integration tests for self-hosted in python so hopefully tests are much easier to write. How difficult would it be to add a test to make sure this change is working as intended?

wedamija:
> Recently we've rewritten integration tests for self-hosted in python so hopefully tests are much easier to write. How difficult would it be to add a test to make sure this change is working as intended?

I'm not totally sure if it's worth writing an integration test here. The code is well tested in Sentry, and it feels like we'd just be duplicating work. It feels like to avoid these kinds of problems it'd be more helpful to figure out how to share these kinds of configs with how things work in prod, or maybe have a way to signal that certain consumers should always be run in self hosted?

hubertdeng123:
> It feels like to avoid these kinds of problems it'd be more helpful to figure out how to share these kinds of configs with how things work in prod, or maybe have a way to signal that certain consumers should always be run in self hosted?

What was the process for rolling this out to single tenant/prod? Ideally, maybe the way to go about this is to have features link to required consumers/infrastructure components and throw a warning or error in tests when pieces of infrastructure are not enabled for self-hosted?

hubertdeng123 (approved):

install/create-kafka-topics.sh
```diff
@@ -14,12 +14,19 @@ done
...
     echo ""
   fi
 done
 
+# This topic must have only a single partition for the consumer to work correctly
+# https://github.com/getsentry/ops/blob/7dbc26f39c584ec924c8fef2ad5c532d6a158be3/k8s/clusters/us/_topicctl.yaml#L288-L295
+
+if ! echo "$EXISTING_KAFKA_TOPICS" | grep -qE "(^| )monitors-clock-tick( |$)"; then
+  $dc exec kafka kafka-topics --create --topic monitors-clock-tick --bootstrap-server kafka:9092 --partitions 1
+fi
```
aldy505, line 30: Question: What would happen if someone accidentally created more than 1 partition for this exact topic?

wedamija: It would cause inconsistencies in the ordering of tasks and likely cause weird behaviour. If you've done so I'd recommend repartitioning to have a single partition

BYK:
Is it also time to consider a unified Sentry consumer at this point?

wedamija:
> Is it also time to consider a unified Sentry consumer at this point?

There's a project in the works to move a lot of celery over to Kafka, so this is being considered

Issue #3107, opened 2024-06-04
Upgrading to 24.5.0 breaks PostgreSQL indexes and causes upgrade to fail sometimes


afaianswq:
### Self-Hosted Version

24.5.0

### CPU Architecture

x86_64

### Docker Version

26.1.3

### Docker Compose Version

2.27.0

### Steps to Reproduce

1, Upgrade Sentry from version 24.4.1 or below to 24.4.2 or above

### Expected Result

Sentry successfully upgraded

### Actual Result

Upgrade either fails (see https://github.com/getsentry/self-hosted/issues/3027) or you'll get PostgreSQL indexes broken which will cause other issues after the upgrade.

The reason for this is because in version 24.4.2 image for postgres was changed from "postgres:14.5" (debian based, classic glibc) to "postgres:14.11-alpine" (alpine, musl libc). PostgreSQL is known to have issues with indexes when libc version or implementation changes. Please refer to this documentation article: https://wiki.postgresql.org/wiki/Locale_data_changes#What_to_do

azaslavsky:
Postgres was bumped to 14.11 for security reasons. You'll need to re-index as described in #3027

azaslavsky:
The other options is to change `postgres:14.11-alpine` to `postgres:14.11`. If this works, we may just change the image going forward.

afaianswq:
Both options are workarounds, not a real solution. This is a BUG and you need to fix it. For example, a proper solution would be to change postgres image back to default (debian-based, i.e. `postgres:14.11`). Because otherwise people will have to change it every time they upgrade their installation. Alternatively, if you want to keep alpine version, you have to figure out ALL affected indexes and add a migration step that will reindex all of them before any other database-related tasks are executed.

afaianswq:
> Postgres was bumped to 14.11 for security reasons.

Also, I just want to make it clear one more time, the problem is not postgres version. Problem is that you have changed from debian-based image to alpine-based image.

azaslavsky:
Does using `postgres:14.11` work for you? I want to get reports from the field before changing the version. Using this workaround will help us decide how to proceed.

afaianswq:
> Does using `postgres:14.11` work for you?

Yes, it does.

hubertdeng123:
I have opened a PR to change this. Thanks for following up here @afaianswq

Issue #3115, opened 2024-05-27
crons: same timeline is shown twice


JeremiaAu:
### Self-Hosted Version

24.5.0

### CPU Architecture

x86_64

### Docker Version

26.1.3

### Docker Compose Version

2.27.0

### Steps to Reproduce

1. Upgrade from version [24.4.1](https://github.com/getsentry/self-hosted/releases/tag/24.4.1) to [24.5.0](https://github.com/getsentry/self-hosted/releases/tag/24.5.0)
2. (Wait for new check-ins)
3. Open Crons

### Expected Result

Only one row per environment

### Actual Result

![image](https://github.com/getsentry/self-hosted/assets/96953097/695bd732-0283-482b-b785-e4a5fa356c25)

hubertdeng123:
I wonder if the reason behind this corruption with the unique constraint is because in between 24.4.1 and 24.5.0 we upgraded postgres to 14, and we started using the alpine image instead. Changing the OS might have lead to some issues here. I wonder if cleaning the duplicates up and then perhaps using the postgres:14 image instead might work?

JeremiaAu:
I also think that the postgres issue mentioned in https://github.com/getsentry/self-hosted/issues/3107 is to blame.

I ran the following command, meant to identify broken indices, and the `sentry_environment_organization_id_name_95a37dc7_uniq` showed up.

```
SELECT DISTINCT indrelid::regclass::text, indexrelid::regclass::text, collname, pg_get_indexdef(indexrelid) 
FROM (SELECT indexrelid, indrelid, indcollation[i] coll FROM pg_index, generate_subscripts(indcollation, 1) g(i)) s 
  JOIN pg_collation c ON coll=c.oid
WHERE collprovider IN ('d', 'c') AND collname NOT IN ('C', 'POSIX');
```
source: https://wiki.postgresql.org/wiki/Locale_data_changes#What_to_do

hubertdeng123:
Yep, I have a feeling that is the case too. We've changed the postgres image used back to a debian based image [here](https://github.com/getsentry/self-hosted/pull/3116). Do you happen to have a backup of your postgres data before you upgraded? If so, depending on your needs it may be better to restore that data. 

Otherwise, I think there are a few roads forward from here. It may be a good idea to perform a backup proceeding.

[...]

PR #3116, opened 2024-06-06
Use non-alpine postgres

hubertdeng123:
should resolve https://github.com/getsentry/self-hosted/issues/3107 and perhaps https://github.com/getsentry/self-hosted/issues/3115 is an issue that came from switching the OS.


azaslavsky:
LGTM, but going to loop in @mdtro to ensure that this still maintains the security properties that necessitated the bump in the first place.

aldy505 (approved):
Do we need to add additional re-indexing command for this one?

mdtro (approved):

PR #3121, opened 2024-06-07
Use general kafka topic creation in self-hosted

hubertdeng123:
Follow up to https://github.com/getsentry/sentry/pull/72338. We can now handle topic creation automatically versus manually listing out topics needed.


BYK:
Nice, love it! A bit weird that this flag is not enabled by default though.

hubertdeng123:
> Nice, love it! A bit weird that this flag is not enabled by default though.

Reason why I made it an opt-in flag is because this command is used during migration tests and devenv and didn't want to risk breaking anything there.

azaslavsky (approved):

msschl:
While upgrading to 24.6.0 i get the following error
```sh
Error: No such option: --create-kafka-topics
Error in install/set-up-and-migrate-database.sh:30.
'$dcr web upgrade --create-kafka-topics' exited with status 2
-> ./install.sh:main:38
--> install/set-up-and-migrate-database.sh:source:30

Cleaning up...
```

PR #3144, opened 2024-06-19
feat(relay): Forward /api/0/relays/* to inner relays

iambriccardo:


jjbayer (approved):

Dav1dde (approved):

nginx/nginx.conf
```diff
@@ -85,6 +85,9 @@ http {
 		location ~ ^/api/[1-9]\d*/ {
 			proxy_pass http://relay;
 		}
+		location ~ ^/api/0/relays/ {
```
oioki, line 88: nit: https://nginx.org/en/docs/http/ngx_http_core_module.html#location
> If the longest matching prefix location has the “^~” modifier then regular expressions are not checked.

This avoids the usage of regex and makes nginx a tiny bit faster:

```suggestion
		location ^~ /api/0/relays/ {
```

iambriccardo: Thanks!

stayallive:
@iambriccardo when deploying this change the relay container simply stops working because it can't connect to the relay container. Isn't the inner relay trying to retrieve the project config from itself now?

Dav1dde:
@stayallive can you try switching the upstream in the Relay config from:

https://github.com/getsentry/self-hosted/blob/da06c0f230454b91fada3980ba2e1e012cedf1d2/relay/config.example.yml#L2

to `upstream: "http://sentry:9000/"`?

Dav1dde:
Actually nvm, the `sentry` is just the Nginx upstream, so should be fine: https://github.com/getsentry/self-hosted/blob/da06c0f230454b91fada3980ba2e1e012cedf1d2/nginx/nginx.conf#L74-L77

stayallive:
I see, my relay is connecting to nginx though... this might have changed in the past... let me correct that!

stayallive:
Okay, user error here. Sorry! 

I think a few (hundred) releases ago there was some other issue that was solved by connecting relay through nginx, but that has long been not needed anymore and was forgotten to revert. Seems to all be working now connecting straight to sentry! Thanks for the hint!

PR #3177, opened 2024-07-02
feat(generic-metrics): Add gauges to docker compose, re-try

ayirr7:
Adding this missing consumer, this is a repeat of https://github.com/getsentry/self-hosted/pull/2757

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -295,7 +295,7 @@ services:
     command: rust-consumer --storage generic_metrics_counters_raw --consumer-group snuba-gen-metrics-counters-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
   snuba-generic-metrics-gauges-consumer:
     <<: *snuba_defaults
-    command: consumer --storage generic_metrics_gauges_raw --consumer-group snuba-gen-metrics-gauges-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
+    command: rust-consumer --storage generic_metrics_gauges_raw --consumer-group snuba-gen-metrics-gauges-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
   snuba-replacer:
     <<: *snuba_defaults
     command: replacer --storage errors --auto-offset-reset=latest --no-strict-offset-reset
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -293,6 +293,9 @@ services:
   snuba-generic-metrics-counters-consumer:
     <<: *snuba_defaults
     command: rust-consumer --storage generic_metrics_counters_raw --consumer-group snuba-gen-metrics-counters-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
+  snuba-generic-metrics-gauges-consumer:
+    <<: *snuba_defaults
+    command: consumer --storage generic_metrics_gauges_raw --consumer-group snuba-gen-metrics-gauges-consumers --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
```
aldy505, line 298: Are you sure that you don't want to use rust-consumer?

aldy505 (approved):

hubertdeng123:
Wouldn't this also require us to flip the feature flags for custom metrics? Otherwise, what is this consumer used for?

docker-compose.yml
ayirr7: Sorry, let me fix that.

ayirr7:
> Wouldn't this also require us to flip the feature flags for custom metrics? 

Could you point me to the feature flags you're referring to? But yes, you are correct that today the gauges consumer is only used by the custom metrics functionality.

aldy505:
> Wouldn't this also require us to flip the feature flags for custom metrics? Otherwise, what is this consumer used for?

I think it's fine since AFAIK Metrics isn't GA yet. We can add the flags in a separate PR.

aldy505:
Ping @ayirr7. We are 5 more days until the next release.

patrickklaeren:
Is there any news on when this might be merged?

I understand this has now missed the window for .7

ayirr7:
> I believe adding the gauge consumer now is a good idea, since it will ingest some data beforehand, that can be viewed later on the UI when GA's arrived -- so the Metrics UI won't be empty when it's enabled.

I believe this should be the case here. The consumer ingest any relevant data. Once any feature(s) are enabled, they can be used to query the data.

hubertdeng123 (approved):

aldy505:
Thanks Riya! @ayirr7

PR #3184, opened 2024-07-05
Update sentry-admin.sh to select its own working directory

theoriginalgri:
`sentry-admin.sh` only works when called from the working directory and not using its absolute path.

This change makes it also callable using its absolute path.

### Legal Boilerplate


BYK (approved):

hubertdeng123 (approved):

theoriginalgri:
I've applied the changes the pre-commit check wanted to see 👍

Issue #3113, opened 2024-06-05
Errors-only Self-hosted


hubertdeng123:
Over time, Sentry infrastructure has become larger as we’ve scaled as a company and released a plethora of features. This has created the behemoth of the present day self-hosted Sentry instance, with its 50+ docker containers. These containers eat at the RAM/CPU of servers, causing large frustration for our user base. Oftentimes, users are not even using most of the features that require all the containers up and running, so these end up as idle processes. We hear the frustration from our users and want to provide a way to deploy smaller versions of self-hosted Sentry that will delight our users that only care about errors. 

# Goals
1. Reduce friction it takes to run self-hosted for errors by reducing number of containers up and running
2. Continue to provide a way for current users to continue using all of Sentry features
3. Provide users an easier experience for running self-hosted Sentry

# Tasks

- [x] add config setting for errors only self hosted
- [x] gate frontend components behind self hosted errors only setting
- [ ] provide config for errors only self-hosted that will significantly reduce infrastructure needed
  - [ ] https://github.com/getsentry/self-hosted/pull/3190
- [ ] disable api endpoints/celery tasks related to non-errors self-hosted
- [ ] testing for self-hosted errors only

romanstingler:
can we at least add CRON, this should be a super lightweight feature

aldy505:
> can we at least add CRON, this should be a super lightweight feature

Cron requires some additional containers, so probably..... no? But I don't know myself, we'll see!

aldy505:
Related https://github.com/getsentry/self-hosted/issues/2773

aldy505:
Closing this since it's implemented already https://github.com/getsentry/self-hosted/pull/3190

PR #3190, opened 2024-07-09
Add errors only self-hosted infrastructure

hubertdeng123:
This does a couple of things:

Sets the default docker compose profile to "feature-complete". This way, users can opt in to using errors-only self-hosted if they want, otherwise nothing changes. 

To enable errors only self hosted

Change .env file
SENTRY_SELF_HOSTED_ERRORS_ONLY = True
COMPOSE_PROFILES = errors-only

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 COMPOSE_PROJECT_NAME=sentry-self-hosted
+SENTRY_SELF_HOSTED_ERRORS_ONLY=false
 COMPOSE_PROFILES=feature-complete
 SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 COMPOSE_PROJECT_NAME=sentry-self-hosted
+COMPOSE_PROFILES=errors-only
 SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 COMPOSE_PROJECT_NAME=sentry-self-hosted
-COMPOSE_PROFILES=errors-only
+COMPOSE_PROFILES=feature-complete
 SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
 # See https://docs.docker.com/compose/compose-file/#ports for more
--- a/.env
+++ b/.env
@@ -1,5 +1,4 @@
 COMPOSE_PROJECT_NAME=sentry-self-hosted
-SENTRY_SELF_HOSTED_ERRORS_ONLY=false
 COMPOSE_PROFILES=feature-complete
 SENTRY_EVENT_RETENTION_DAYS=90
 # You can either use a port number or an IP:PORT combo for SENTRY_BIND
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -108,11 +108,15 @@ x-sentry-defaults-feature-complete: &sentry_defaults-feature-complete
       <<: *depends_on-default
     vroom:
       <<: *depends_on-default
+    profiles:
+      - feature-complete
 x-snuba-defaults: &snuba_defaults
   <<: *restart_policy
   depends_on:
     clickhouse:
       <<: *depends_on-healthy
+    zookeeper:
+      <<: *depends_on-default
     kafka:
       <<: *depends_on-healthy
     redis:
@@ -390,7 +394,7 @@ services:
     volumes:
       - "sentry-symbolicator:/data"
   web:
-    <<: *sentry_defaults-feature-complete
+    <<: [*sentry_defaults, *sentry_defaults-feature-complete]
     ulimits:
       nofile:
         soft: 4096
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -335,6 +390,22 @@ services:
...
+      <<: *healthcheck_defaults
+      test:
+        - "CMD"
+        - "/bin/bash"
+        - "-c"
+        # Courtesy of https://unix.stackexchange.com/a/234089/108960
+        - 'exec 3<>/dev/tcp/127.0.0.1/9000 && echo -e "GET /_health/ HTTP/1.1\r\nhost: 127.0.0.1\r\n\r\n" >&3 && grep ok -s -m 1 <&3'
+    profiles:
+      - feature-complete
+  web-errors-only:
```
aldy505, line 408: Since we don't modify the `nginx.conf` (or maybe we don't... yet), should we specify `container_name` to have the same value as the usual `web` container? https://docs.docker.com/compose/compose-file/05-services/#container_name

docker-compose.yml
```diff
@@ -453,6 +561,45 @@ services:
...
+        read_only: true
+        source: ./nginx
+        target: /etc/nginx
+      - sentry-nginx-cache:/var/cache/nginx
+    depends_on:
+      - web-errors-only
+      - relay-errors-only
+    profiles:
+      - errors-only
+  relay-errors-only:
```
aldy505, line 582: Shouldn't relay behave the same for errors-only and feature-complete ones?

docker-compose.yml
aldy505: Tbh, I don't know if it'll work, but should we give it a try?

docker-compose.yml
hubertdeng123: Yes, it pretty much does. However, I wasn't able to get to a clean solution here with the docker compose architecture and wanted to avoid merging docker compose files. The difference is the `depends_on` clause for docker compose. For relay-errors-only, we it depends on `web-errors-only`, which has a different set of services it depends on to be considered healthy.

docker-compose.yml
hubertdeng123: I considered this, but the default naming of containers also uses the `COMPOSE_PROJECT_NAME`, and I didn't want to overwrite that for our users.

hubertdeng123: hm this will be a problem with both the install script and nginx conf, it would be good to use the same service name

hubertdeng123: I just cleaned this up now :)

docker-compose.yml
aldy505: Ah got it

aldy505:
Gonna review later. But we definitely need to update the docs.

sentry/sentry.conf.example.py
```diff
@@ -363,3 +363,8 @@ def get_internal_network():
 # https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS
 
 # CSRF_TRUSTED_ORIGINS = ["https://example.com", "http://127.0.0.1:9000"]
+
+# If you would like to use self-hosted Sentry with only errors enabled, please set this
```
aldy505, line 367: Should we also define what's "errors only" and what's not? I mean it's clear you'd know what "errors only" means if you previously use Sentry. But I don't think newcomers will instantly understand about what features are available on the "feature complete" profile.

aldy505 (approved):
Should also link the issue: https://github.com/getsentry/self-hosted/issues/3113

sentry/sentry.conf.example.py
hubertdeng123: Yep, I think we should. Going to also need to update docs, which I'll do today.

docker-compose.yml
```diff
@@ -36,24 +36,8 @@ x-sentry-defaults: &sentry_defaults
       <<: *depends_on-default
     snuba-api:
       <<: *depends_on-default
-    snuba-errors-consumer:
```
hubertdeng123, line 39: These are not needed to bring up the web container and other ingest consumers

.env
```diff
@@ -1,4 +1,6 @@
 COMPOSE_PROJECT_NAME=sentry-self-hosted
+SENTRY_SELF_HOSTED_ERRORS_ONLY=false
+COMPOSE_PROFILES=feature-complete
```
azaslavsky, line 3:
```suggestion
COMPOSE_PROFILES=sentry-feature-complete
```
Then get rid of `SENTRY_SELF_HOSTED_ERRORS_ONLY`, rename the profile to `sentry-feature-complete` in the compose.yml, and check `COMPOSE_PROFILES=sentry-feature-complete` in `...conf.py`. As it stands, we have two booleans representing one state ("is this errors only or not?"). We should avoid this redundant flagging, and just use the single boolean if possible.

hubertdeng123: done

azaslavsky (approved):

sentry/sentry.conf.example.py
```diff
@@ -363,3 +363,6 @@ def get_internal_network():
...
+SENTRY_SELF_HOSTED_ERRORS_ONLY = env("COMPOSE_PROFILES") == "feature-complete"
```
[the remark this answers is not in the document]
hubertdeng123: You're right htis is a bug, thanks for raising. I just checked my local `sentry.conf.py` and it uses `SENTRY_SELF_HOSTED_ERRORS_ONLY = env("COMPOSE_PROFILES") != "feature-complete"` when I was testing this locally. Opened https://github.com/getsentry/self-hosted/pull/3220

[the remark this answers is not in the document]
stayallive: I see, I wasn't 100% sure but I just also debugged a self-hosted instance very bravely using nightly and was caught off guard by this 😅 Thanks for the quick PR!

sentry/sentry.conf.example.py
stayallive: See: https://develop.sentry.dev/self-hosted/#errors-only-self-hosted-sentry

soupdumplings: Thank you. I really need this feature!

hubertdeng123: Yep, projects are always included.

soupdumplings: https://develop.sentry.dev/self-hosted/#errors-only-self-hosted-sentryDoes the "traces" mentioned in the document refer to "stack traces" and are they no longer supported under "errors only"?

aldy505: No, traces refers to the submenu inside the "Explore" menu. 
image

Stack traces is included on the errors-only since it's a crucial component to the error monitoring mechanism

soupdumplings: Thank you very much for your patient answers. I use Sentry as our team's crash analysis platform. It's really powerful and useful. The "errors only" version is just perfect for me.

dosinabox:
Thank you for this option. But how do requirements change exactly for errors-only mode? Previously, 8 GB RAM was enough for the full set of features in version 23.7.0 (that I currently use in production), it was increased to 16 GB in version 23.11.1, and it looks like the same 16 GB is still required for the limited errors-only version 24.7.1.
I just wanted to get rid of all the features that I never use and be able to limit resource consumption, but it turns out that I actually need to double my costs.

aldy505:
> Thank you for this option. But how do requirements change exactly for errors-only mode? Previously, 8 GB RAM was enough for the full set of features in version 23.7.0 (that I currently use in production), it was increased to 16 GB in version 23.11.1, and it looks like the same 16 GB is still required for the limited errors-only version 24.7.1.
> 
> I just wanted to get rid of all the features that I never use and be able to limit resource consumption, but it turns out that I actually need to double my costs.

To be honest, I don't think either of the employee or the community have measured it. We'll make an update when we do. It should be lower than the current spec though!

Issue #3167, opened 2024-06-25
Postgres alpine image introduces unique constraint index errors


hubertdeng123:
Between self-hosted releases 24.4.2 and 24.6.0, we've noticed that users have been experiencing issues with duplicate environments/releases. This is likely due to a mistake on our side to introduce the alpine version of postgres [here](https://github.com/getsentry/self-hosted/pull/2975). This has since been reverted back to the debian image, but there is likely quite a few of you experiencing issues. 

We are creating this issue to document user issues and potential solutions. From here on out, we will only be using the debian image of postgres. 

https://github.com/getsentry/self-hosted/issues/3161
https://github.com/getsentry/self-hosted/issues/3166

hubertdeng123:
For users that are attempting to upgrade from the alpine image to the latest 24.6.0 or above, all indexes involving columns of type text, varchar, char, and citext should be reindexed before the instance is put into production ([source](https://wiki.postgresql.org/wiki/Locale_data_changes#What_to_do)). 

For users that are attempting to fix duplicate rows of data, the process may be pretty manual 🥲. It will likely involve removing the duplicate rows of data, along with models with foreign key relations.

candux:
I had the same problems and fixed them in the database with the follwing commands!
**THIS is dangerous and will delete data! Use at your own risk! Make a backup before**
```
ALTER TABLE ONLY public.sentry_commitfilechange
    DROP CONSTRAINT sentry_commitfilechange_commit_id_4c6f7ec25af34227_uniq;
DELETE from public.sentry_commitfilechange ou where (SELECT count(*) from public.sentry_commitfilechange inr where inr.commit_id = ou.commit_id and inr.filename = ou.filename) > 1;
ALTER TABLE ONLY public.sentry_commitfilechange
    ADD CONSTRAINT sentry_commitfilechange_commit_id_4c6f7ec25af34227_uniq UNIQUE (commit_id, filename);
    

DROP INDEX sentry_release_organization_id_f0a7ec9ba96de76;
DELETE from sentry_release_project where release_id in (SELECT id from public.sentry_release ou where (SELECT count(*) from public.sentry_release inr where inr.organization_id = ou.organization_id and inr.version = ou.version) > 1);

DELETE from sentry_releaseprojectenvironment where release_id in (SELECT id from public.sentry_release ou where (SELECT count(*) from public.sentry_release inr where inr.organization_id = ou.organization_id and inr.version = ou.version) > 1);

[...]

PR #3197, opened 2024-07-11
Check postgres os before proceeding with install

hubertdeng123:
We want to avoid situations like what happened in https://github.com/getsentry/self-hosted/issues/3167

### DIFF
--- a/install/set-up-and-migrate-database.sh
+++ b/install/set-up-and-migrate-database.sh
@@ -9,6 +9,12 @@ until $dc exec postgres psql -U postgres -c "select 1" >/dev/null 2>&1 || [ $RET
   sleep 1
 done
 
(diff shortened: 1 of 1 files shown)


azaslavsky (approved):

install/set-up-and-migrate-database.sh
```diff
@@ -9,6 +9,12 @@ until $dc exec postgres psql -U postgres -c "select 1" >/dev/null 2>&1 || [ $RET
   sleep 1
 done
 
+os=$(docker compose exec postgres cat /etc/os-release | grep 'ID=debian')
```
aldy505, line 12: Why are we not using `$dc exec`?

aldy505 (approved):

install/set-up-and-migrate-database.sh
hubertdeng123: good catch

Issue #3168, opened 2024-06-26
Upgrading Self Hosted Sentry to New Version 24.6.0 is Causing Issue


Swaroop-1613:
### Self-Hosted Version

24.5.1

### CPU Architecture

x86_84

### Docker Version

20.10.25

### Docker Compose Version

v2.19.1

### Steps to Reproduce

We tried upgrading Sentry to the latest version 24.6.0 in our test environment. While upgrading, we encountered with the below error 
`Error in install/upgrade-clickhouse.sh:15.
'$dc up -d clickhouse' exited with status 1
-> ./install.sh:main:25
--> install/upgrade-clickhouse.sh:source:15
Cleaning up...`

We tried going back to the previous version - 24.5.1 but we were encountering the below error:

`Error response from daemon: Container ****************** is restarting, wait until the container is running
Error in install/upgrade-clickhouse.sh:21.
'version=$($dc exec clickhouse clickhouse-client -q 'SELECT version()')' exited with status 1
-> ./install.sh:main:25
--> install/upgrade-clickhouse.sh:source:21
Cleaning up...
Cleaning up project directory and file based variables
00:00
ERROR: Job failed: exit status 1`

### Expected Result

We need to upgrade the Self Hosted Sentry Application from 24.5.1 to 24.6.0.

### Actual Result

Please find the below screenshots for the same:

![image](https://github.com/getsentry/self-hosted/assets/173468575/7a69beea-3477-4765-b8ba-2c810ce60367)

<img width="554" alt="Screenshot 2024-06-26 113217" src="https://github.com/getsentry/self-hosted/assets/173468575/57890f05-4db9-4597-8b8f-6fde498c0afe">

Swaroop-1613:
Hey @hubertdeng123 could you please help us with the issue.
Thanks in advance

hubertdeng123:
If you are upgrading your sentry instance, why is sentry-kafka missing? Is it perhaps named something else? Or did you end up removing the docker volume before upgrading?

TimShilov:
The OP mentioned the upgrade, but I have just encountered the same error on 24.5.1 without even trying to upgrade.
The Sentry was working fine for weeks after the last upgrade and then suddenly stopped (it sometimes happens when we have a spike of events).

[...]

PR #3210, opened 2024-07-19
fix: clickhouse up requires kafka and zookeeper volumes

joshuarli:
Fixes https://github.com/getsentry/self-hosted/issues/3168.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,12 +20,12 @@ source install/check-latest-commit.sh
 source install/check-minimum-requirements.sh
 
 # Let's go! Start impacting things.
-# Upgrading clickhouse needs to come first before turning things off, since we need the old clickhouse image
+source install/create-docker-volumes.sh
+source install/update-docker-volume-permissions.sh
+# Upgrading clickhouse needs to come before turning things off, since we need the old clickhouse image
 # in order to determine whether or not the clickhouse version needs to be upgraded.
 source install/upgrade-clickhouse.sh
 source install/turn-things-off.sh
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -20,12 +20,12 @@ source install/check-latest-commit.sh
 source install/check-minimum-requirements.sh
 
 # Let's go! Start impacting things.
-# Upgrading clickhouse needs to come first before turning things off, since we need the old clickhouse image
+source install/create-docker-volumes.sh
+source install/update-docker-volume-permissions.sh
+# Upgrading clickhouse needs to come before turning things off, since we need the old clickhouse image
 # in order to determine whether or not the clickhouse version needs to be upgraded.
 source install/upgrade-clickhouse.sh
```
hubertdeng123, line 27: I placed this strategically here, because upgrading clickhouse requires knowledge of past versions of clickhouse before proceeding. The only way I could foresee doing that would be to check if a past clickhouse volume exists, then bring up clickhouse to check the version.

Are you encountering errors during installation? I would expect the `upgrade-clickhouse.sh` script to be skipped completely for you

hubertdeng123: maybe there's stuff from sentry dev that's causing this to be run?

joshuarli: > upgrading clickhouse requires knowledge of past versions of clickhouse before proceeding

ahh i see

> Are you encountering errors during installation? 

no, i was just trying to reproduce the linked issue and it only happens if the volume's deleted (somehow), it makes sense now that it's required, so not an issue on our part

Issue #3214, opened 2024-07-22
I cannot see the Traces and Metrics on the UI


trixobird:
### Self-Hosted Version

24.7.0

### CPU Architecture

x86_64

### Docker Version

25.0.3

### Docker Compose Version

v2.29.0

### Steps to Reproduce

1. Open the UI

### Expected Result

On the left menu I expected to see Traces and Metrics

### Actual Result

I was not able to see these two options

aldy505:
Yeah those two are behind a feature flags.

For Metrics, please refer to https://github.com/getsentry/self-hosted/issues/2698#issuecomment-1884061830

For Traces, please add `"organizations:performance-trace-explorer",` onto the list of feature flags on the `sentry/sentry.conf.py` file.

PR #3239, opened 2024-08-01
feat: enable traces

aldy505:
**NOTE: I don't know if the feature flag will be removed in the future. But considering we have `errors-only` version of Sentry, I don't think this will be removed, at least for the near future. So I'm adding this here.**

Some people are already asking about it.

On Github: https://github.com/getsentry/self-hosted/issues/3214
On Discord: https://discord.com/channels/621778831602221064/796028405833007104/1260595924774813757, https://discord.com/channels/621778831602221064/796028405833007104/1266233909197410335

This is the "Traces" page:

image

### Legal Boilerplate

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -301,6 +301,7 @@ SENTRY_FEATURES.update(
             "organizations:insights-addon-modules",
             "organizations:mobile-ttid-ttfd-contribution",
             "organizations:performance-calculate-score-relay",
+            "organizations:performance-trace-explorer",
             "organizations:standalone-span-ingestion",
             "organizations:starfish-browser-resource-module-image-view",
             "organizations:starfish-browser-resource-module-ui",


TheDevMinerTV (commented):
works on my instance

gggritso:
@k-fish @Zylphrex thoughts?

gggritso:
As an aside, IMO "Beta" features should not be enabled by default. Folks who are interested are free to enable the feature on their instance! Maybe what we need is product documentation to help self-hosted users enable these features?

Zylphrex:
> As an aside, IMO "Beta" features should not be enabled by default. Folks who are interested are free to enable the feature on their instance! Maybe what we need is product documentation to help self-hosted users enable these features?

Agreed. In general, we should not enable features labeled beta in self hosted instances by default. Users can enable them on a case by case basis if they wish. But we should only enable a feature by default once it has been moved out of beta. In this case, there are some significant redesigns coming up and we would like to address them before we enable the feature by default.

aldy505:
> As an aside, IMO "Beta" features should not be enabled by default. Folks who are interested are free to enable the feature on their instance! Maybe what we need is product documentation to help self-hosted users enable these features?

There's always this: https://develop.sentry.dev/self-hosted/#enabling-preview-features

The docs should be on this repo's issue. But since this feature is simple enough to set up, I went in and add this instead. 

Compared to other features which requires tweaks here and there, requiring a new consumer container and stuff.

PR #3260, opened 2024-08-12
Remove cdc and wal2json and use the default postgres entrypoint

beezz:
CDC and wal2json in no longer used and has been removed in prod long time ago. This PR attempts to remove all the traces of cdc and wal2json from self-hosted setup possibly using the default postgresql docker entrypoint . 

### Legal Boilerplate


BYK:
Yes, thanks for doing this!

Not sure if you were able to dig it out yourself but here's when all this stuff was introduced: #938.

Looks like we need a removal from the `.env` file too?

beezz:
> Yes, thanks for doing this!
> 
> Not sure if you were able to dig it out yourself but here's when all this stuff was introduced: #938.
> 
> Looks like we need a removal from the `.env` and `.gitignore` files too?
> 
> Oh, and the `init_hba` file

Thanks for quick response! I removed the last bits 👍

BYK:
> Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /home/runner/work/self-hosted/self-hosted/self-hosted/postgres
  Error in install/set-up-and-migrate-database.sh:4.
  
Hmm, let's see why that is

BYK:
Looks like this needs to go too: https://github.com/getsentry/self-hosted/blob/053f4010b61adf4677d8ac81488d41dd95461fba/docker-compose.yml#L140-L145

beezz:
> Looks like this needs to go too:
> 
> https://github.com/getsentry/self-hosted/blob/053f4010b61adf4677d8ac81488d41dd95461fba/docker-compose.yml#L140-L145

oh, thanks! Removed the read only bind as well now.

hubertdeng123 (approved):

PR #3263, opened 2024-08-13
Migrate to zookeeper-less kafka 

hubertdeng123:
follow up to https://github.com/getsentry/self-hosted/pull/2445, which was reverted since it broke existing installs


install/turn-things-off.sh
```diff
@@ -6,6 +6,10 @@ if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
 else
   # Clean up old stuff and ensure nothing is working while we install/update
   $dc down -t $STOP_TIMEOUT --rmi local --remove-orphans
+  # TODO(getsentry/self-hosted#2489)
+  if docker volume ls | grep -qw sentry-zookeeper; then
+    docker volume rm sentry-zookeeper
```
BYK, line 11: Well, I love this that said if the upgrade fails for some reason you have now lost data?

hubertdeng123: Moved this removal to the end of the installation then to ensure it will only run if the upgrade succeeds.

hubertdeng123:
I just tested this upgrade on our dogfood instance, and it seems to be running successfully

BYK (approved):
🚀

IanWoodard (approved):

PR #3275, opened 2024-08-16
Mandate minimum requirements for ram/cpu

hubertdeng123:
get rid of soft requirements for ram/cpu, enforce this as a hard requirement instead.
16 GB of ram, 4 CPU cores


joshuarli (approved):

IanWoodard (approved):

BYK:
Would be great to record the reasoning for this change somewhere

lukasleitsch:
I'm using a server (Hetzner) with 16GB of RAM, but a few MB are missing. The install script fails with:

```
FAIL: Required minimum RAM available to Docker is 15900 MB, found 15611 MB
```

Would be great to decrease the limit a little bit.

danilogco:
@joshuarli is it possible to reduce the limit? Just like @lukasleitsch, my instance hosted in AWS it's supposed to be 16GB of RAM, but it's a little below of the imposed limit. Well, I changed the file manually inside the EC2 to make it work.

joshuarli:
Yes, sorry. Thanks for the feedback! https://github.com/getsentry/self-hosted/pull/3290

PR #3292, opened 2024-08-25
Update check-minimum-requirements.sh

mhshakouri:
fix wrong version comparison in minimum requirements check

This humble edit, tries to fix wrong version comparison in `check-minimum-requirements.sh` file.

### Legal Boilerplate

### DIFF
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -4,8 +4,11 @@ source install/_min-requirements.sh
 
 # Check the version of $1 is greater than or equal to $2 using sort. Note: versions must be stripped of "v"
 function vergte() {
-  printf "%s\n%s" $1 $2 | sort --version-sort --check=quiet --reverse
-  echo $?
+  if [[ "$(printf "%s\n%s" "$1" "$2" | sort --version-sort | head -n 1)" == "$2" ]]; then
+    return 0  # $1 is greater than or equal to $2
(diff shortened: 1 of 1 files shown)


aminvakil (approved):
I like this approach better than relying on return code of sort command.

aldy505 (approved):

joshuarli:
Can you provide an example of why it's wrong?

hubertdeng123:
I'm in agreement with @joshuarli here, would like an example

aminvakil:
I couldn't find a wrong comparison as well, but still

> I like this approach better than relying on return code of sort command.

BYK:
Closing due to lack of response from @mhshakouri. Happy to revisit once/if we get a response to the questions raised.

mhshakouri:
I'm sorry to reach too late here, the pull request closing email alert got my attention in the end.
I remember that I had an issue trying run this on a local machine, the version detector for docker, ran in to an issue of detecting the version wrongly. I can not remember the exact version numbers, but the version was like 23 as min and I had I guess 24 or something like that, I remember this was kinda about mis handling the minor versions to  a string like comparison.

any how, I can try again to reproduce this issue, but dont have the time on my hand rn.
@BYK

BYK:
@mhshakouri no worries! You are always welcome if you have this issue again or anything else :)

PR #3365, opened 2024-10-06
ref: allow hosted js sdk bundles

aldy505:
Since this is hanging on my branch, I guess I'll make it out here. A part of making the JS SDK loader script available on the self-hosted sentry. Refer to https://github.com/getsentry/sentry/issues/22715.

Since not everyone is happy on hosting the JS SDK themself, therefore I made it into a feature flag. Set `SETUP_JS_SDK_ASSETS=1` on the `.env` file to enable the script. Also set `SETUP_JS_SDK_KEEP_OLD_ASSETS=1` to never delete (or basically cleanup) previous JS SDK version.

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -460,6 +460,7 @@ services:
         source: ./nginx
         target: /etc/nginx
       - sentry-nginx-cache:/var/cache/nginx
+      - sentry-nginx-www:/var/www
     depends_on:
       - web
       - relay
@@ -528,6 +529,11 @@ volumes:
   sentry-clickhouse:
     external: true
   sentry-symbolicator:
+    external: true
+  # This volume stores JS SDK assets, this is created as an external volume
+  # to minimize hiccups created by the nginx container and its' dependencies.
+  # The data inside this volume should be cleaned periodically on upgrades.
+  sentry-nginx-www:
     external: true
   # This volume stores profiles and should be persisted.
   # Not being external will still persist data across restarts.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -534,7 +534,6 @@ volumes:
   # to minimize hiccups created by the nginx container and its' dependencies.
   # The data inside this volume should be cleaned periodically on upgrades.
   sentry-nginx-www:
-    external: true
   # This volume stores profiles and should be persisted.
   # Not being external will still persist data across restarts.
   # It won't persist if someone does a docker compose down -v.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -530,9 +530,8 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
-  # This volume stores JS SDK assets, this is created as an external volume
-  # to minimize hiccups created by the nginx container and its' dependencies.
-  # The data inside this volume should be cleaned periodically on upgrades.
+  # This volume stores JS SDK assets and the data inside this volume should
+  # be cleaned periodically on upgrades.
   sentry-nginx-www:
   # This volume stores profiles and should be persisted.
   # Not being external will still persist data across restarts.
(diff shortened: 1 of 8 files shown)


nginx/nginx.conf
```diff
@@ -88,6 +88,10 @@ http {
 		location ^~ /api/0/relays/ {
 			proxy_pass http://relay;
 		}
+		location ^~ /js-sdk/ {
+			autoindex on;
+			root  /var/www/js-sdk;
+		}
```
aldy505, line 94:
```suggestion
		}
```

BYK:
Wow, this looks quite nice! I'd make it opt-out rather than opt-in as it the assets shouldn't take too much space and this is a great quality of life improvement.

I still don't get the reasoning behind using an external volume compared to a project-specific volume tho. Can you expand on that a bit?

aldy505:
> Wow, this looks quite nice! I'd make it opt-out rather than opt-in as it the assets shouldn't take too much space and this is a great quality of life improvement.
> 
> 
> 
> I still don't get the reasoning behind using an external volume compared to a project-specific volume tho. Can you expand on that a bit?

@BYK I made it quite a while ago so I don't really remember. Probably because the nginx container has web container as the dependency, so I think we shouldn't start the nginx container at all. 

If there is something like "docker compose volume create", that'd be nice.

BYK:
>@BYK I made it quite a while ago so I don't really remember. Probably because the nginx container has web container as the dependency, so I think we shouldn't start the nginx container at all.

I mean you don't need to spin up `nginx` (and its dependencies) to be able to use the volume? We can bind that to a different service or just use `--no-deps` to not start any other dependencies. You may add a pseudo service with a base Debian image that defines all volumes on it and use that as the "install helper". We can add that to a profile such as `installer` so it is never included in regular setups.

WDYT?

aldy505:
> > @BYK I made it quite a while ago so I don't really remember. Probably because the nginx container has web container as the dependency, so I think we shouldn't start the nginx container at all.
> 
> I mean you don't need to spin up `nginx` (and its dependencies) to be able to use the volume? We can bind that to a different service or just use `--no-deps` to not start any other dependencies. You may add a pseudo service with a base Debian image that defines all volumes on it and use that as the "install helper". We can add that to a profile such as `installer` so it is never included in regular setups.
> 
> WDYT?

In the end, I used `--no-deps`.

But I'd still prefer making this opt-in instead. I don't know about other people though.

aldy505:
Finally.

docker-compose.yml
```diff
@@ -529,6 +530,10 @@ volumes:
     external: true
   sentry-symbolicator:
     external: true
+  # This volume stores JS SDK assets, this is created as an external volume
+  # to minimize hiccups created by the nginx container and its' dependencies.
+  # The data inside this volume should be cleaned periodically on upgrades.
```
BYK, line 535: Just helping you out :)

```suggestion
  # This volume stores JS SDK assets and the data inside this volume should
  # be cleaned periodically on upgrades.
```

install/setup-js-sdk-assets.sh
```diff
@@ -0,0 +1,37 @@
+# This will only run if the SETUP_JS_SDK_ASSETS environment variable is set to 1
+if [[ "${SETUP_JS_SDK_ASSETS:-}" == "1" ]]; then
```
hubertdeng123, line 2: IMO if we're going to include this as part of the install script moving forwards, I'd like to see it in the `.env` file just to surface this as something people can set up. We can leave it as default for opt-in

hubertdeng123 (commented):
Looks pretty good, just one minor comment. This is awesome work!

install/setup-js-sdk-assets.sh
aldy505: Oh not you too :(

aldy505: Oh sorry I read that wrong. Lol.

aldy505: Okay done

hubertdeng123 (approved):

PR #3390, opened 2024-10-18
fix(sentry-admin): Do not wait for command finish to display output

Makhonya:
Currently sentry-admin.sh script saves the output of the command to a separate variable. This makes the command "freeze" if it requires any input from the user (like sentry-admin.sh restore). The change should provide the output directly to the shell.

Also changed contributing.md as it seemed outdated and updated requirements-dev.txt with the missing `cryptography` package.

### Legal Boilerplate

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,13 @@
 ## Testing
 
-Validate changes to the setup by running the integration test:
+### Running Tests with Pytest
 
-```shell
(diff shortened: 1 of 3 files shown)


BYK:
LGTM but deferring to @hubertdeng123 for final approval as he created the python based tests.

hubertdeng123 (approved):
Also looks reasonable to me, thanks for this!

PR #3417, opened 2024-11-07
fix: Use js.sentry-cdn.com for JS SDK downloads

BYK:


vgrozdanic (approved):

aldy505 (approved):
Lets just do this for now.

Issue #3366, opened 2024-10-07
Failure due to - Redis is loading the dataset in memory


klemen-df:
### Self-Hosted Version

24.9.0

### CPU Architecture

x86_64

### Docker Version

27.3.1

### Docker Compose Version

2.29.7

### Steps to Reproduce

1. docker compose up -d

### Expected Result

Sentry should start successfully

### Actual Result

sentry-self-hosted-web-1 fails with 'sentry.exceptions.InvalidConfiguration: Redis is loading the dataset in memory'

I think this is due to:

```
  redis:
    <<: *restart_policy
    image: "redis:6.2.14-alpine"
    healthcheck:
      <<: *healthcheck_defaults
      **test: redis-cli ping**
```

So, in the test, we only test if the redis container is up and there's no check if the data was loaded.

If I do this it's more or less always successfull. 

`      **test: sleep 30; redis-cli ping**
`
Since this is not a proper solution, can the test be improved?

Sometimes it's also ok to do it two steps:

```
docker compose up redis -d
docker compose up -d
```

hubertdeng123:
Thanks for writing in. Is this error crashing your web instance?

klemen-df:
It sometimes causes issues when starting, yes. It's a race condition.

Either way - it's not ok to say that if 'ping' goes throught, the redis service is ready. 'ping' just means that the container is up but it doesn't necessarily mean that the database is loaded.

bijancot:
@klemen-df i did facing this issue. But in my case the sentry is not health from the start. So my workaround is to start the redis and  flush the redis with assumption some cache or data is outdated and not well processed so it's safe for me to start flushing it up. Then everythings work fine

klemen-df:
what about this?

```
redis-cli INFO Persistence | grep loading
```

1: in loading
0: loaded

bijancot:
> what about this?
> 
> ```
> redis-cli INFO Persistence | grep loading
> ```
> 
> 1: in loading 0: loaded

At that time, that's not crossing my mind (sorry). And i'm not checking that one

hubertdeng123:
Looks like what we can do is redis-cli ping | grep PONG

https://stackoverflow.com/questions/67904609/how-do-you-perform-a-healthcheck-in-the-redis-docker-image

PR #3422, opened 2024-11-11
feat(healthcheck): Improve redis healthcheck

hubertdeng123:
Will close https://github.com/getsentry/self-hosted/issues/3366

Improves the current redis healthcheck to also factor in loading dataset in memory


bc-sentry (approved):
LGTM

PR #3427, opened 2024-11-13
feat: add Redis configuration for improved memory management

Hassanzadeh-sd:
As Sentry continues to evolve, effective resource management becomes crucial for maintaining performance and stability. This update includes configurations that will help optimize Redis's memory usage, ensuring that the system runs efficiently under varying loads.

**Key Changes:**

- **Added `maxmemory` Directive**: Configured Redis to limit its memory usage to a specified size. This prevents excessive memory consumption and helps maintain system stability.
- **Set `maxmemory-policy` to `allkeys-lru`**: This policy allows Redis to evict the least recently used keys when it reaches the memory limit, ensuring that frequently accessed data remains available while older, less-used data is removed.

Thank you for considering this enhancement to improve Redis's performance within the Sentry ecosystem!


docker-compose.yml
```diff
@@ -110,6 +110,10 @@ services:
       test: redis-cli ping | grep PONG
     volumes:
       - "sentry-redis:/data"
+      - type: bind
+        read_only: true
+        source: ./redis/redis.conf
+        target: /usr/local/etc/redis/redis.conf
```
BYK, line 116: Would you mind either mounting the entire `redis` directory or moving the `redis.conf` file to the root level?

See this comment/discussion for context: https://github.com/getsentry/self-hosted/pull/3395#discussion_r1830788235

redis/redis.conf
```diff
@@ -0,0 +1,26 @@
...
+# constraints are not a concern.
+#
+# Alternatively, you can specify a limit, such as 'maxmemory 15gb', to restrict Redis to 
+# using a maximum of 15 gigabytes of memory. 
+#
+# Example:
+# maxmemory 0         # Unlimited memory usage
+# maxmemory 15gb     # Limit memory usage to 15 GB
+
+maxmemory 0
```
BYK, line 15: What is the default and shall really set this to unlimited?

BYK (changes requested):
Thank you so much for the PR!

Requested a change and left a question. After that I think this is good to merge but I'll also consult our SRE team to make sure.

Finally, if you could add this boilerplate to your PR description, we'd be able to merge at ease: https://github.com/getsentry/self-hosted/blob/master/.github/PULL_REQUEST_TEMPLATE.md

redis/redis.conf
Hassanzadeh-sd: Here’s an improved version of your text for clarity and professionalism: The default value of the maxmemory setting in Redis depends on the system architecture:

    64-bit Systems: By default, there is no limit on memory usage. This allows Redis to utilize as much RAM as the operating system permits until it runs out of available memory.
    32-bit Systems: The implicit memory limit is typically set to 3GB, which is a consequence of the limitations inherent in 32-bit addressing.

I believe we can set the default value to unlimited, allowing users to adjust this setting as needed based on their specific requirements.

docker-compose.yml
Hassanzadeh-sd: hi BYK, as you said I changed it to these values and paths:
```
    volumes:
      - "sentry-redis:/data"
      - type: bind
        read_only: true
        source: ./redis.conf
        target: /usr/local/etc/redis/redis.conf
```

aminvakil (approved):

BYK (approved):
Got word from a colleague: "For a self hosted Relay itself, that config would be fine."

BYK:
Note to followers: if the eviction policy kicks in, there might be data loss as Redis is not just used as a cache but also a processing queue/storage.

mwarkentin:
In prod our redis are split into many different clusters and instances, so I don't know what the best setting would be for a deployment where everything is running within a single redis. It might be worth considering `volatile-lru` instead of `allkeys-lru`:

```
volatile-lru: Evict the least recently used keys that have the expire field set to true.
```

> The volatile-lru and volatile-random policies are mainly useful when you want to use a single Redis instance for both caching and for a set of persistent keys. However, you should consider running two separate Redis instances in a case like this, if possible.

inoa-dmpassy:
AFAIK, redis doesnt read from redis.conf by default
https://hub.docker.com/_/redis
```
Additionally, if you want to use your own redis.conf ...

You can create your own Dockerfile that adds a redis.conf from the context into /data/, like so.

FROM redis
COPY redis.conf /usr/local/etc/redis/redis.conf
CMD [ "redis-server", "/usr/local/etc/redis/redis.conf" ]

Alternatively, you can specify something along the same lines with docker run options.

$ docker run -v /myredis/conf:/usr/local/etc/redis --name myredis redis redis-server /usr/local/etc/redis/redis.conf

Where /myredis/conf/ is a local directory containing your redis.conf file. Using this method means that there is no need for you to have a Dockerfile for your redis container.
```

There's no reference in the docker-compose to use this redis.conf

BYK:
@inoa-dmpassy - wow, thanks for the heads up! Fix coming up

PR #3432, opened 2024-11-19
fix(redis): Use a safer eviction rule

BYK:
Follow up to https://github.com/getsentry/self-hosted/pull/3427#issuecomment-2485612688


mwarkentin (approved):
LGTM!

PR #3447, opened 2024-11-27
chore(issue-template): ask for machine specification and provide link to security policy

aldy505:
To remove roundtrip of asking people "what's your machine specification look like?"

### Legal Boilerplate

### DIFF
--- a/.github/ISSUE_TEMPLATE/problem-report.yml
+++ b/.github/ISSUE_TEMPLATE/problem-report.yml
@@ -45,8 +45,10 @@ body:
       label: Machine Specification
       description: Make sure your system meets the [minimum system requirements of Sentry](https://develop.sentry.dev/self-hosted/#required-minimum-system-resources).
       options:
-        - label: My system meets the [minimum system requirements of Sentry](https://develop.sentry.dev/self-hosted/#required-minimum-system-resources)
+        - label: My system meets the minimum system requirements of Sentry
           required: true
+    validations:
+      required: true
   - type: textarea
     id: repro
     attributes:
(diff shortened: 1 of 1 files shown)


bc-sentry:
Similar sentiment to @aminvakil.  We can ask for this information, but it would seem to be marginally useful if other users ran into a machine-specific installation or runtime issue.   On the flip side, by adding this field I wouldn't want it to imply to the poster that we will have some ability to try and reproduce their issue on that hardware.  In most cases, we do not have the bandwidth to reproduce issues and make a best effort try to respond with advice and/or transfer the issue to a product-specific team for triage.

BYK:
I propose adding a checkbox saying something like "My system meets [the minimum system requirements](https://develop.sentry.dev/self-hosted/#required-minimum-system-resources) of Sentry"

aldy505:
> I propose adding a checkbox saying something like "My system meets [the minimum system requirements](https://develop.sentry.dev/self-hosted/#required-minimum-system-resources) of Sentry" instead of asking for specs.

That sounds better.

aldy505:
This will be delayed, since I need to remember which of my device did I use to create this PR.....

aldy505:
> @aldy505 you can always do `gh pr checkout 3447` if you use GitHub CLI 😉

I know but I'm too lazy lol. Let me just find my device.

aldy505:
Alright, done. See the preview here https://github.com/getsentry/self-hosted/blob/cbda363095049c53a9d77ed17f5447d1d5f8a7a8/.github/ISSUE_TEMPLATE/problem-report.yml

BYK (approved):
Awesome 😎

PR #3458, opened 2024-12-04
feat(release): Replace release bot with GH app

Jeffreyhung:
More details: https://www.notion.so/sentry/DACI-Replace-GitHub-bot-accounts-with-GitHub-Apps-getsentry-release-15109965d1204a91b9be71c49e8b66e0?pvs=4

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,15 +46,8 @@ jobs:
     name: Create release on self-hosted dogfood instance
     needs: release
     steps:
-      - name: Get auth token
-        id: token
-        uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
-        with:
-          app-id: ${{ vars.SENTRY_RELEASE_BOT_CLIENT_ID }}
-          private-key: ${{ secrets.SENTRY_RELEASE_BOT_PRIVATE_KEY }}
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -40,9 +46,15 @@ jobs:
...
+      - name: Get auth token
+        id: token
+        uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
+        with:
+          app-id: ${{ vars.SENTRY_RELEASE_BOT_CLIENT_ID }}
+          private-key: ${{ secrets.SENTRY_RELEASE_BOT_PRIVATE_KEY }}
       - uses: actions/checkout@v4
         with:
-          token: ${{ secrets.GH_RELEASE_PAT }}
+          token: ${{ steps.token.outputs.token }}
```
BYK, line 57: I don't think we need a special token here. @hubertdeng123 @chadwhitacre can you weigh in? Why is `GITHUB_TOKEN` not enough for the release?

hubertdeng123: Hmm I would think we shouldn't need this either, this should just be for a release on our self-hosted dogfood instance so it would surprise me if we need this here. I think we can omit the token completely for the checkout step here

Jeffreyhung: Thanks for the insight! Removed it and let's see how it goes, we can always add it back if figured out later that it's actually needed.

BYK (approved):

PR #3459, opened 2024-12-05
fix(redis): Actually use custom config

BYK:
Follow up to https://github.com/getsentry/self-hosted/pull/3427#issuecomment-2518128717 where we created and mounted a custom Redis config only to not use it :facepalm:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -117,6 +117,9 @@ services:
       nofile:
         soft: 10032
         hard: 10032
+    command:
+      - redis-server
+      - /usr/local/etc/redis/redis.conf
   postgres:
     <<: *restart_policy
     # Using the same postgres version as Sentry dev for consistency purposes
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -117,6 +117,9 @@ services:
       nofile:
         soft: 10032
         hard: 10032
+    command:
```
hubertdeng123, line 120: nit: I think these may be more readable

```
command: ["redis-server", "/usr/local/etc/redis/redis.conf"]
command: redis-server /usr/local/etc/redis/redis.conf
```

BYK: I'll go with the array syntax

docker-compose.yml
```diff
@@ -117,6 +117,9 @@ services:
...
+      - redis-server
+      - /usr/local/etc/redis/redis.conf
```
BYK, line 122:
```suggestion
    command: ["redis-server", "/usr/local/etc/redis/redis.conf"]
```

hubertdeng123 (approved):

aldy505 (approved):

Issue #3330, opened 2024-09-17
Relay Server: Not Enough Memory on Health Check even though stats show otherwise


LordSimal:
### Self-Hosted Version

24.8.0

### CPU Architecture

x86_64

### Docker Version

27.2.1

### Docker Compose Version

2.29.2

### Steps to Reproduce

Can't really tell how to reproduce, since it just happens out of nowhere.

### Expected Result

Sentry receives errors again

### Actual Result

Sentry stops receiving errors after 1-2 days of normal usage.

Checking the docker logs there are a lot of these entries present:
```
relay-1                                         | 2024-09-17T06:33:00.437945Z ERROR relay_server::services::health_check: Not enough memory, 32351698944 / 33568419840 (96.38% >= 95.00%)
relay-1                                         | 2024-09-17T06:33:00.437982Z ERROR relay_server::services::health_check: Health check probe 'system memory' failed
```

but checking either htop we have enough RAM
![Image](https://github.com/user-attachments/assets/75729c06-db60-48e3-8031-132f6c785677)

as well as checking docker container stats there are no containers using > 95% RAM
```
CONTAINER ID   NAME                                                                CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS
48244b39b406   sentry-self-hosted-nginx-1                                          0.10%     11.74MiB / 31.26GiB   0.04%     829MB / 832MB     11.8MB / 131kB    13
4188cd46c1fe   sentry-self-hosted-relay-1                                          0.54%     515.6MiB / 31.26GiB   1.61%     896MB / 2.06GB    363MB / 284MB     40
ca96eedda64d   sentry-self-hosted-generic-metrics-consumer-1                       0.56%     342.4MiB / 31.26GiB   1.07%     177MB / 290MB     17.8MB / 82MB     21
5dedcba8621c   sentry-self-hosted-monitors-clock-tick-1                            0.28%     162.4MiB / 31.26GiB   0.51%     36MB / 33.5MB     35.8MB / 29.5MB   6
8e98edce4698   sentry-self-hosted-subscription-consumer-generic-metrics-1          0.28%     323.9MiB / 31.26GiB   1.01%     37.7MB / 34.5MB   13.2MB / 68.7MB   13
ca542ffd958a   sentry-self-hosted-attachments-consumer-1                           0.48%     500.2MiB / 31.26GiB   1.56%     16.9MB / 15.2MB   26.8MB / 66.6MB   19
[...]

PR #3486, opened 2024-12-20
chore(relay): provide opt-in max_memory_percent config as workaround for failing healthcheck

aldy505:
See https://github.com/getsentry/self-hosted/issues/3330

### Legal Boilerplate


BYK (approved):
I think this should be the default as I don't think the inferred memory value in a docker compose environment would ever be accurate

PR #3488, opened 2024-12-21
ci: Cache postgres volume after first migration

BYK:
This patch caches all DB volumes based on the sentry and snuba images to avoid doing the same migrations over and over for every test run.

This shaved off a whole minute from "Install self-hosted" jobs (so ~20% speed increase).

Left side: cached re-run -- Right side: no-cache initial run
image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,7 +129,7 @@ jobs:
         id: cache_key
         run: |
           # Set permissions for docker volumes so we can cache and restore
-          sudo chmod -R 777 /var/lib/docker/volumes/
+          sudo chmod 711 /var/lib/docker/
           source .env
           SENTRY_IMAGE_SHA=$(docker buildx imagetools inspect $SENTRY_IMAGE --format "{{json .Manifest}}" | jq -r '.digest')
           echo "::set-output name=SENTRY_IMAGE_SHA::$SENTRY_IMAGE_SHA"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -217,7 +217,6 @@ jobs:
 
       - name: Integration Test
         run: |
-          docker compose up --wait
           if [ "${{ matrix.compose_version }}" = "v2.19.0" ]; then
             pytest --reruns 3 --cov --junitxml=junit.xml _integration-test/ --customizations=${{ matrix.customizations }}
           else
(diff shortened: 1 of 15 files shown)


BYK:
>Can we have an env or something like that to punch through cache, I don't think of a reason which someone would need this right now (not using cache), and we can always add it later, what do you think?

Don't think this would be useful as this is only happening on CI and changing the cache env variable would require a new commit where we can simply bump the cache key or disable the cache step if we want to.

>On another matter, shouldn't we take postgres, clickhouse version into cache id, so if that gets changed, cache gets invalidated?

Don't think so as we should be fine with upgrades too. We still run migrations, upgrade etc. We just happen to start from a warm state which is quite similar to our upgrade test.

.github/workflows/test.yml
```diff
@@ -66,9 +71,48 @@ jobs:
...
+          echo "SNUBA_IMAGE_SHA=$SNUBA_IMAGE_SHA" >> $GITHUB_OUTPUT
+
+      - name: Restore DB Volumes Cache
+        id: restore_cache
+        uses: actions/cache/restore@v4
+        with:
+          key: db-volumes-v4-${{ steps.cache_key.outputs.SENTRY_IMAGE_SHA }}-${{ steps.cache_key.outputs.SNUBA_IMAGE_SHA }}
+          restore-keys: |
+            db-volumes-v4-${{ steps.cache_key.outputs.SENTRY_IMAGE_SHA }}
+            db-volumes-v4-
```
hubertdeng123, line 93: It does not seem like this is intentional? Perhaps 

`db-volumes-v4-${{ steps.cache_key.outputs.SNUBA_IMAGE_SHA }}` ?

hubertdeng123:
This is great, wondering how much time this shaves off CI time

.github/workflows/test.yml
BYK: Completely intentional. These are prefix matched in order for stale cache restoration. So if we cannot match on both sentry and snuba image SHA, then we fallback to one with sentry image sha regardless of snuba, and finally, it just restores _any cache_ as long as it is v4.

BYK:
>This is great, wondering how much time this shaves off CI time

It is in the description 😅:

>This shaved off a whole minute from "Install self-hosted" jobs (so ~20% speed increase).

hubertdeng123 (approved):

PR #3490, opened 2024-12-22
ref(geoip): Remove geoipupdate from compose

BYK:
`geoipupdate` is not used by any other service nor it is needed for any
service to run. Moreover, it is a one-shot command, causing
`docker compose up --wait` to fail when it exits with a non-zero status.
This happens when one has not yet set up their credentials and they may
choose to never do this.

This PR removes `geoipupdate` from the `docker-compose.yml` file and
moves the command directly into the geoip related script. One may run
this whenever they want to update their GeoIP database.

This PR needs an accompanying docs change.

It also has the side-effect of making integration tests slightly faster (due to the initial `docker compose up --wait`), and more reliable as we used to rely on the implicit `docker compose run web` from the `sentry_admin` test. This also skewed the test run duration of `sentry_admin` test by including the system startup time there.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -134,6 +134,7 @@ jobs:
 
       - name: Integration Test
         run: |
+          docker compose up --wait
           if [ "${{ matrix.compose_version }}" = "v2.19.0" ]; then
             pytest --reruns 3 --cov --junitxml=junit.xml _integration-test/ --customizations=${{ matrix.customizations }}
           else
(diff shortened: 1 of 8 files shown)


aldy505 (approved):

aldy505 (approved):
I don't remember which version since Docker introduce the `--wait` flag. But since most people are likely using the updated version of their Docker & Docker Compose, this should be good to go.

BYK:
@aldy505 I mean, we test (and support) 2.19 and 2.26 which both seem to support the flag and that's kind of all I care about 😅

aminvakil (approved):

hubertdeng123 (approved):

PR #3491, opened 2024-12-23
ref(snuba): Combine bootstrap & migrate for faster bootstrap

BYK:
I think we split these actions in the past due to some lack of options for them to work together properly. Right now looks like `bootstrap` would automatically migrate and propagates the `force` flag.


hubertdeng123 (approved):
CI seems to be looking fine with this

PR #3496, opened 2024-12-31
ci: Skip DB ops during install completely on cache hit

BYK:
Follow up to #3488

A new record: 2m 8s for installing self-hosted:
image

Issue #1808, opened 2022-11-14
Cleaning nodestore_node table


aminvakil:
### Problem Statement

Over time `nodestore_node` table gets bigger and bigger and currently there is no procedure to clean it up.

A forum comment explaining what `nodestore_node` is by @untitaker :

> To answer the question about purpose of nodestore:

> nodestore stores the raw, unindexed parts of the data of an event. For example all event breadcrumbs and contexts are stored there. Clickhouse only stores the data you can search and aggregate by. It only has indices, so to speak.

> nodestore has multiple backends. On [sentry.io](http://sentry.io/) it is not postgres, but google bigtable. We use the postgres backend for local development and for onpremise, but for large-scale installations it is suboptimal.

> We don’t have a solution right now for the storage problem in any case, sorry. One thing you should ensure actually runs is `sentry cleanup`, but it’s definitely possible that nodestore takes this amount of disk space.

https://forum.sentry.io/t/postgres-nodestore-node-table-124gb/12753/3

### Solution Brainstorm

There was [an idea suggested on forum](https://forum.sentry.io/t/postgres-is-growing-rapidly-manual-cleanup/13666/14) which worked for me, but I lost all event details.
Something like this would work: 
```
DELETE FROM public.nodestore_node WHERE “timestamp” < NOW() - INTERVAL '1 day';
VACUUM FULL public.nodestore_node;
```
Change 1 day according to your needs.

Maybe put this in a `cron` container which gets run every night, we should think about its performance issues though, this took a long time to get executed on our instance, maybe because it wasn't run before, but I'm not sure.

aminvakil:
cc @BYK @chadwhitacre @hubertdeng123

BYK:
There are a few interesting and important things:

1. We can never add the `VACUUM FULL public.nodestore_node;` part in any kind of production setup as it will lock the whole table for an indefinite time. It also requires additional space on disk so it has a high potential to put the system in an inoperable and hard-to-recover state.
2. The `sentry cleanup` command that we trigger via the cron _does_ run `nodestore.cleanup()`:
   https://github.com/getsentry/sentry/blob/1d24b160a74e10759d20fa323f5251bffee3f7b5/src/sentry/runner/commands/cleanup.py#L260-L268 which is defined at https://github.com/getsentry/sentry/blob/1d24b160a74e10759d20fa323f5251bffee3f7b5/src/sentry/nodestore/django/backend.py#L54-L62

[...]

PR #3498, opened 2024-12-31
feat: Use S3 node store with seaweedfs

BYK:
Enables S3 node store using [SeaweedFS](https://github.com/seaweedfs/seaweedfs) and [sentry-nodestore-s3](https://github.com/stayallive/sentry-nodestore-s3) by @stayallive 

This should alleviate all the issues stemming from (ab)using PostgreSQL as the node store.


aldy505:
Any reason why you didn't use SeaweedFS per what you said yesterday?

garage.toml
```diff
@@ -0,0 +1,29 @@
+metadata_dir = "/var/lib/garage/meta"
+data_dir = "/var/lib/garage/data"
+db_engine = "sqlite"
+
+replication_factor = 1
+# idk why rcp_ is needed without replication but ok
+rpc_bind_addr = "[::]:3901"
+rpc_public_addr = "127.0.0.1:3901"
```
aldy505, line 8: Since this is on Docker, I think it's better if we use `0.0.0.0:3901` as the hostname, or just `:3901`

garage.toml
```diff
@@ -0,0 +1,29 @@
+metadata_dir = "/var/lib/garage/meta"
+data_dir = "/var/lib/garage/data"
+db_engine = "sqlite"
```
aldy505, line 3: We should use `lmdb` instead. See https://garagehq.deuxfleurs.fr/documentation/reference-manual/configuration/#db_engine

install/bootstrap-garage.sh
```diff
@@ -0,0 +1,25 @@
+echo "${_group}Bootstrapping garage (node store)..."
+
+$dc up --wait garage
+garage="$dc exec garage /garage"
+
+if [[ $($garage bucket list | tail -1 | awk '{print $1}') != 'nodestore' ]]; then
+  node_id=$($garage status | tail -1 | awk '{print $1}')
+  $garage layout assign -z dc1 -c 100G "$node_id"
```
aldy505, line 8: Should this `100G` be a variable somewhere?

docker-compose.yml
```diff
@@ -207,6 +210,14 @@ services:
...
       timeout: 10s
       retries: 30
+  garage:
+    image: dxflrs/garage:v1.0.1
+    volumes:
+      - type: bind
+        read_only: true
+        source: ./garage.toml
+        target: /etc/garage.toml
+      - "sentry-garage:/var/lib/garage/"
```
aldy505, line 220:
```suggestion
  garage:
    image: dxflrs/garage:v1.0.1
    volumes:
      - type: bind
        read_only: true
        source: ./garage.toml
        target: /etc/garage.toml
      - "sentry-garage:/var/lib/garage/"
    healthcheck:
      test: "wget -no-verbose --tries=1 --spider http://localhost:3903/health || exit 1"
      interval: 20s
      timeout: 20s
      retries: 10      
```

sentry/sentry.conf.example.py
```diff
@@ -84,6 +84,25 @@ def get_internal_network():
...
+    "read_through": True,
+    "compression": False,  # we have compression enabled in Garage itself
+    "endpoint_url": "http://garage:3900",
+    "bucket_path": "nodestore",
+    "bucket_name": "nodestore",
+    "retry_attempts": 3,
+    "region_name": "garage",
+    "aws_access_key_id": "<GARAGE_KEY_ID>",
+    "aws_secret_access_key": "<GARAGE_SECRET_KEY>",
+}
```
aldy505, line 104: (Docs) should we provide ways for the user to offload these to actual S3 or something?

install/bootstrap-garage.sh
```diff
@@ -0,0 +1,25 @@
...
+
+  $garage bucket create nodestore
+  key_info=$($garage key create nodestore-key | head -3 | tail -2)
+  key_id=$(echo "$key_info" | head -1 | awk '{print $3}')
+  key_secret=$(echo "$key_info" | tail -1 | awk '{print $3}')
+
+  $garage bucket allow --read --write --owner nodestore --key nodestore-key
+
+  sed -i -e "s/<GARAGE_KEY_ID>/$key_id/" $SENTRY_CONFIG_PY
+  sed -i -e "s/<GARAGE_SECRET_KEY>/$key_secret/" $SENTRY_CONFIG_PY
```
aldy505, line 19: Should we throw an error or something if these weren't found on the conf file? A preflight maybe?

garage.toml
BYK: This entire `rpc_` section is obsolete as we are in single-node mode. Otherwise I think you are right.

garage.toml
BYK: Just read that page, I think we should stick with `sqlite` here due to the following issues in `lmdb`:

>Several users have reported corrupted LMDB database files after an unclean shutdown (e.g. a power outage). This situation can generally be recovered from if your cluster is geo-replicated (by rebuilding your metadata db from other nodes), or if you have saved regular snapshots at the filesystem level.

We will not have any snapshots or replication so in case of an unclean shutdown and DB corruption, it would be a nightmare (or impossible) to repair the storage.

>Keys in LMDB are limited to 511 bytes. This limit translates to limits on object keys in S3 and sort keys in K2V that are limted to 479 bytes.

Don't know how long our object keys are but I don't want to take any chances.

Sounds like `sqlite` should be good enough for our single-node, (hopefully) smaller scale deployment.

WDYT?

install/bootstrap-garage.sh
BYK: Yes, I think we should add a new `GARAGE_STORAGE_SIZE` env var to `.env`. That said not sure if that makes much sense as we would not honor any changes to that after the initial installation. Unless this actually _reserves_ 100G, I think leaving it hard-coded to a "good enough" value and then documenting how to change this if needed would be a better option.

Thoughts?

docker-compose.yml
```diff
@@ -139,6 +141,7 @@ services:
   kafka:
     <<: *restart_policy
     image: "confluentinc/cp-kafka:7.6.1"
+    user: root
```
hubertdeng123, line 144: any reason we're using root here?

docker-compose.yml
```diff
@@ -207,6 +210,19 @@ services:
...
+    image: dxflrs/garage:v1.0.1
+    volumes:
+      - type: bind
+        read_only: true
+        source: ./garage.toml
+        target: /etc/garage.toml
+      - "sentry-garage:/var/lib/garage/"
+    healthcheck:
+      test: "wget -no-verbose --tries=1 --spider http://localhost:3903/health || exit 1"
+      interval: 20s
```
hubertdeng123, line 223: This seems rather long, is this waiting for any other service to come up first? Would it be useful to include a depends_on block here?

install/bootstrap-garage.sh
BYK: Not sure if we really need this as if we cannot find that information, it implies we are not using the example file?

sentry/sentry.conf.example.py
BYK: Maybe something under the "experimental" part?

docker-compose.yml
BYK: This is from #3496, unrelated to this patch and should be a no-op once the other one is merged.

docker-compose.yml
BYK: Nope, we can definitely lower it. Right now it doesn't even work so I'll revert it for now.

BYK:
@aldy505
>Any reason why you didn't use SeaweedFS per what you said yesterday?

Well I started with that and realized 3 things:

1. It really is not geared towards single-node setups and have nodes with different roles. This makes is more challenging to scale up or set up in our setup
2. It has this paid admin interface. Not a deal breaker but it is clear that it is geared towards more "professional" setups
3. Its S3 API interface support is not really great

Garage fits the bill much better as it is explicitly created for smaller setups like this, easy to expand without specialized roles, doesn't have any paid thing in it, and has much more decent and familiar S3 interface support.

garage.toml
aldy505: Okay, let's try sqlite.

sentry/sentry.conf.example.py
aldy505: Probably.

install/bootstrap-garage.sh
aldy505: It's for people that already has the `sentry.conf.py` file and didn't check the new updates from `sentry.conf.example.py`. I'd prefer to have this as a warning though.

install/bootstrap-garage.sh
aldy505: I don't know if 100G will be immediately allocated by Garage. If it's not (meaning the current storage space won't be modified) then I think it's good. If it's not, I think it's better to just allocate it to 25G space.

sentry/sentry.conf.example.py
aldy505: Probably would be better if we put it on the Experimental -> External Storage page. Will backlog that.

doc-sheet:
> It really is not geared towards single-node setups and have nodes with different roles. This makes is more challenging to scale up or set up in our setup

when I tried seaweedfs last time (and I still use it for sourcemap/profile storage tbh) it had single node ability via `weed server` command.
Like
```
weed server -filter=true -s3=true -master=true -volume=true
```

Some of them enabled by default.

aldy505:
> I think garage/minio simpler for small setups, seaweedfs looks necessary for mid to high setups because all other services I know keep files as is.

@doc-sheet Hey, I'm going to work on this PR. I'd think seaweed is better for self-hosted Sentry. One thing I don't like about Garage is that we need to specify the storage allocation beforehand, if we set it to 100GB, there might be some people that have more data than 100GB, I don't want that to cause any issues.

That said, since you said you've used seaweed before: How was your experience? How does it compare to MinIO or Ceph?

> And thousands of thousands small files like profiles not ideal to store on most popular filesystems i guess.

Yeah if we set up an object storage, we might as well move filestore & profiles there too. But let's focus on nodestore first.

doc-sheet:
> How was your experience? How does it compare to MinIO or Ceph?

It is a bit strange sometimes. But it is fine.

It has [multiple options](https://github.com/seaweedfs/seaweedfs/wiki/Directories-and-Files) for filer store.
I didn't try leveldb storage aiming to fault tolerance.

At first I tried redis it worked for several months and then... I just lost all data.
It was there physically but wasn't available from API (s3 or web) - each list call returned different results.

I don't know if issue was in redis or weed itself. i suspect [bug with ttl](https://github.com/seaweedfs/seaweedfs/pull/6827) could be the reason too.

But after that incident I wiped cluster and started new one with scylla as a filer backend and it works fine for almost a year already despite that ttl bug.

Seaweedfs have multiple versions like
- 3.89
- 3.89_full
- 3.89_large_disk
- 3.89_large_disk_full

I suggest to use large_disk always. Documentation is not clear but it is easy to reach that limit
https://github.com/seaweedfs/seaweedfs/wiki/FAQ#how-to-configure-volumes-larger-than-30gb

I don't know difference between full and normal and just use `_large_disk_full` builds :)

Also I don't use s3 auth - I was too lazy to set it up.

Other than all that I have no problems and barely touched it after initial setup. It just works.
I added some volumes but not removed any yet.

As for minio and ceph.
I never used ceph.

But minio was the reason to look for alternatives.

Tons of profiles from js-sdk stored as different files started to affect my monitoring script and soon it might start to affect minio performance too.

And it is not that easy to scale minio. And probably impossible to optimize for small-files storage. At least in my low-cost setup.

doc-sheet:
> let's focus on nodestore first.

If seaweedfs would control ttl then there is another catch.
I'm not sure if it is possible to control ttl with s3-api already.

weed have it's own settings for collections and it creates collection for each s3-bucket.
https://github.com/seaweedfs/seaweedfs/wiki/S3-API-FAQ#setting-ttl

But if sentry itself would cleanup old data I guess there is no difference.

aldy505:
> > How was your experience? How does it compare to MinIO or Ceph?
> 
> It is a bit strange sometimes. But it is fine.
> 
> It has [multiple options](https://github.com/seaweedfs/seaweedfs/wiki/Directories-and-Files) for filer store. I didn't try leveldb storage aiming to fault tolerance.
> 
> At first I tried redis it worked for several months and then... I just lost all data. It was there physically but wasn't available from API (s3 or web) - each list call returned different results.
> 
> I don't know if issue was in redis or weed itself. i suspect [bug with ttl](https://github.com/seaweedfs/seaweedfs/pull/6827) could be the reason too.
> 
> But after that incident I wiped cluster and started new one with scylla as a filer backend and it works fine for almost a year already despite that ttl bug.
> 
> Seaweedfs have multiple versions like
> 
> * 3.89
> * 3.89_full
> * 3.89_large_disk
> * 3.89_large_disk_full
> 
> I suggest to use large_disk always. Documentation is not clear but it is easy to reach that limit https://github.com/seaweedfs/seaweedfs/wiki/FAQ#how-to-configure-volumes-larger-than-30gb
> 
> I don't know difference between full and normal and just use `_large_disk_full` builds :)
> 
> Also I don't use s3 auth - I was too lazy to set it up.
> 
> Other than all that I have no problems and barely touched it after initial setup. It just works. I added some volumes but not removed any yet.

Good to know about Seaweed

> As for minio and ceph. I never used ceph.
> 
> But minio was the reason to look for alternatives.
> 
> Tons of profiles from js-sdk stored as different files started to affect my monitoring script and soon it might start to affect minio performance too.
> 
> And it is not that easy to scale minio. And probably impossible to optimize for small-files storage. At least in my low-cost setup.

Ah so everyone has the same experience with minio.

> > let's focus on nodestore first.
> 
> If seaweedfs would control ttl then there is another catch. I'm not sure if it is possible to control ttl with s3-api already.
> 
> weed have it's own settings for collections and it creates collection for each s3-bucket. https://github.com/seaweedfs/seaweedfs/wiki/S3-API-FAQ#setting-ttl
> 
> But if sentry itself would cleanup old data I guess there is no difference.

The sentry cleanup job only cleans up the one on filesystem. If we're using S3, it won't clean up anything. We need to configure S3 data cleanup on our own.

doc-sheet:
Looks like i missed that seaweedfs now have an ability to control ttl with s3 api. And I even linked to correct section of FAQ. :)

I'd like to look into new integraton with seaweedfs.

Nad by the way I like [the idea](https://github.com/getsentry/self-hosted/issues/1808#issuecomment-2924227031) of expanding sentry images.

I am myself install some packages and modules.

Like maybe an extra step in install to build user provided Dockerfiles.

aldy505:
> Nad by the way I like [the idea](https://github.com/getsentry/self-hosted/issues/1808#issuecomment-2924227031) of expanding sentry images.
> 
> I am myself and install some packages and modules.
> 
> Like maybe an extra step in install to build user provided Dockerfiles.

Yes but I don't think people would go to non-default setup if they don't need anything.

aldy505:
This is interesting, very close to MinIO yet is far lightweight. https://github.com/rustfs/rustfs

aldy505:
I just tried trying to set things up with rustfs. It didn't work. I can't configure the administrative side of S3. See my PR here https://github.com/getsentry/self-hosted/pull/3821

aldy505:
@BYK Do you remember why we didn't use MinIO? Was it about licensing issue?

BYK:
>@BYK Do you remember why we didn't use MinIO? Was it about licensing issue?

Both licensing and reported performance and scalability issues reported by others.

aldy505:
Now, what's missing is the retention days. We can't change the retention days mid installation, the bucket needs to be recreated.

aldy505:
Integration tests isn't passing. I think we should hold this off for a bit.

hubertdeng123:
@aldy505 I also noticed that recently we've added objectstore. Perhaps related here?
https://github.com/getsentry/sentry/pull/97271

aldy505:
> @aldy505 I also noticed that recently we've added objectstore. Perhaps related here? [getsentry/sentry#97271](https://github.com/getsentry/sentry/pull/97271)

@hubertdeng123 I asked Jan last week, it's not being used on SaaS yet. Quoting him:

> Right, we're planning to make this an intermediary layer to some backend - we do not have a strong story for self hosted yet. We wouldn't be using postgres for sure, instead offer two alternatives: Any S3-compatible backend or raw disk.
> Our first use case is event attachments followed, by release files and debug files. We do consider replacing nodestore, but it's not on our roadmap yet. Will likely take months to get to the point where we can plan that .

aldy505:
Yay it's green.

One bad thing is that we need to tell people to have at least another 16GB set aside for swapfile.

aldy505:
Ah right, we're missing the lifecycle thing.

Previously, I've seen folks that increase the number of retention days. So if we want to keep that behaviour, it wouldn't be possible by setting S3 retention days (and modifying it every once in a while during installation). I can think of setting 1 cron container, but would it be a sensible thing to do?

BYK:
@aldy505

>We should have a proper migration path for existing installs

Don't think we have this either?

aldy505:
Config migration is done and is behind a prompt/flag. Next up is to think how to manage the retention. Should we do it with a cron/scheduled job (cons: probably heavy process, pros: can be configured dynamically), or a S3 lifecycle (cons: can't be configured dynamically, pros: won't be a heavy process).

aldy505:
I'm just gonna go forward with this https://github.com/seaweedfs/seaweedfs/wiki/S3-API-FAQ#setting-ttl

aldy505:
Great, I believe that's all.

aldy505:
@aminvakil @doc-sheet Hi, would you mind reviewing this PR?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
+echo "${_group}Bootstrapping seaweedfs (node store)..."
+
+$dc up --wait seaweedfs postgres
```
BYK, line 3: Does the `--wait` thing work for `podman`?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+$dc exec seaweedfs apk add --no-cache s3cmd
+$dc exec seaweedfs mkdir -p /data/idx/
+s3cmd="$dc exec seaweedfs s3cmd"
+
+bucket_list=$($s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' ls)
+
+if [[ $($bucket_list | tail -1 | awk '{print $3}') != 's3://nodestore' ]]; then
```
BYK, line 10: This `if` condition really needs some explanation about what it is checking and how -- like what are the assumptions here?

sentry/sentry.conf.example.py
```diff
@@ -95,6 +95,31 @@ def get_internal_network():
...
+# Previously, it used PostgreSQL as the backend, but this didn't scale for
+# high-throughput environments. Read more about this in the documentation:
+# https://develop.sentry.dev/backend/application-domains/nodestore/
+#
+# Through this setting, you can use the provided blob storage or
+# your own S3-compatible API from your infrastructure.
+# Other backend implementations for node storage developed by the community
+# are available in public GitHub repositories.
+
+SENTRY_NODESTORE = "sentry_nodestore_s3.S3PassthroughDjangoNodeStorage"
```
BYK, line 112: Do we need Passthrough for fresh installs?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+      echo "to modify your sentry.conf.py file contents."
+      echo "Do you want us to do it automatically for you?"
+      echo
+
+      yn=""
+      until [ ! -z "$yn" ]; do
+        read -p "y or n? " yn
+        case $yn in
+        y | yes | 1)
+          export APPLY_AUTOMATIC_CONFIG_UPDATES=1
```
BYK: This feels dangerous as once you export it like this wouldn't it also enable auto update for `pgbouncer`? I think that should be a separate questions and decision?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+          echo
+          echo -n "Thank you."
+          ;;
+        n | no | 0)
+          export APPLY_AUTOMATIC_CONFIG_UPDATES=0
+          echo
+          echo -n "Alright, you will need to update your sentry.conf.py file manually before running 'docker compose up'."
```
BYK, line 34: Again, provide details and link to this PR (or some doc). Also skip setting up seaweed at this point or at least mention that the service will be running for no good reason?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+      echo
+      echo "  APPLY_AUTOMATIC_CONFIG_UPDATES=1 to apply automatic updates"
+      echo "  APPLY_AUTOMATIC_CONFIG_UPDATES=0 to not apply automatic updates"
+      echo
+      sleep 5
+    fi
+
+    if [[ "$APPLY_AUTOMATIC_CONFIG_UPDATES" == 1 ]]; then
+      nodestore_config=$(sed -n '/SENTRY_NODESTORE/,/[}]/{p}' sentry/sentry.conf.example.py)
+      if [[ $($dc exec postgres psql -qAt -U postgres -c "select exists (select * from nodestore_node limit 1)") = "f" ]]; then
```
BYK, line 56: Up until this point you don't need the `postgres` service so maybe instead of bringing it up at the beginning, just use `$dcr postgres psql ...` here?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+        nodestore_config=$(echo -e "$nodestore_config" | sed '$s/\}/    "read_through": True,\n    "delete_through": True,\n\}/')
+      fi
+      echo "$nodestore_config" >>$SENTRY_CONFIG_PY
+    fi
+  fi
+
+  $dc exec seaweedfs mkdir -p /data/idx/
+  $s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' mb s3://nodestore
```
BYK, line 64: Why are these commands repeated (from lines 6 and 7 above)?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+        <Status>Enabled</Status>
+        <Filter></Filter>
+        <Expiration>
+            <Days>$SENTRY_EVENT_RETENTION_DAYS</Days>
+        </Expiration>
+    </Rule>
+</LifecycleConfiguration>
+EOF
+  )
+  $dc exec seaweedfs sh -c "printf '%s' '$lifecycle_policy' > /tmp/nodestore-lifecycle-policy.xml"
```
BYK, line 82: Yes, make it a config file and mount it? Is this because of the `$SENTRY_EVENT_RETENTION_DAYS` variable? If yes, we may wanna construct a custom entry point script for seaweed so when people change this value and restart, it is updated without running `./install.sh` again?

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+  $s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' setlifecycle /tmp/nodestore-lifecycle-policy.xml s3://nodestore
+
+  echo "Making sure the bucket lifecycle policy is all set up correctly..."
+  $s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' getlifecycle s3://nodestore
```
BYK, line 86: This seems it belongs to tests instead of here? Is this because of the unknown env variable value?

install/bootstrap-s3-nodestore.sh
aldy505: Restarting seaweed wouldn't take anything into effect. You will need to execute `s3cmd setlifecycle` for that

install/bootstrap-s3-nodestore.sh
aldy505: Good point. Will check later.

sentry/sentry.conf.example.py
aldy505: This is your original code, so I assume, yes.

install/bootstrap-s3-nodestore.sh
aldy505: Ah good point.

sentry/sentry.conf.example.py
BYK: I would not assume anything about my original code 😅

install/bootstrap-s3-nodestore.sh
BYK: And I'm saying we should put whatever is necessary into a custom entrypoint script for seaweed so it picks the changes up?

aldy505: I don't like that approach, it complicates stuff.

BYK: Okay, until someone compalins let's go with the simpler approach :)

aldy505:
> Can we make it opt-in for a release? And get some feedback regarding this change?

@aminvakil You're saying this shouldn't be the default? And probably switch it as a default at a later stage? That works for me.

BYK:
I'd say default for fresh installs. And since we prompt people for existing installs, it should be fine?

aminvakil:
@aldy505 Yes.

@BYK Yes.

I have zero knowledge about `seaweedfs` and very limited knowledge about S3 storages, therefore I'd suggest merging this after 15th, so people who use main branch can test it in different environments they have.

But again that's because my limited knowledge of S3 storages and I do not know who big this change is.

install/bootstrap-s3-nodestore.sh
```diff
@@ -0,0 +1,90 @@
...
+        <Filter></Filter>
+        <Expiration>
+            <Days>$SENTRY_EVENT_RETENTION_DAYS</Days>
+        </Expiration>
+    </Rule>
+</LifecycleConfiguration>
+EOF
+  )
+  $dc exec seaweedfs sh -c "printf '%s' '$lifecycle_policy' > /tmp/nodestore-lifecycle-policy.xml"
+  $s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' setlifecycle /tmp/nodestore-lifecycle-policy.xml s3://nodestore
```
aminvakil, line 83: What does `--region=us-east-1` mean here?

sentry/sentry.conf.example.py
```diff
@@ -95,6 +95,31 @@ def get_internal_network():
...
+SENTRY_NODESTORE_OPTIONS = {
+    "compression": True,
+    "endpoint_url": "http://seaweedfs:8333",
+    "bucket_path": "nodestore",
+    "bucket_name": "nodestore",
+    "region_name": "us-east-1",
```
aminvakil, line 118: Is `region_name` mandatory?

install/bootstrap-s3-nodestore.sh
aldy505: It's the default region name for AWS S3. It doesn't mean anything for self-hosted, but we can only put known region values.

sentry/sentry.conf.example.py
aldy505: Yep. Otherwise it'll complain of missing region name.

aldy505:
> @aldy505 Yes.
> 
> 
> 
> @BYK Yes.
> 
> 
> 
> I have zero knowledge about `seaweedfs` and very limited knowledge about S3 storages, therefore I'd suggest merging this after 15th, so people who use main branch can test it in different environments they have.
> 
> 
> 
> But again that's because my limited knowledge of S3 storages and I do not know who big this change is.

Great! I'll change it later tonight.

aldy505:
Uh, I mean lets do what @BYK said. I still prefer to merge all this before 15th.

PR #3506, opened 2025-01-02
ci: Only test on compose 2.26 w/ customizations

BYK:
Docker Compose is much more robust nowadays compared to the past where we had to maintain tests for both v1 and v2 and then a specific version of v2. Hence, we are removing tests for the older versions of Docker Compose with this patch. We also remove the separate tests for customizations and w/o customizations as the one with customizations should cover the one without them anyway. This reduces the CI workload to 25% of what it was


BYK:
To be able to land this, we need to change the required checks in repo configuration.

hubertdeng123 (approved):
seems about time 👍

PR #3516, opened 2025-01-09
ci: Faster and smarter backup/restore tests

BYK:
From
```
================== 11 passed, 4 warnings in 762.35s (0:12:42) ==================
```

to
```
================== 11 passed, 4 warnings in 343.58s (0:05:43) ==================
```


_integration-test/conftest.py
```diff
@@ -1,58 +1,22 @@
...
-
-def pytest_addoption(parser):
-    parser.addoption("--customizations", default="disabled")
-
 
 @pytest.fixture(scope="session", autouse=True)
 def configure_self_hosted_environment(request):
     subprocess.run(
-        ["docker", "compose", "--ansi", "never", "up", "-d"],
+        ["docker", "compose", "--ansi", "never", "up", "--wait"],
```
hubertdeng123, line 15: Is this needed anymore? Seeing https://github.com/getsentry/self-hosted/blob/63334cbcc23132e4d6832809ed399c75f48e4683/action.yaml#L135

BYK: I'm not sure but I'll try removing it and see what happens.

_integration-test/test_backup.py
```diff
@@ -42,21 +43,42 @@ def test_backup(setup_backup_restore_env_variables):
 
 def test_import(setup_backup_restore_env_variables):
     # Bring postgres down and recreate the docker volume
+    subprocess.run(["docker", "compose", "--ansi", "never", "down"], check=True)
+    for name in ("postgres", "clickhouse", "kafka"):
```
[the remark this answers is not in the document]
BYK: Just good hygine as this simulates a "clean db" which is what we want for import to work properly. I can leave them as they are which means we are restoring to a "dirty" DB (not just Postgres). I cannot just recreate them (delete/create) as they also need migrations, hence here we are.

I guess I should add a comment about this here then?

hubertdeng123 (approved):

PR #3519, opened 2025-01-10
ci: Move e2e test action into the repo

BYK:
This is an initial transitionary patch before landing #3516. Once we land this, we will update users of the old action to use this one and remove that repo. Then land #3516 safely.

Great thing is, with this patch and the subsequent update to getsentry/action-self-hosted-e2e-tests to use this one, all the repos would be using the Docker Volume caching we introduced in #3488.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,0 +1,149 @@
+name: "Sentry self-hosted end-to-end tests"
+inputs:
+  project_name:
+    required: false
+    description: "e.g. snuba, sentry, relay, self-hosted"
+  image_url:
+    required: false
+    description: "The URL to the built relay, snuba, sentry image to test against."
+  CODECOV_TOKEN:
+    required: false
+    description: "The Codecov token to upload coverage."
+
+runs:
+  using: "composite"
+  steps:
+    - name: Go into self-hosted directory
+      shell: bash
+      run: cd ${{ github.action_path }}
+
+    - name: Configure to use the test image
+      if: inputs.project_name && inputs.image_url
+      shell: bash
+      run: |
+        image_var=$(echo ${{ inputs.project_name }}_IMAGE | tr '[:lower:]' '[:upper:]')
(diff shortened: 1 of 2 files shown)


action.yaml
```diff
@@ -0,0 +1,149 @@
...
+          /var/lib/docker/volumes/sentry-kafka/_data
+
+    - name: Install self-hosted
+      env:
+        SKIP_DB_MIGRATIONS: ${{ steps.restore_cache.outputs.cache-hit == 'true' && '1' || '' }}
+      shell: bash
+      run: |
+        # This is for the cache restore on Kafka to work in older releases
+        docker run --rm -v "sentry-kafka:/data" busybox chown -R 1000:1000 /data
+        # Add some customizations to test that path
```
hubertdeng123, line 93: Is the goal here to test customizations as default for these tests? Currently the behavior is different so wanted to hear your thoughts here.

hubertdeng123 (commented):
> Once we land this, we will point https://github.com/getsentry/action-self-hosted-e2e-tests/blob/main/action.yml to this repo, land https://github.com/getsentry/self-hosted/pull/3516, then update users of the old action to use this one and remove that repo.

Is it needed to point https://github.com/getsentry/action-self-hosted-e2e-tests/blob/main/action.yml to this repo? Would it be better to put up PR's in snuba/relay directly to use the action.yaml workflow in the self-hosted repo to verify expected behavior before merging. Then, archive the `action-self-hosted-e2e-tests` repo once that looks stable

action.yaml
BYK: Yup. Just removing the double work and doing it with customizations at all times.

BYK:
@hubertdeng123 yup you're right. I also realized this just before your comment and updated the PR description (and the commit log)

hubertdeng123 (approved):
🚀

PR #3521, opened 2025-01-11
Introduce patches with external kafka

aldy505:
Continuation of https://github.com/getsentry/sentry-docs/pull/11847

### Legal Boilerplate


aldy505:
@BYK Do you want to take a look?

patches/external-kafka.sh
```diff
@@ -0,0 +1,250 @@
...
+  subscription-consumer-transactions:
+    <<: *sentry_defaults
+  subscription-consumer-metrics:
+    <<: *sentry_defaults
+  subscription-consumer-generic-metrics:
+    <<: *sentry_defaults
+EOF
+)
+if [[ -f "docker-compose.override.yml" ]]; then
+  echo "🚨 docker-compose.override.yml already exists. You will need to modify it manually:"
```
BYK, line 236: Maybe try to be more helpful by sharing _what_ needs to be changed in the file? Like "you need to remove the `kafka` service" or something among those lines.

BYK (commented):
Good start but I'd extract the patch files from the shell script for readability and maintainability.

You may rename the top-level `patches` folder you created to `optional-modifications` and then have an actual `patches` folder under that which has the patch files namespaces by the script name like `external-kafka/sentry.conf.py.patch` etc.

patches/external-kafka.sh
aldy505: Although I print the `docker-compose.override.yml` content, yeah, this might be helpful too.

aldy505:
I don't remember why this PR stalled. Do you remember? @BYK

BYK:
>I don't remember why this PR stalled. Do you remember? @BYK

It was the patch file vs inline discussion. I still think patch strings embedded in shell scripts is a bad idea for maintenance 😄

aldy505:
> >I don't remember why this PR stalled. Do you remember? @BYK
> 
> 
> 
> It was the patch file vs inline discussion. I still think patch strings embedded in shell scripts is a bad idea for maintenance 😄 

Ahhh! You're right. Lets _debate_ about that more! 😆😆

aldy505:
Just read this, I don't really like this approach though. You can't see the documentation/thought process on how to achieve _that_ state of configuration. https://stackoverflow.com/a/9982282

aldy505:
@BYK do you have any idea to bypass pre-commit on `.patch` files?

.pre-commit-config.yaml
```diff
@@ -21,4 +20,5 @@ repos:
   - id: check-symlinks
   - id: end-of-file-fixer
   - id: trailing-whitespace
+    exclude: '\\.patch$'
```
aldy505, line 23: I gave up.

aldy505:
There we go

aldy505:
@BYK @hubertdeng123 Can you help review these? Thanks!

BYK (approved):
We may wanna test these patches as I'm sure they'll go out of date pretty quickly.

aldy505:
> We may wanna test these patches as I'm sure they'll go out of date pretty quickly.

I'll put that on a separate PR. I want to find a way to make these patches not a blocker if something changes, especially if the person making that change doesn't know how to resolve the patch.

PR #3522, opened 2025-01-11
ci: Less volatile cache keys

BYK:
Instead of using direct image hashes, only use hashes from migrations folders for each respective image for cache key generation. Should increase cache hit rate significantly as we don't have migrations much.

Also swaps the key order from `sentry-snuba` to `snuba-senry` assuming Snuba has less frequent migration additions.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -62,6 +62,8 @@ runs:
         sudo chmod o+x /var/lib/docker
         sudo chmod -R o+rwx /var/lib/docker/volumes
         source ${{ github.action_path }}/.env
+        # See https://explainshell.com/explain?cmd=ls%20-Rv1rpq
+        # for that long `ls` command
         SENTRY_MIGRATIONS_MD5=$(docker run --rm --entrypoint bash $SENTRY_IMAGE -c 'ls -Rv1rpq src/sentry/migrations/' | md5sum | cut -d ' ' -f 1)
         echo "SENTRY_MIGRATIONS_MD5=$SENTRY_MIGRATIONS_MD5" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


action.yaml
```diff
@@ -62,19 +62,19 @@ runs:
         sudo chmod o+x /var/lib/docker
         sudo chmod -R o+rwx /var/lib/docker/volumes
         source ${{ github.action_path }}/.env
-        SENTRY_IMAGE_SHA=$(docker buildx imagetools inspect $SENTRY_IMAGE --format "{{println .Manifest.Digest}}")
-        echo "SENTRY_IMAGE_SHA=$SENTRY_IMAGE_SHA" >> $GITHUB_OUTPUT
-        SNUBA_IMAGE_SHA=$(docker buildx imagetools inspect $SNUBA_IMAGE --format "{{println .Manifest.Digest}}")
-        echo "SNUBA_IMAGE_SHA=$SNUBA_IMAGE_SHA" >> $GITHUB_OUTPUT
+        SENTRY_MIGRATIONS_MD5=$(docker run --rm --entrypoint bash $SENTRY_IMAGE -c 'ls -Rv1rpq src/sentry/migrations/' | md5sum | cut -d ' ' -f 1)
```
BYK, line 65:
```suggestion
        # See https://explainshell.com/explain?cmd=ls%20-Rv1rpq
        # for that long `ls` command
        SENTRY_MIGRATIONS_MD5=$(docker run --rm --entrypoint bash $SENTRY_IMAGE -c 'ls -Rv1rpq src/sentry/migrations/' | md5sum | cut -d ' ' -f 1)
```

hubertdeng123 (approved):

PR #3535, opened 2025-01-16
breaking: Upgrade min Compose version to 2.23.2

BYK:
In this version, there's a new `--pull` argument for `docker compose run` which we will start leveraging, especially with `sentry-admin` command.

Should come with a slight speed boost.


aldy505 (approved):

action.yaml
```diff
@@ -39,26 +39,13 @@ runs:
...
-        sudo curl -L https://github.com/docker/compose/releases/download/${{ env.COMPOSE_VERSION }}/docker-compose-`uname -s`-`uname -m` -o "${{ env.COMPOSE_PATH }}/docker-compose"
-        sudo chmod +x "${{ env.COMPOSE_PATH }}/docker-compose"
+      uses: ./get-compose-action
 
     - name: Compute Docker Volume Cache Key
       id: cache_key
       shell: bash
       run: |
-        source ${{ github.action_path }}/.env
+        source ${{ github.action_path }}.env
```
hubertdeng123, line 48: Is this intentional?

hubertdeng123 (approved):

action.yaml
BYK: Yup, that variable has a trailing slash which is a bit confusing 😅

aminvakil (approved):

aminvakil:
I opened https://github.com/getsentry/sentry-docs/pull/12376 which should be merged after this.

PR #3539, opened 2025-01-16
fix: Use correct path for get compose action

hubertdeng123:
self-hosted CI runs in other repos are failing since they're not using the right path for getting docker compose version as seen in https://github.com/getsentry/snuba/actions/runs/12818778331/job/35745427857?pr=6784

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -39,7 +39,7 @@ runs:
         fi
 
     - name: Get Compose
-      uses: ./get-compose-action
+      uses: ${{ github.action_path }}/get-compose-action
 
     - name: Compute Docker Volume Cache Keys
       id: cache_key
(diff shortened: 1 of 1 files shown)


aldy505 (approved):

BYK (approved):

action.yaml
```diff
@@ -39,7 +39,7 @@ runs:
         fi
 
     - name: Get Compose
-      uses: ./get-compose-action
+      uses: getsentry/self-hosted/get-compose-action@master
```
BYK, line 42: Actually, this should have been `${{ github.action_path }}get-compose-action` not sure how I missed this (or maybe a merge error). I'll submit a PR that does this and see if it works or fails.

BYK: Ah remembered now, can't do what I want due to: https://github.com/orgs/community/discussions/27048

hubertdeng123: I tried that one in this PR but discovered it didn't work haha

PR #3543, opened 2025-01-17
fix: Use local path for get-compose-action

BYK:
Follow up to #3539


BYK:
Ah, we cannot do this because: https://github.com/orgs/community/discussions/27048

hubertdeng123:
haha I tried this too

PR #3588, opened 2025-02-25
refactor: move system.url-prefix under systems settings section

leeoocca:
Not sure why, but `system.url-prefix` and `system.internal-url-prefix` were not under `System Settings` in sentry/config.example.yml

The PR simply moves the three lines to the right section.

### Legal Boilerplate

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -45,6 +45,10 @@ mail.host: 'smtp'
 # System Settings #
 ###################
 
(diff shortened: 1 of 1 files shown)


aldy505 (approved):

aminvakil (approved):

BYK (approved):

BYK:
Probably because `internal-url` was added later on, so an oversight. Thanks for the fix!

PR #3601, opened 2025-03-04
Fix unbound variable error in install script

brettdh:
If the docker-compose version was less than the minimum required version, the install script would error out like this:

```
❯ ./install.sh
<snip>
▶ Checking minimum requirements ...
Found Docker version 27.5.1
install/check-minimum-requirements.sh: line 17: install: unbound variable
```

With this patch, the script correctly identifies the out-of-date docker-compose version and bails out with a more helpful message:
```
❯ ./install.sh
<snip>
▶ Checking minimum requirements ...
Found Docker version 27.5.1
FAIL: Expected minimum docker compose version to be 2.132.2 but found 2.32.4-desktop.1
```

(Note that `2.132.2` is not a real docker-compose version; I've edited the `install/_min-requirements.sh` file locally for testing, to make the installed version out of date.)

Also tested that the success case still works (by removing the aforementioned local edit):
```
❯ ./install.sh
<snip>
▶ Checking minimum requirements ...
Found Docker version 27.5.1
Found Docker Compose version 2.32.4-desktop.1
```

### Legal Boilerplate

### DIFF
--- a/install/_lib.sh
+++ b/install/_lib.sh
@@ -56,7 +56,6 @@ function ensure_file_from_example {
 # Check the version of $1 is greater than or equal to $2 using sort. Note: versions must be stripped of "v"
 function vergte() {
   printf "%s\n%s" $1 $2 | sort --version-sort --check=quiet --reverse
-  echo $?
 }
 
 SENTRY_CONFIG_PY=sentry/sentry.conf.py
--- a/install/check-minimum-requirements.sh
+++ b/install/check-minimum-requirements.sh
@@ -8,13 +8,13 @@ if [[ -z "$DOCKER_VERSION" ]]; then
   exit 1
 fi
 
-if [[ "$(vergte ${DOCKER_VERSION//v/} $MIN_DOCKER_VERSION)" -eq 1 ]]; then
+if ! vergte ${DOCKER_VERSION//v/} $MIN_DOCKER_VERSION; then
   echo "FAIL: Expected minimum docker version to be $MIN_DOCKER_VERSION but found $DOCKER_VERSION"
   exit 1
 fi
 echo "Found Docker version $DOCKER_VERSION"
 
-if [[ "$(vergte ${COMPOSE_VERSION//v/} $MIN_COMPOSE_VERSION)" -eq 1 ]]; then
+if ! vergte ${COMPOSE_VERSION//v/} $MIN_COMPOSE_VERSION; then
   echo "FAIL: Expected minimum $dc_base version to be $MIN_COMPOSE_VERSION but found $COMPOSE_VERSION"
   exit 1
 fi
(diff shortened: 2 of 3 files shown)


aminvakil (changes requested):
I think the problem relies somewhere else, why `COMPOSE_VERSION` should not be set and therefore inbound variable, maybe it's related to https://github.com/getsentry/self-hosted/pull/3595 ?

@brettdh Can you please paste output of these two commands?
```
docker-compose version
docker compose version
```

BYK:
Not sure why or how this happens but should be solved with #3604.

brettdh:
@aminvakil 

```
❯ docker-compose version
Docker Compose version v2.32.4-desktop.1

❯ docker compose version
Docker Compose version v2.32.4-desktop.1
```

brettdh:
I figured out the root cause by adding `| tee ./vergte.out` to the `$(vergte ...)` invocation in the script and line named in the error output. Here's the content:

```
❯ cat vergte.out
Error in install/_lib.sh:57.
'sort --version-sort --check=quiet --reverse' exited with status 1
-> ./install.sh:main:20
--> install/check-minimum-requirements.sh:source:17
---> install/_lib.sh:vergte:57

Cleaning up...
1
```

Because `cleanup()` in `install/error-handling.sh` prints to stdout rather than stderr, this output gets printed as part of `vergte`, then the caller captures it and treats part of it as a variable named `install`, which is unbound, hence the error.

Changing `vergte` to just rely on its own exit code rather than forwarding the exit code of `sort` via stdout resolves this. Changing `cleanup()` to print to stderr would also likely resolve it, but I don't see a reason that `vergte` needs to use stdout when the exit code is all that matters.

aminvakil:
There was a PR changing `vergte` function, so that it would not rely on return code, but I cannot find and don't remember why it wasn't merged, maybe I was dreaming someone had changed it ? :)

BYK:
Okay, if tests etc pass I think I'm convinced that this is a better approach.

Huge props to @brettdh for the detective work and the patience to explain it to us!

BYK (approved):

brettdh:
Fixed the formatting violations that pre-commit flagged.

install.sh
```diff
@@ -19,6 +19,9 @@ trap_with_arg cleanup ERR INT TERM EXIT
 source install/check-latest-commit.sh
 source install/check-minimum-requirements.sh
 
+echo "BAILING OUT EARLY"
```
hubertdeng123, line 22: This does not look like this should be here 🤔

hubertdeng123:
Other than that, this change seems reasonable

install.sh
brettdh: 🤦 thank you; my `git commit -a` mistake

brettdh:
I think I'm fighting with vscode auto save behavior on the formatting issues. Sorry for the noise; will actually set up pre-commit and fix it that way soon.

brettdh:
Ok it should be good now, including formatting and conflict resolution with recent changes on `master`.

aminvakil (approved):

aldy505 (approved):

BYK (approved):

PR #3604, opened 2025-03-05
ref: Less complicated docker compose detection

BYK:
With #3595, we now check both `docker-compose` and `docker compose` versions so this patch removes the implicit fallback to `docker-compose` for `$dc_base` and makes it explicit.

### DIFF
--- a/install/dc-detect-version.sh
+++ b/install/dc-detect-version.sh
@@ -9,22 +9,20 @@ fi
 echo "${_group}Initializing Docker Compose ..."
 
 # To support users that are symlinking to docker-compose
-dc_base="$(docker compose version &>/dev/null && echo 'docker compose' || echo 'docker-compose')"
+dc_base="$(docker compose version &>/dev/null && echo 'docker compose' || echo '')"
 dc_base_standalone="$(docker-compose version &>/dev/null && echo 'docker-compose' || echo '')"
 
-COMPOSE_VERSION=$($dc_base version --short || echo '')
-STANDALONE_COMPOSE_VERSION=$($dc_base_standalone version --short &>/dev/null || echo '')
+COMPOSE_VERSION=$([ -z "$dc_base" ] && $dc_base version --short || echo '')
+STANDALONE_COMPOSE_VERSION=$([ -z "$dc_base_standalone" ] && $dc_base_standalone version --short &>/dev/null || echo '')
 
 if [[ -z "$COMPOSE_VERSION" && -z "$STANDALONE_COMPOSE_VERSION" ]]; then
   echo "FAIL: Docker Compose is required to run self-hosted"
   exit 1
 fi
 
-if [[ ! -z "${STANDALONE_COMPOSE_VERSION}" ]]; then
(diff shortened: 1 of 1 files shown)


install/dc-detect-version.sh
```diff
@@ -9,22 +9,20 @@ fi
 echo "${_group}Initializing Docker Compose ..."
 
 # To support users that are symlinking to docker-compose
-dc_base="$(docker compose version &>/dev/null && echo 'docker compose' || echo 'docker-compose')"
+dc_base="$(docker compose version &>/dev/null && echo 'docker compose' || echo '')"
 dc_base_standalone="$(docker-compose version &>/dev/null && echo 'docker-compose' || echo '')"
 
-COMPOSE_VERSION=$($dc_base version --short || echo '')
-STANDALONE_COMPOSE_VERSION=$($dc_base_standalone version --short &>/dev/null || echo '')
+COMPOSE_VERSION=$([ -z "$dc_base" ] && $dc_base version --short || echo '')
```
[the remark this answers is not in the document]
BYK: This was a logic error on my end, switched to `-n` instead of `-z` which tests the opposite. Should be fine now, waiting for tests.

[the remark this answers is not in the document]
BYK: Which indeed passed 🥳

aminvakil (approved):

install/dc-detect-version.sh
```diff
@@ -9,22 +9,20 @@ fi
 echo "${_group}Initializing Docker Compose ..."
 
 # To support users that are symlinking to docker-compose
-dc_base="$(docker compose version &>/dev/null && echo 'docker compose' || echo 'docker-compose')"
+dc_base="$(docker compose version --short &>/dev/null && echo 'docker compose' || echo '')"
 dc_base_standalone="$(docker-compose version &>/dev/null && echo 'docker-compose' || echo '')"
```
hubertdeng123, line 13: Might as well use --short here too 
```suggestion
dc_base_standalone="$(docker-compose version --short &>/dev/null && echo 'docker-compose' || echo '')"
```

install/dc-detect-version.sh
```diff
@@ -9,22 +9,20 @@ fi
...
 
-COMPOSE_VERSION=$($dc_base version --short || echo '')
-STANDALONE_COMPOSE_VERSION=$($dc_base_standalone version --short &>/dev/null || echo '')
+COMPOSE_VERSION=$([ -n "$dc_base" ] && $dc_base version --short || echo '')
```
hubertdeng123, line 15: Personally, I think this is more clear for this line and the one below it, but I will leave that up to you:

```
if [ -n "$dc_base" ]; then
  COMPOSE_VERSION=$($dc_base version --short || echo '')
fi
```

hubertdeng123 (approved):

install/dc-detect-version.sh
hubertdeng123: I'm ok with it as it is

PR #3611, opened 2025-03-10
feat: provide monitoring-related configurations

aldy505:
In accordance with https://github.com/getsentry/sentry-docs/pull/12660

### Legal Boilerplate

### DIFF
--- a/relay/config.example.yml
+++ b/relay/config.example.yml
@@ -18,3 +18,14 @@ processing:
 #
 # health:
 #  max_memory_percent: 1.0
+
+# If you have statsd server, you can utilize that to monitor self-hosted Relay.
+# To start, uncomment the following line and adjust the options as needed.
+#
+# metrics:
+#  statsd: "100.100.123.123:8125" # It's recommended to use IP address instead of domain name
+#  prefix: "sentry.relay" # Adjust this to your needs, default is "sentry.relay"
+#  sample_rate: 1.0 # Adjust this to your needs, default is 1.0
+#  # `periodic_secs` is the interval for periodic metrics emitted from Relay.
+#  #  Setting it to `0` seconds disables the periodic metrics.
+#  periodic_secs: 5
--- a/relay/config.example.yml
+++ b/relay/config.example.yml
@@ -23,7 +23,7 @@ processing:
 # To start, uncomment the following line and adjust the options as needed.
 #
 # metrics:
-#  statsd: "100.100.123.123:8125" # It's recommended to use IP address instead of domain name
+#  statsd: "100.100.123.123:8125" # It is recommended to use IP address instead of domain name
 #  prefix: "sentry.relay" # Adjust this to your needs, default is "sentry.relay"
 #  sample_rate: 1.0 # Adjust this to your needs, default is 1.0
 #  # `periodic_secs` is the interval for periodic metrics emitted from Relay.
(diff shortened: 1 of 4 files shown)


relay/config.example.yml
```diff
@@ -18,3 +18,14 @@ processing:
 #
 # health:
 #  max_memory_percent: 1.0
+
+# If you have statsd server, you can utilize that to monitor self-hosted Relay.
+# To start, uncomment the following line and adjust the options as needed.
+#
+# metrics:
+#  statsd: "100.100.123.123:8125" # It's recommended to use IP address instead of domain name
```
aminvakil, line 26:
```suggestion
#  statsd: "100.100.123.123:8125" # It is recommended to use IP address instead of domain name
```

sentry/sentry.conf.example.py
```diff
@@ -396,3 +396,21 @@ def get_internal_network():
...
+
+# By default, Sentry uses dummy statsd monitoring backend that is a no-op.
+# If you have a statsd server, you can utilize that to monitor self-hosted
+# Sentry for "sentry"-related containers.
+#
+# To start, uncomment the following line and adjust the options as needed.
+
+# SENTRY_METRICS_BACKEND = 'sentry.metrics.statsd.StatsdMetricsBackend'
+# SENTRY_METRICS_OPTIONS: dict[str, Any] = {
+#     'host': '100.100.123.123', # It's recommended to use IP address instead of domain name
```
aminvakil, line 412:
```suggestion
#     'host': '100.100.123.123', # It is recommended to use IP address instead of domain name
```

symbolicator/config.example.yml
```diff
@@ -6,3 +6,10 @@ logging:
 metrics:
   statsd: null
 sentry_dsn: null # TODO: Automatically fill this with the internal project DSN
+
+# If you have statsd server, you can utilize that to monitor self-hosted Symbolicator.
+# To start, uncomment the following line and adjust the options as needed.
+#
+# metrics:
+#   statsd: "100.100.123.123:8125" # It's recommended to use IP address instead of domain name
```
aminvakil, line 14:
```suggestion
#   statsd: "100.100.123.123:8125" # It is recommended to use IP address instead of domain name
```

symbolicator/config.example.yml
```diff
@@ -6,3 +6,10 @@ logging:
...
+#   prefix: "sentry.symbolicator" # Adjust this to your needs, default is "symbolicator"
```
aminvakil, line 15: I'm confused, if default is `symbolicator`, let's have it here as well, right?
Or is the default `sentry.symbolicator`, then change the comment?

```suggestion
#   prefix: "symbolicator" # Adjust this to your needs, default is "symbolicator"
```

aminvakil (commented):
There is this rule in ansible, and I'm used to it, I don't know if applies to sentry as well, but it's good IMHO.

https://docs.ansible.com/ansible/latest/dev_guide/style_guide/grammar_punctuation.html#contractions

symbolicator/config.example.yml
aldy505: Default is "symbolicator", but I prefer everything to be prefixed with "sentry.", just to put everything nicely

BYK (approved):
Love this, thank you so much! 🚀

Issue #3614, opened 2025-03-13
SETUP_JS_SDK_ASSETS=1 wrong folder permissions


felixbuenemann:
### Self-Hosted Version

25.2.0

### CPU Architecture

x86_64

### Docker Version

28.0.1

### Docker Compose Version

2.33.1

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

After enabling `SETUP_JS_SDK_ASSETS=1` in `.env.custom` and re-running `./install.sh` the installer script `install/setup-js-sdk-assets.sh` downloads Sentry SDK assets into the docker volume "sentry-nginx-www", which is then mounted into the nginx container at `/var/www/`.

However since the folders are created with 750 permisions and owned by root:root, the `nginx` process rnning as user nginx cannot read the files and logs a permission denied and returns a 403 error.

After a `chmod o+rx /var/www/jd-sdk /var/www/js-sdk/*` in the container, the files are served successfully. Maybe `umask` is set incorrectly, while the install script is run in the container.

### Expected Result

The SDK files should be served without error.

### Actual Result

The SDK files are served with a 403 error:

```
curl -I http://localhost/js-sdk/8.55.0/bundle.tracing.replay.min.js
HTTP/1.1 403 Forbidden
Server: nginx
Date: Thu, 13 Mar 2025 20:36:58 GMT
Content-Type: text/html
Content-Length: 146
Connection: keep-alive
```

Nginx error log:

```
2025/03/13 20:36:58 [error] 36#36: *987 open() "/var/www/js-sdk/8.55.0/bundle.tracing.replay.min.js" failed (13: Permission denied), client: 127.0.0.1, server: , request: "HEAD /js-sdk/8.55.0/bundle.tracing.replay.min.js HTTP/1.1", host: "localhost"
```

aldy505:
> However since the folders are created with 750 permisions and owned by root:root, the nginx process rnning as user nginx cannot read the files and logs a permission denied and returns a 403 error.
>
> After a chmod o+rx /var/www/jd-sdk /var/www/js-sdk/* in the container, the files are served successfully. Maybe umask is set incorrectly, while the install script is run in the container.

This is interesting. Do you happen to use any specific volume driver for Docker?

aldy505:
On another note, if you want to create a PR before me, you're free to do so! :D

felixbuenemann:
> Do you happen to use any specific volume driver for Docker?

[...]

PR #3616, opened 2025-03-14
fix: js-sdk directory/file permission should be set correctly

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/3614

### Legal Boilerplate

### DIFF
--- a/_unit-test/js-sdk-assets-test.sh
+++ b/_unit-test/js-sdk-assets-test.sh
@@ -32,7 +32,7 @@ echo "Pass"
 
 # Files should be owned by the nginx user
 echo "Testing file ownership"
-test $($sdk_files | awk '$3=="nginx" { print $0 }' | wc -l) == "6"
+test $($sdk_files | awk '$3=="root" { print $0 }' | wc -l) == "6"
 echo "Pass"
 
 report_success
--- a/_unit-test/js-sdk-assets-test.sh
+++ b/_unit-test/js-sdk-assets-test.sh
@@ -34,7 +34,7 @@ echo "Pass"
 echo "Testing file ownership"
 directory_owners=$(echo "$sdk_files" | awk '$3=="root" { print $0 }' | wc -l)
 echo "$directory_owners"
-test "$directory_owners" == "6"
+test "$directory_owners" == "8"
 echo "Pass"
 
 report_success
--- a/_unit-test/js-sdk-assets-test.sh
+++ b/_unit-test/js-sdk-assets-test.sh
@@ -30,9 +30,11 @@ echo "Testing file sizes"
 test "23" == "$non_empty_file_count"
 echo "Pass"
 
-# Files should be owned by the nginx user
+# Files should be owned by the root user
 echo "Testing file ownership"
-test $($sdk_files | awk '$3=="root" { print $0 }' | wc -l) == "6"
+directory_owners=$(echo "$sdk_files" | awk '$3=="root" { print $0 }' | wc -l)
+echo "$directory_owners"
+test "$directory_owners" == "6"
 echo "Pass"
 
 report_success
--- a/_unit-test/js-sdk-assets-test.sh
+++ b/_unit-test/js-sdk-assets-test.sh
@@ -30,4 +30,9 @@ echo "Testing file sizes"
 test "17" == "$non_empty_file_count"
 echo "Pass"
 
+# Files should be owned by the nginx user
+echo "Testing file ownership"
+test $($sdk_files | awk '$3=="nginx" { print $0 }' | wc -l) == "5"
+echo "Pass"
+
 report_success
--- a/_unit-test/js-sdk-assets-test.sh
+++ b/_unit-test/js-sdk-assets-test.sh
@@ -20,19 +20,19 @@ echo $total_directories
 test "5" == "$total_directories"
 echo "Pass"
 
-# `sdk_tree` should output "5 directories, 17 files"
+# `sdk_tree` should output "6 directories, 23 files"
 echo "$sdk_tree"
-test "5 directories, 17 files" == "$(echo "$sdk_tree")"
+test "6 directories, 23 files" == "$(echo "$sdk_tree")"
 echo "Pass"
 
 # Files should all be >1k (ensure they are not empty)
 echo "Testing file sizes"
-test "17" == "$non_empty_file_count"
+test "23" == "$non_empty_file_count"
 echo "Pass"
 
 # Files should be owned by the nginx user
 echo "Testing file ownership"
-test $($sdk_files | awk '$3=="nginx" { print $0 }' | wc -l) == "5"
+test $($sdk_files | awk '$3=="nginx" { print $0 }' | wc -l) == "6"
 echo "Pass"
 
 report_success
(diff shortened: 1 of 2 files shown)


BYK:
I'm surprised that this wasn't caught by the tests. We should look into that too.

BYK (changes requested):
Can we not resolve the issue by just running the curl command with the nginx user instead?

aldy505:
> Can we not resolve the issue by just running the curl command with the nginx user instead?

Is it not?

BYK:
>Is it not?

Don't know? If it is, we should not have this issue shall we? At least that's what I get from the original issue report?

install/setup-js-sdk-assets.sh
```diff
@@ -36,5 +36,10 @@ if [[ "${SETUP_JS_SDK_ASSETS:-}" == "1" ]]; then
   # Download those versions & variants using curl
   $dcr --no-deps --rm -v "sentry-nginx-www:/var/www" nginx curl -w '%{response_code} %{url}\n' --no-progress-meter --compressed --retry 3 --create-dirs -fLo "/var/www/js-sdk/#1/#2.min.js" "https://browser.sentry-cdn.com/${versions}/${variants}.min.js" || true
```
aminvakil, line 37: This will do the trick as said by @BYK in https://github.com/getsentry/self-hosted/pull/3616#pullrequestreview-2684920970.
```suggestion
  $dcr --no-deps --rm -v "sentry-nginx-www:/var/www" --user=nginx nginx curl -w '%{response_code} %{url}\n' --no-progress-meter --compressed --retry 3 --create-dirs -fLo "/var/www/js-sdk/#1/#2.min.js" "https://browser.sentry-cdn.com/${versions}/${variants}.min.js" || true
```

aldy505: I haven't got time to be on my computer again (I'm on mobile now), but IIRC the user should be "www-data" instead of "nginx". But I'll recheck this later tomorrow.

aminvakil: I checked and it's nginx.

aminvakil: BTW we should check this doesn't get on an old nginx container which already has the files owned by root, as it will then if it wants to get overwritten by nginx user.

I think it's a temporary volume which does not persist during install.sh, but we should make sure.

aldy505: > BTW we should check this doesn't get on an old nginx container which already has the files owned by root, as it will then if it wants to get overwritten by nginx user.
> 
> I think it's a temporary volume which does not persist during install.sh, but we should make sure.

I'm a bit unsure now. Let's try this one first.

aldy505:
Ah we got this on CI

```
Found JS SDKs: v4.6.6, v5.30.0, v6.19.7, v7.120.3, v8.55.0
  curl: Error creating directory /var/www/js-sdk/4.6.6
  curl: (23) Failed writing received data to disk/application
  000 https://browser.sentry-cdn.com/4.6.6/bundle.min.js
  exit status 23
  find: /var/www/js-sdk: No such file or directory
  exit status 1
  fail 👎 with exit code 1

```

install/setup-js-sdk-assets.sh
```diff
@@ -34,7 +34,12 @@ if [[ "${SETUP_JS_SDK_ASSETS:-}" == "1" ]]; then
   variants="{bundle,bundle.tracing,bundle.tracing.replay,bundle.replay,bundle.tracing.replay.feedback,bundle.feedback}"
 
   # Download those versions & variants using curl
-  $dcr --no-deps --rm -v "sentry-nginx-www:/var/www" nginx curl -w '%{response_code} %{url}\n' --no-progress-meter --compressed --retry 3 --create-dirs -fLo "/var/www/js-sdk/#1/#2.min.js" "https://browser.sentry-cdn.com/${versions}/${variants}.min.js" || true
+  $dcr --no-deps --rm -v "sentry-nginx-www:/var/www" --user=nginx nginx curl -w '%{response_code} %{url}\n' --no-progress-meter --compressed --retry 3 --create-dirs -fLo "/var/www/js-sdk/#1/#2.min.js" "https://browser.sentry-cdn.com/${versions}/${variants}.min.js" || true
```
aminvakil, line 37: What do you think of adding this prior to downloading files by `nginx` user?
```
$dcr --no-deps --rm -v "sentry-nginx-www:/var/www" nginx mkdir -p "/var/www/js-sdk/#1"
```

aldy505:
Okay just checked mine. It's owned by `root`, not `nginx`.

image

aldy505:
The release window is closing, let's just choose the easy route for now.

install/setup-js-sdk-assets.sh
```diff
@@ -27,14 +27,20 @@ if [[ "${SETUP_JS_SDK_ASSETS:-}" == "1" ]]; then
   latest_js_v6=$(echo "$loader_registry" | $jq -r '.versions | reverse | map(select(.|any(.; startswith("6.")))) | .[0]')
   latest_js_v7=$(echo "$loader_registry" | $jq -r '.versions | reverse | map(select(.|any(.; startswith("7.")))) | .[0]')
   latest_js_v8=$(echo "$loader_registry" | $jq -r '.versions | reverse | map(select(.|any(.; startswith("8.")))) | .[0]')
+  latest_js_v9=$(echo "$loader_registry" | $jq -r '.versions | reverse | map(select(.|any(.; startswith("9.")))) | .[0]')
```
BYK, line 30: sneaky sneaky?

BYK (approved):
Merging for the release, please let's revisit this immediately after and find a better solution.

_unit-test/js-sdk-assets-test.sh
```diff
@@ -30,4 +30,9 @@ echo "Testing file sizes"
 test "17" == "$non_empty_file_count"
 echo "Pass"
 
+# Files should be owned by the nginx user
+echo "Testing file ownership"
+test $($sdk_files | awk '$3=="nginx" { print $0 }' | wc -l) == "5"
```
aldy505:
```suggestion
test $($sdk_files | awk '$3=="root" { print $0 }' | wc -l) == "5"
```

install/setup-js-sdk-assets.sh
BYK: well you just broke unit tests to start? :D

_unit-test/js-sdk-assets-test.sh
```diff
@@ -20,14 +20,19 @@ echo $total_directories
...
+test "6 directories, 23 files" == "$(echo "$sdk_tree")"
 echo "Pass"
 
 # Files should all be >1k (ensure they are not empty)
 echo "Testing file sizes"
-test "17" == "$non_empty_file_count"
+test "23" == "$non_empty_file_count"
+echo "Pass"
+
+# Files should be owned by the nginx user
```
aminvakil, line 33:
```suggestion
# Files should be owned by the root user
```

aldy505:
Since 25.3.0 is released and this is not merged yet. Let's think of a proper solution instead.

BYK:
@aldy505 still merged as it fixes a problem and adds a test for the issue. Trusting you to follow up with a better solution.

PR #3634, opened 2025-03-25
Minimum requirements for 'errors-only' profile

madest92:
Using the [errors-only](https://develop.sentry.dev/self-hosted/experimental/errors-only/) profile, fewer resources are required. About 2 times.

### DIFF
--- a/install/_min-requirements.sh
+++ b/install/_min-requirements.sh
@@ -5,9 +5,9 @@ MIN_COMPOSE_VERSION='2.32.2'
 # 16 GB minimum host RAM, but there'll be some overhead outside of what
 # can be allotted to docker
 if [[ "$COMPOSE_PROFILES" == "errors-only" ]]; then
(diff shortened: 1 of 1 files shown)


aldy505:
Does the number of RAM & CPU works on your instance (which I assume is running the errors-only mode)? Because I assume Kafka and Clickhouse would still take a lot of resources.

BYK:
I'll leave this to @aldy505. Please ping me when you think this is ready to be merged.

madest92:
Yes, it works. At startup, of course, the CPU works at full capacity. And it takes a little more time to fully launch the Sentry.
image
image

aldy505:
That's wonderful! Thank you for trying errors-only and reporting back to us! 😄

aldy505 (approved):

BYK (approved):
Thanks so much for helping us and the community!

PR #3635, opened 2025-03-26
chore(relay): specify spool.enveloppe.max_backpressure_memory_percent configuration for handling relay's failing healthcheck

aldy505:
Although a fix is being rolled out, that does not mean every relay instance would suddenly be fixed. We would need to still provide a workaround for people to try out. Refer to this specific issue comment: https://github.com/getsentry/self-hosted/issues/3330#issuecomment-2751092219

### Legal Boilerplate

### DIFF
--- a/relay/config.example.yml
+++ b/relay/config.example.yml
@@ -21,6 +21,7 @@ processing:
 #  max_memory_percent: 1.0
 # spool:
 #   envelopes:
+#     path: "/your/path"
 #     max_backpressure_memory_percent: 1.0
 
 # If you have statsd server, you can utilize that to monitor self-hosted Relay.
--- a/relay/config.example.yml
+++ b/relay/config.example.yml
@@ -14,10 +14,14 @@ processing:
 
 # In some cases, relay might fail to find out the actual machine memory
 # therefore it makes the healthcheck fail and events can't be submitted.
+# See https://github.com/getsentry/self-hosted/issues/3330
 # As a workaround, uncomment the following line:
 #
 # health:
 #  max_memory_percent: 1.0
+# spool:
+#   envelopes:
+#     max_backpressure_memory_percent: 1.0
 
 # If you have statsd server, you can utilize that to monitor self-hosted Relay.
 # To start, uncomment the following line and adjust the options as needed.
--- a/relay/config.example.yml
+++ b/relay/config.example.yml
@@ -14,18 +14,18 @@ processing:
 
 # In some cases, relay might fail to find out the actual machine memory
 # therefore it makes the healthcheck fail and events can't be submitted.
-# See https://github.com/getsentry/self-hosted/issues/3330
-# As a workaround, uncomment the following line:
+# See https://github.com/getsentry/self-hosted/issues/3330 for more details.
+# As a workaround, uncomment the following `health` and `spool` sections:
 #
 # health:
 #  max_memory_percent: 1.0
 # spool:
 #   envelopes:
-#     path: "/your/path"
+#     path: "/tmp/relay-spool-envelopes"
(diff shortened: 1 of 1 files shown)


iambriccardo:
To add more context, which might be helpful for others experiencing the same issue, Relay has two memory settings:
* `max_memory_percent` – This is the threshold at which health check failures will occur, and Relay will drop any incoming envelopes.
* `max_backpressure_memory_percent` – This threshold is used by Relay's internal buffer (only when configured to spool to disk) to determine when the buffer should stop producing events for downstream services. This effectively stops events from passing through Relay, but they will accumulate in memory and be routinely spooled to disk. This mechanism increases Relay’s resilience in case of upstream failures or backlogs in internal service queues.

When both are set to 100% (1.0), the system will continue accepting data and the buffer will forward it without issues, but this will eventually cause the process to run out of memory (OOM).

As noted before, the buffer will stop producing events downstream only if `max_backpressure_memory_percent` is reached and a `path` is specified. Specifying a `path` instructs Relay to spool to a SQLite database, which will be created at that location. If no `path` is provided, the `max_backpressure_memory_percent` will not be read by Relay.

relay/config.example.yml
```diff
@@ -14,10 +14,14 @@ processing:
...
 # In some cases, relay might fail to find out the actual machine memory
 # therefore it makes the healthcheck fail and events can't be submitted.
+# See https://github.com/getsentry/self-hosted/issues/3330
 # As a workaround, uncomment the following line:
 #
 # health:
 #  max_memory_percent: 1.0
+# spool:
+#   envelopes:
+#     max_backpressure_memory_percent: 1.0
```
iambriccardo, line 24: You must specify a path where Relay will spool events, only in that case the memory check will be used to determine whether to stop producing events downstream:
```suggestion
# spool:
#   envelopes:
#     path: "/your/path"
#     max_backpressure_memory_percent: 1.0
```

BYK (changes requested):
Rejecting for my comment and mostly for @iambriccardo's comment.

relay/config.example.yml
```diff
@@ -14,10 +14,15 @@ processing:
...
 # In some cases, relay might fail to find out the actual machine memory
 # therefore it makes the healthcheck fail and events can't be submitted.
+# See https://github.com/getsentry/self-hosted/issues/3330
 # As a workaround, uncomment the following line:
 #
 # health:
 #  max_memory_percent: 1.0
+# spool:
+#   envelopes:
+#     path: "/your/path"
```
aminvakil, line 24: Can we have an example of this path?

aminvakil (commented):
What is the default path? Is there a on?

relay/config.example.yml
aldy505: It should just be a path residing on the Docker volume... or since relay doesn't have a persisted volume, probably it's best to use `/tmp/foo` directory?

aminvakil: I'm quite unfamiliar with this change, so could you please bear with me and tell me, is it going to be more than 4GB?

I've seen odd situations when using `/tmp` and filling it with files bigger than 4GB size.

If that's not the case, then correct, `/tmp/foo` seems right to me.

BYK: Who not mount a volume to `relay` for this?

aldy505: From the docs, the default size is 500MB. We should be fine, no need for another volume. https://docs.sentry.io/product/relay/options/#spooling

aldy505:
Ping @aminvakil @BYK

aminvakil:
> Ping @aminvakil @BYK

I don't have any other comment or anything.

BYK (approved):

PR #3673, opened 2025-04-22
feat(install): Adds support for podman(compose)

DuncanConroy:
Adds support for podman(compose), while maintaining compatibility with docker

Introduces a new script to detect podman vs. docker. Distinguishes between docker and podman minimum versions and substitutes uses of docker with a variable instead.

Closes https://github.com/getsentry/self-hosted/issues/369

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #3710, opened 2025-05-24
add shellcheck action to lint bash scripts

doc-sheet:
Hello.
I would like to add [shellcheck](https://www.shellcheck.net/) linter for install scripts.

To avoid messing with https://github.com/getsentry/self-hosted/pull/3673 podman PR (and to not edit all scripts at once) this action only checks `install/_lib.sh`

I'm willing to slowly add all other `*.sh` scripts too.

Some trade-offs I made:
1. checks restricted to `install/_lib.sh` but action runs on any `**.sh` change.
1. [SC1090](https://www.shellcheck.net/wiki/SC1090) and [SC1091](https://www.shellcheck.net/wiki/SC1091) disabled because I don't have full context on how env-sources are used
1. `ensure_file_from_example()` excluded as kinda false-positive, pure bash solution won't be more readable I suppose
1. direct shellcheck run is used because tool is preinstalled and it's output could easily be converted to action annotations

Warnings from unmodified lib.sh are available in [this PR](https://github.com/doc-sheet/self-hosted/pull/2/files#diff-1b8181453221651d519090bfd43b23252b324c5fc3756c6f61de3c6aeed9a5c1)

### Legal Boilerplate

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,40 @@
+---
+name: "ShellCheck"
+on:
+  push:
+    paths:
+      - "**.sh"
+    branches: [master]
+  pull_request:
+    paths:
+      - "**.sh"
+    branches: [master]
+
+jobs:
+  shellcheck:
+    name: ShellCheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run ShellCheck
+        run: |
+          git diff --name-only -z `git merge-base origin/master HEAD` -- \
+            'install/_lib.sh' \
+          | xargs -0 -r -- \
+            shellcheck \
+              --shell=bash \
+              --exclude=SC1090,SC1091 \
+              --format=json1 \
+          | jq -r '
+            .comments
+            | map(.level |= if ([.] | inside(["info", "style"])) then "notice" else . end)
+            | .[] as $note
+            | "::\($note.level) file=\($note.file),line=\($note.line),endLine=\($note.endLine),col=\($note.column),endColumn=\($note.endColumn)::[SC\($note.code)] \($note.message)"
+          ' \
+          | grep . >&2 && exit 1
+
+          exit 0
--- a/install/_lib.sh
+++ b/install/_lib.sh
@@ -33,6 +33,7 @@ function ensure_file_from_example {
     echo "$target already exists, skipped creation."
   else
     # sed from https://stackoverflow.com/a/25123013/90297
+    # shellcheck disable=SC2001
     example="$(echo "$target" | sed 's/\.[^.]*$/.example&/')"
     if [[ ! -f "$example" ]]; then
       echo "Oops! Where did $example go? 🤨 We need it in order to create $target."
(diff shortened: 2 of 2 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Run ShellCheck
+        run: |
+          git diff --name-only -z `git merge-base origin/master HEAD` -- \
+            'install/_lib.sh' \
+          | xargs -0 -r -- \
+            shellcheck \
+              --shell=bash \
+              --exclude=SC1090,SC1091 \
+              --format=json1 \
+          | jq -r '
```
aminvakil, line 32: What is the necessity of piping output of shellcheck to `jq`?

aminvakil (changes requested):
Thank you for providing this PR.

Honestly I'm not a fan of `shellcheck` on big projects myself, running `shellcheck` locally on self-hosted shows many errors and warnings which many are not a big deal.

I'd suggest running `shellcheck` locally and fix any issue which you may see and provide a PR for them, but I do not see a benefit of adding `shellcheck` in case we want to ignore it everywhere.

.github/workflows/shellcheck.yml
doc-sheet: just to replace some keys like `info` and `column` with github action [command](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error).
`notice` and `col` in this case.

And to format shellcheck's json as one-line command like `::error file=file.sh,col=123::message`

doc-sheet:
> in case want to ignore it everywhere

My point is to fix warnings first (which I plan to do anyway iteratively) and then don't ignore it anymore. I just don't want to break other people PRs with minor changes of quotes and exports here and there.

I just hope code without warnings would help with adding new features and refactoring.

aldy505:
@doc-sheet Just to let you know: i'd love to get this merged since this will bring quality improvement. But I'm currently working on other things (as you can see on Discord -- it's a chaos lol). What's blocking this PR is the CI not passing, if you were to be able to get that right, we can merge immediately.

doc-sheet:
Hi. At the time of creating PR I checked and this test was failling before too. Looks like there were some changes in python sdk.
I'll look into it.

And tbh I don't know what discord server you mention :)

doc-sheet:
Tests pass after rebase.

Fixed by #3714

aldy505:
> And tbh I don't know what discord server you mention :)

https://discord.gg/sentry this should be it

doc-sheet:
Thank you, didn't know it exists.

aldy505:
@aminvakil wanna take another look?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  shellcheck:
+    name: ShellCheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
aldy505, line 21: Just a nitpick, do we really need to fetch all git commit history?

aldy505 (approved):

.github/workflows/shellcheck.yml
doc-sheet: By default actions/checkout fetches current branch only so to get master branch and list of changed files - yes.
But I didn't experiment with depth, maybe full history is too much.

I think it would be faster to just check all shell-scripts after fixing all linter's warnings instead of calling git diff, so it's just a temporary solution.

aldy505: Ah got it.

aminvakil:
@doc-sheet Could you please break something, and revert it, so we can see the failure in this PR?

doc-sheet:
Sure but I have an example in my fork https://github.com/doc-sheet/self-hosted/pull/2/files#diff-1b8181453221651d519090bfd43b23252b324c5fc3756c6f61de3c6aeed9a5c1

Is it ok?

aminvakil (approved):
Thanks for adding this, we can remove it later if we saw we are ignoring it all the times.

doc-sheet:
Hm. DId sdk update again?

doc-sheet:
~~Yeah... https://github.com/getsentry/sentry-python/releases/tag/2.30.0~~
Nevermind it is just an incompatibility with 3.0 https://github.com/getsentry/sentry-python/releases/tag/3.0.0a2

aldy505:
From Anton Pirker: 
> oh yes. thanks for pointer @untitaker  (we removed the client parameter in the 3.0.0 version. we will take care of this.
> quick fix is to use a Python SDK <3.0.0 for the time being.

aldy505:
there you go, thanks! @doc-sheet

PR #3738, opened 2025-06-06
feat: Add taskbroker + worker + scheduler

markstory:
Add a taskbroker, scheduler and worker pod. While these pods aren't doing much right now, in a subsequent release they will and the cron/worker containers will be removed/replaced by these new containers.

Refs getsentry/taskbroker#267

### TODO
- [ ] Get dockerhub repository created
- [ ] Validate image publishing to dockerhub.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -482,6 +482,26 @@ services:
         <<: *depends_on-healthy
       web:
         <<: *depends_on-healthy
+  taskbroker:
+    <<: *restart_policy
+    image: "$TASKBROKER_IMAGE"
+    environment:
+      TASKBROKER_KAFKA_CLUSTER: "kafka:9092"
+      TASKBROKER_KAFKA_DEADLETTER_CLUSTER: "kafka:9092"
+      TASKBROKER_DB_PATH: "/opt/sqlite/taskbroker-activations.sqlite"
+    ports:
+      - "$SENTRY_BIND:50051/tcp"
+    volumes:
+      - sentry-taskbroker:/opt/sqlite
+    depends_on:
+      kafka:
+        <<: *depends_on-healthy
+  taskscheduler:
+    <<: *sentry_defaults
+    command: run taskworker-scheduler
+  taskworker:
+    <<: *sentry_defaults
+    command: run taskworker --concurrency=4 --rpc-host=taskbroker:50051 --num-brokers=1
   vroom:
     <<: *restart_policy
     image: "$VROOM_IMAGE"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -536,6 +556,10 @@ volumes:
   # Not being external will still persist data across restarts.
   # It won't persist if someone does a docker compose down -v.
   sentry-vroom:
+  # This volume stores task data that is inflight
+  # It should persist across restarts. If this volume is
+  # deleted, up to ~2048 tasks will be lost.
```
aldy505, line 559: This begs a question. If the volume if deleted, then how should one recover 2048 lost tasks?

docker-compose.yml
```diff
@@ -482,6 +482,26 @@ services:
...
+      - sentry-taskbroker:/opt/sqlite
+    depends_on:
+      kafka:
+        <<: *depends_on-healthy
+  taskscheduler:
+    <<: *sentry_defaults
+    command: run taskworker-scheduler
+  taskworker:
+    <<: *sentry_defaults
+    command: run taskworker --concurrency=4 --rpc-host=taskbroker:50051 --num-brokers=1
```
aldy505, line 502: Where is this `concurrency=4` coming from?

manishrawat1992: https://docs.python.org/3/library/multiprocessing.html

aldy505: No I mean, why does it has to be exactly 4?

docker-compose.yml
markstory: You would have to reset offsets in Kafka.

docker-compose.yml
markstory: It doesn't have to be 4. In saas, we typically run either 16 or 32 concurrent workers. If left undefined the default is `1` which is will limit throughput of a self-hosted install as tasks are used extensively in ingest. 4 felt like a decent balance of providing more processing power without consuming a ton of memory.

docker-compose.yml
```diff
@@ -482,6 +482,26 @@ services:
...
         <<: *depends_on-healthy
+  taskbroker:
+    <<: *restart_policy
+    image: "$TASKBROKER_IMAGE"
+    environment:
+      TASKBROKER_KAFKA_CLUSTER: "kafka:9092"
+      TASKBROKER_KAFKA_DEADLETTER_CLUSTER: "kafka:9092"
+      TASKBROKER_DB_PATH: "/opt/sqlite/taskbroker-activations.sqlite"
+    ports:
+      - "$SENTRY_BIND:50051/tcp"
```
aldy505: This must not be exposed.

```suggestion
```

Issue #3744, opened 2025-06-13
Customize Sentry DSN URL


yildizozgur:
### Problem Statement

Hi,

We have a self hosted sentry. And we use an ingress proxy url for external calls. On Sentry instance, DSN url is configured based on (I suppose) SENTRY_URL_PREFIX. Is there any specific parameter to be able to use overwriting only DSN url? I need to set it with our external ingress proxy url, and don't want to change SENTRY_URL_PREFIX. Because SENTRY_URL_PREFIX is basically set as an internal url.

Thanks a lot.

aldy505:
Yes, set this on your `sentry.conf.py`:
```py
# URI Prefixes for generating DSN URLs
# (Defaults to URL_PREFIX by default)
SENTRY_ENDPOINT = "https://sentry.ingest.example.com"
```

Then restart your `web` container. 

If you have the time & chance, you may add this into the `sentry.conf.example.py` file on this repo, so more people can know about this!

yildizozgur:
Thank you very much @aldy505 . It seems worked. I am going to add to sentry.conf.example.py

PR #3747, opened 2025-06-13
feat: enable customization sentry DSN endpoint

yildizozgur:
It is possible to customize DSN url by the configuration in sentry.conf.py.
This configuration option is added in sentry.conf.example.py.
https://github.com/getsentry/self-hosted/issues/3744 

### Legal Boilerplate


sentry/sentry.conf.example.py
```diff
@@ -341,6 +341,14 @@ def get_internal_network():
 # https://django-csp.readthedocs.io/en/latest/configuration.html
 # CSP_SCRIPT_SRC += ["example.com"]
 
+############################
+# Sentry Endpoint Settings #
+############################
+
+# URI Prefixes for generating DSN URLs
+# (Defaults to URL_PREFIX by default)
+SENTRY_ENDPOINT = "https://sentry.ingest.example.com"
```
aldy505: The default should be commented. We let people uncomment this line if they need it.

yildizozgur: Thank you very much @aldy505 for the notice. I have fiexed.

aldy505 (approved):

sentry/sentry.conf.example.py
aldy505: Thank you for contributing! We'll wait for others to review this, and we'll merge after that.

sentry/sentry.conf.example.py
```diff
@@ -341,6 +341,14 @@ def get_internal_network():
 # https://django-csp.readthedocs.io/en/latest/configuration.html
 # CSP_SCRIPT_SRC += ["example.com"]
 
+############################
+# Sentry Endpoint Settings #
+############################
+
+# URI Prefixes for generating DSN URLs
+# (Defaults to URL_PREFIX by default)
```
aminvakil:
```suggestion
# (default is URL_PREFIX)
```

aminvakil (commented):
Let's avoid duplication :)

aldy505:
@hubertdeng123 @BYK anyone wanna take a look?

BYK (approved):
Oh interesting, thanks for bringing this up! May also want to add a reason why people may want to use this over `URL_PREFIX`?

aldy505:
> Oh interesting, thanks for bringing this up! May also want to add a reason why people may want to use this over `URL_PREFIX`?

@BYK like me, I have different domain for ingestion and web UI. That way, the web UI can only be accessed via private corporate network, and the ingestion can be accessed publicly.

BYK:
@aldy505 that is the exact kind of comment I wanted to see _in the code_ not on GitHub 😉

yildizozgur:
Thank you very much for your support and feedback. For the next time, I will do it more carefully.

aldy505:
> Thank you very much for your support and feedback. For the next time, I will do it more carefully.

@yildizozgur Rest easy, I'll create a follow up PR sometime next week

BYK:
@yildizozgur 
>Thank you very much for your support and feedback. For the next time, I will do it more carefully.

My comment was for @aldy505, you did great and thanks a lot for taking the time to submit a PR! 🤗

PR #3750, opened 2025-06-14
ci: run tests on arm64

aldy505:
Let's see if this works

### Legal Boilerplate

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,10 @@ defaults:
 jobs:
   unit-test:
     if: github.repository_owner == 'getsentry'
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, ubuntu-24.04-arm64]
     name: "unit tests"
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


aldy505:
Ah, symbolicator's still missing the nightly images published to DockerHub. Sigh..

aminvakil (commented):
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories

> Linux [Public preview] 	4 	16 GB 	14 GB 	arm64 	[ubuntu-24.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md), [ubuntu-22.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md)

Arm is still in public preview of GHA, so maybe we should take cautions about adding it, as failing arm actions will prevent us from merging x86_64 fixes as well.

Also is there a problem with ubuntu 22? Changing multiple things in one PR always alerts me :-)

aldy505:
> https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories
> 
> > Linux [Public preview] 	4 	16 GB 	14 GB 	arm64 	[ubuntu-24.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md), [ubuntu-22.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md)
> 
> Arm is still in public preview of GHA, so maybe we should take cautions about adding it, as failing arm actions will prevent us from merging x86_64 fixes as well.

That was the thoughts of Sentry employees the first time I implemented this on all repositories (sentry, snuba, etc). We can remove it if it became troublesome, and it's not a required precondition in order to merge.

> Also is there a problem with ubuntu 22? Changing multiple things in one PR always alerts me :-)

I don't remember that there are Ubuntu 22 ARM runners. So I went ahead for 24.04.

PR #3752, opened 2025-06-16
fix(tests): Bump timeout

hubertdeng123:
CI is failing and we added many moving parts to sentry, wondering if we need to bump the timeout here


aldy505:
@hubertdeng123 can you try bringing back taskbroker?

hubertdeng123:
@aldy505 sure, to make sure that vroom changes are the culprit

hubertdeng123:
yep, looks like failures are def from vroom

PR #3755, opened 2025-06-17
fix ownership

oioki:
After https://github.com/getsentry/vroom/pull/593, `vroom` container runs as non-privileged user `vroom` with static uid=1000. This broke some integration tests as vroom uses profiles storage as local files in `/var/lib/sentry-profiles` in self-hosted installations.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -514,8 +514,6 @@ services:
   vroom:
     <<: *restart_policy
     image: "$VROOM_IMAGE"
-    depends_on:
-      - vroom-init
     environment:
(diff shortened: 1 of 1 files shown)


BYK (changes requested):
Okay this definitely verifies @hubertdeng123's assesment so thanks so much for that!

That said we are definitely not going with this `<service_name>-init` pattern here as services should handle these in their entry scripts. Or if this is a one time migration (doesn't look like it as busting caches did not help with the failures) we put these in the install script.

aldy505:
Closing as this is no longer the case

PR #3756, opened 2025-06-17
fix: Try fixing the build

BYK:
Use different Sentry image versions to see if it fixes the build


BYK:
Okay it _was_ vroom as @hubertdeng123 already figured out.

PR #3759, opened 2025-06-17
fix(vroom): Explicitly set PROFILES_DIR for upcoming change

BYK:
PROFILES_DIR was defaulting to `/var/lib/sentry-profiles` which requires root access. When Vroom image decided to go with non-root default user, this started causing permission issues. Now the image is being refactored and it will not use `/var/lib/sentry-profiles` as the default path so we need to explicitly pass it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -511,7 +511,7 @@ services:
     environment:
       SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
       SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
-      SENTRY_BUCKET_PROFILES: file://localhost//var/lib/sentry-profiles
+      PROFILES_DIR: "/var/lib/sentry-profiles"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -529,6 +529,10 @@ services:
         BASE_IMAGE: "$VROOM_IMAGE"
     entrypoint: "/entrypoint.sh"
     environment:
+      SENTRY_KAFKA_BROKERS_PROFILING: "kafka:9092"
+      SENTRY_KAFKA_BROKERS_OCCURRENCES: "kafka:9092"
+      PROFILES_DIR: "/var/lib/sentry-profiles"
+      SENTRY_SNUBA_HOST: "http://snuba-api:1218"
```
aldy505: I don't think we need all these. But oh well.

BYK: Why not? I thought these are required to get vroom running properly?

aldy505: it's not for the cleanup container since we don't connect to snuba and kafka

PR #3760, opened 2025-06-17
Revert "fix(vroom): Explicitly set PROFILES_DIR for upcoming change"

hubertdeng123:
Reverts getsentry/self-hosted#3759


install/ensure-correct-permissions-profiles-dir.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo "${_group}Ensuring correct permissions on profiles directory ..."
+$dc run --rm vroom chown -R 1000:1000 /var/lib/sentry-profiles
```
BYK: Pretty sure you need the following
```suggestion
$dc run --rm --entrypoint /bin/bash --user root vroom chown -R 1000:1000 /var/lib/sentry-profiles
```

hubertdeng123:
Permissions look to be correct, but still failing CI? :/

PR #3780, opened 2025-06-25
chore: provide detailed note for sentry endpoint settings

aldy505:
Follow up for https://github.com/getsentry/self-hosted/pull/3747


BYK (approved):
❤️

PR #3781, opened 2025-06-26
fix: Increase timeout for flakey test

tobias-wilfert:
This test has been incredibly flakey as of late which is impacting the Relay CI, as we run these test there as well.
This now tries to mitigate the flakeyness by increasing the timeout.

If you think there is a better way to approach this please do not hesitate to comment, same for if you think this is not the correct way to tackle the flakeyness.

Note: This test also seems to be flakey in self-hosted, this can be seen by search for the test name in github, since codecov will post a comment with the name of the failed test: https://github.com/search?q=org%3Agetsentry+test_01_basics.py&type=pullrequests

### DIFF
--- a/_integration-test/test_01_basics.py
+++ b/_integration-test/test_01_basics.py
@@ -22,7 +22,7 @@ SENTRY_CONFIG_PY = "sentry/sentry.conf.py"
 SENTRY_TEST_HOST = os.getenv("SENTRY_TEST_HOST", "http://localhost:9000")
 TEST_USER = "test@example.com"
 TEST_PASS = "test123TEST"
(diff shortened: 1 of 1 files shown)


aldy505:
@BYK I'm deferring this to you

BYK:
@tobias-wilfert this test was not flaky. If it became flaky recently (sounds like it has) then the solution is not to increase the timeout but figure out why the ingestion got slower. Can we get someone on this please?

tobias-wilfert:
I can look further into it but the fact that a "slower ingestion" only seems to manifest in these self-hosted tests and not in any of our Relay tests already seems fishy to me.

Issue #1226, opened 2021-12-27
Make system.secret-key configurable via environment


glensc:
### Problem Statement

I would like to keep `sentry/config.yml` in a git repository and share it with two instances: https://sentry.test.onpremise.example.com and https://sentry.production.onpremise.example.com. committing value of `system.secret-key` is not appropriate there.

### Solution Brainstorm

There's `sentry/sentry.conf.py` python code, which imports a bunch of stuff from the environment, why can't that key also be there?

An alternative would be to place the secret key to a file, for example: `sentry/secret-key`

aldy505:
@BYK @hubertdeng123 do you think this is something that's good to work on? For me, this is easy enough though. Will be a similar work with @williamdes's PR #3664

BYK:
I'm okay with env

PR #3783, opened 2025-06-29
feat: make `system.secret-key` configurable from environment variables

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/1226

Generate `SENTRY_SYSTEM_SECRET_KEY` with a secret value on `.env.custom`.
This also introduces the usage of `.env.custom` across self-hosted deployment. Might break some people's automation though, but hopefully it won/t if they don't change their `sentry/sentry.conf.py` file.

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -54,7 +54,7 @@ system.internal-url-prefix: 'http://web:9000'
 # A new key can be generated with `$ sentry config generate-secret-key`
 #
 # If you are using SENTRY_SYSTEM_SECRET_KEY that is being set on your `.env` or `.env.custom` file,
-# you should comment this line below.
+# you should remove this line below as it won't be used anyway.
 system.secret-key: '!!changeme!!'
 
 # The ``redis.clusters`` setting is used, unsurprisingly, to configure Redis


aldy505:
I made it much more simpler now.

sentry/config.example.yml
```diff
@@ -52,6 +52,9 @@ system.internal-url-prefix: 'http://web:9000'
 # If this file ever becomes compromised, it's important to generate a new key.
 # Changing this value will result in all current sessions being invalidated.
 # A new key can be generated with `$ sentry config generate-secret-key`
+#
+# If you are using SENTRY_SYSTEM_SECRET_KEY that is being set on your `.env` file,
```
aminvakil:
```suggestion
# If you are using SENTRY_SYSTEM_SECRET_KEY that is being set on your `.env` or `.env.custom` file,
```

aldy505: :laughing:

sentry/sentry.conf.example.py
```diff
@@ -64,10 +64,25 @@ def get_internal_network():
 # and thus various UI optimizations should be enabled.
 SENTRY_SINGLE_ORGANIZATION = True
 
+# Sentry event retention days specifies how long events are retained in the database.
+# This should be set on your `.env` or `.env.custom` file, instead of modifying
+# the value here.
+# NOTE: The longer the days, the more disk space is required.
```
BYK, line 70:
```suggestion
# NOTE: The longer the days, the more disk space is required and 90 days is the maximum value you can use.
```

sentry/config.example.yml
```diff
@@ -52,6 +52,9 @@ system.internal-url-prefix: 'http://web:9000'
 # If this file ever becomes compromised, it's important to generate a new key.
 # Changing this value will result in all current sessions being invalidated.
 # A new key can be generated with `$ sentry config generate-secret-key`
+#
+# If you are using SENTRY_SYSTEM_SECRET_KEY that is being set on your `.env` or `.env.custom` file,
+# you should comment this line below.
```
BYK:
```suggestion
# you should remove this line below as it won't be used anyway.
```

sentry/sentry.conf.example.py
aldy505: > and 90 days is the maximum value you can use.

Actually, no. You can go for 365 days. See https://github.com/getsentry/sentry/issues/94602

aminvakil (approved):

sentry/sentry.conf.example.py
BYK: Well, I was told many places in the system assume a maximum of 90 days that's why I suggested this. And the issue you linked kind of shows that? :D

aldy505: But still :D

BYK: 🤷🏻‍♂️

Issue #3304, opened 2024-08-31
Add Uptime Monitoring


aldy505:
### Problem Statement

Uptime Monitoring feature is out and it's on Early Release. We should bring that here if it hits General Availability.

Product discussion: https://github.com/getsentry/sentry/discussions/51110
Announcement: https://sentry.io/changelog/uptime-monitoring-early-adopters-release/
Documentation: https://docs.sentry.io/product/alerts/uptime-monitoring/

### Solution Brainstorm

The way we'll tell you how to configure this should be versioned (learned this the hard way from User Feedback feature). Grab the version you're currently on (or if yours is higher than any of these, choose the highest one).

- Setup guide for 24.8.0

aldy505:
## Setup guide for 24.8.0

Date of writing: 2024-08-31

Does this works for versions below 24.8.0? I have no idea.

On your `docker-compose.yml`, add a few new containers:
```yaml
  uptime-results:
    <<: *sentry_defaults
    command: run consumer uptime-results --consumer-group uptime-results
    profiles:
      - feature-complete
  uptime-checker:
    <<: *restart_policy
    # IMPORTANT, READ THIS!
    # If you trust me enough (although we probably never met), you can use the Docker image I built to save some time, so you can use this:
    # image: "aldy505/sentry-uptime-checker:8ffa29b89b319e4b8e51154e82569796ea74c3af"
    #
    # But if you don't trust me enough and prefer to build from source yourself, use this instead (I'm using this on my instance).
    # If you're using the `image` section above, obviously you'll need to remove this `build` section (or just comment it out).
    build: ./uptime-checker-git
    command: run --config /config.yml
    volumes:
      - type: bind
        read_only: true
        source: ./uptime-checker/config.yml
        target: /config.yml
    depends_on:
      kafka:
        <<: *depends_on-healthy
      redis:
        <<: *depends_on-healthy
    profiles:
      - feature-complete
```

On your `sentry/sentry.conf.py`, add some new feature flags:
```python
SENTRY_FEATURES.update(
    {
        feature: True
        for feature in (
          # The rest of your features
        )
        # Uptime Monitoring related flags
        + (
            "organizations:uptime-api-create-update",
            "organizations:uptime-automatic-hostname-detection",
            "organizations:uptime-automatic-subscription-creation",
            "organizations:uptime-rule-api",
            "organizations:uptime-create-issues",
            "organizations:uptime-settings",
            "organizations:uptime-visible", # This is for issues platform
[...]

PR #3787, opened 2025-07-02
feat(uptime): Enable uptime in self-hosted

evanpurkhiser:
This brings uptime features self hosted sentry installations

This was previously blocked on docker images for the uptime-checker being available.

There is a feature flag and option that need to be enabled for now. But we'll likely remove these in the future.

Fixes https://github.com/getsentry/self-hosted/issues/3304

A few notes

- We may want to expose an official way to disable internal-ip filtering in the uptime-checker. Since self-hosted users may wish to monitor internal services, and it's unclear that we filter those out.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -305,6 +305,11 @@ services:
     command: rust-consumer --storage spans --consumer-group snuba-spans-consumers --auto-offset-reset=latest --max-batch-time-ms 1000 --no-strict-offset-reset
     profiles:
       - feature-complete
+  snuba-uptime-results-consumer:
+    <<: *snuba_defaults
+    command: rust-consumer --storage uptime_monitor_checks --consumer-group snuba-uptime-results --auto-offset-reset=latest --max-batch-time-ms 750 --no-strict-offset-reset
+    profiles:
+      - feature-complete
   symbolicator:
     <<: *restart_policy
     image: "$SYMBOLICATOR_IMAGE"
@@ -413,6 +418,11 @@ services:
   monitors-clock-tasks:
     <<: *sentry_defaults
     command: run consumer monitors-clock-tasks --consumer-group monitors-clock-tasks
+    profiles:
+      - feature-complete
+  uptime-results:
+    <<: *sentry_defaults
+    command: run consumer uptime-results --consumer-group uptime-results
     profiles:
       - feature-complete
   post-process-forwarder-transactions:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -533,6 +543,20 @@ services:
       - sentry-vroom:/var/vroom/sentry-profiles
     profiles:
       - feature-complete
+  uptime-checker:
+    <<: *restart_policy
+    image: "$UPTIME_CHECKER_IMAGE"
+    command: run
+    environment:
+      UPTIME_CHECKER_RESULTS_KAFKA_CLUSTER: kafka:9092
+      UPTIME_CHECKER_REDIS_HOST: redis://redis:6379
```
aldy505, line 552: Should we also expose `UPTIME_CHECKER_ALLOW_INTERNAL_IPS` and `UPTIME_CHECKER_HTTP_CHECKER_DNS_NAMESERVERS` here? That being commented out is good enough for future references.

sentry/sentry.conf.example.py
```diff
@@ -314,9 +314,23 @@ def get_internal_network():
...
+        # Uptime related flags
+        + (
+            "organizations:uptime-create-issues",
+            # TODO(epurkhiser): We can remove remove these in 25.8.0 since
+            # we'll have released this issue group type
+            # (https://github.com/getsentry/sentry/pull/94827)
+            "organizations:issue-uptime-domain-failure-visible",
+            "organizations:issue-uptime-domain-failure-ingest",
+            "organizations:issue-uptime-domain-failure-post-process-group",
+        )
```
aldy505: Shouldn't we need `organizations:uptime` here?

https://github.com/getsentry/sentry/blob/cce6829b019961bfb233ed27eb4f303ce0c8fd72/src/sentry/features/permanent.py#L109-L110

I know that :point_up: defaults to `True`, yet we have `errors-only` mode in which shouldn't expose the uptime component.

docker-compose.yml
aldy505:
```suggestion
    environment:
      UPTIME_CHECKER_RESULTS_KAFKA_CLUSTER: kafka:9092
      UPTIME_CHECKER_REDIS_HOST: redis://redis:6379
      # Set to `true` will allow uptime checks against private IP addresses
      UPTIME_CHECKER_ALLOW_INTERNAL_IPS: false
      # DNS name servers to use when making checks in the http checker. Separated by commas.
      #UPTIME_CHECKER_HTTP_CHECKER_DNS_NAMESERVERS: "8.8.8.8,8.8.4.4"
      # The number of times to retry failed checks before reporting them as failed
      UPTIME_CHECKER_FAILURE_RETRIES: "0"
```

evanpurkhiser: Yeah that sounds good

Issue #1573, opened 2022-07-08
uWSGI reports full listen queue


emmatyping:
### Version

22.7.0.dev0

### Steps to Reproduce

1. Run `./install.sh`
2. `docker compose up -d`
3. `docker compose logs web`

### Expected Result

No uWSGI errors.

### Actual Result

The web container logs contain

```
uWSGI listen queue of socket "127.0.0.1:46721" (fd: 3) full
```

It seems that uWSGI is reporting that the socket connection queue is full. This is weird for 2 reasons:

 - This is a brand new instance of sentry self-hosted
 - The limit is 100 connections(!)

Edit: there is some more info on the listen queue in the uWSGI docs https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html#the-listen-queue

aminvakil:
We receive 300 requests per second on average on our sentry instance and do not have a problem, probably because we're using a nginx installed on host to reverse proxy to sentry instance, so we're not hitting sentry with 300 requests directly.

I guess it's reasonable for any uWSGI instance to use a reverse proxy in front of it instead of delivering requests to it directly.

Do you think it suffices if we document this?

aminvakil:
@ValentinFrancois Sorry for late response.

Based your configuration `nginx` can keep a limited set of open connections to backend you've configured, and uses those connections to pass the traffic.

Therefore there is no need to open a new connection to backend for each new connection.

ValentinFrancois:
@aminvakil you're mentioning [this mechanism](https://www.nginx.com/blog/http-keepalives-and-web-performance/), right? I think I understand better now
![image](https://user-images.githubusercontent.com/24223132/196034849-55285e5a-6a07-47be-9063-63ee9697adfe.png)

aminvakil:
@ValentinFrancois Exactly!

Nekuromento:
Encountering this error on M1 mac host (I think this is caused by https://github.com/unbit/uwsgi/issues/2406)
Anyone managed to find a workaround?

yanghua-ola:
> We receive 300 requests per second on average on our sentry instance and do not have a problem, probably because we're using a nginx installed on host to reverse proxy to sentry instance, so we're not hitting sentry with 300 requests directly.
> 
> I guess it's reasonable for any uWSGI instance to use a reverse proxy in front of it instead of delivering requests to it directly.
> 
> Do you think it suffices if we document this?

[...]

PR #3792, opened 2025-07-08
fix: set harakiri Django option to 30s

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/1573

From @guoard and @aminvakil who found the issue of "uWSGI reports full listen queue" was caused mostly by workers taking longer times to finish. For folks with bigger Sentry installation, they might want to increase the `proxy_read_timeout` and `harakiri` values to a longer (acceptable) time.

See the GitHub issue linked above for more details.

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -235,7 +235,7 @@ SENTRY_WEB_OPTIONS = {
     # caused by GIL issues or deadlocks.
     # Ensure nginx `proxy_read_timeout` configuration (default: 30)
     # on your `nginx.conf` file to be at least 5 seconds longer than this.
-    # "harakiri": 30,
+    # "harakiri": 25,
     # Some stuff so uwsgi will cycle workers sensibly
     "max-requests": 100000,
     "max-requests-delta": 500,


sentry/sentry.conf.example.py
```diff
@@ -230,6 +230,9 @@ def get_internal_network():
     "workers": 3,
     "threads": 4,
     "memory-report": False,
+    # Kills requests that take longer than 30s
+    # which can help avoid stuck workers caused by GIL issues or deadlocks
+    "harakiri": 30,
```
BYK: I recommend documenting this first instead of making it the default?

nginx.conf
```diff
@@ -64,7 +64,7 @@ http {
 	proxy_set_header X-Forwarded-For $remote_addr;
 	proxy_set_header X-Forwarded-Proto $scheme;
 	proxy_set_header X-Request-Id $request_id;
-	proxy_read_timeout 30s;
+	proxy_read_timeout 35s;
```
BYK: Should this value be in sycn with the `harakiri` value below? If so we should at least document this if not find a way to link them through an env variable if possible.

BYK (changes requested):

nginx.conf
guoard: Values are based on [zulip/zulip#31909](https://github.com/zulip/zulip/pull/31909).

BYK: Ah, so they _are_ related and we should document the relationship if we cannot automate it somehow.

sentry/sentry.conf.example.py
```diff
@@ -230,6 +230,12 @@ def get_internal_network():
     "workers": 3,
     "threads": 4,
     "memory-report": False,
+    # The `harakiri` option terminates requests that take longer than the
+    # defined amount of time (in seconds) which can help avoid stuck workers
+    # caused by GIL issues or deadlocks.
+    # Modifying this value will also require you to modify `proxy_read_timeout`
+    # on your `nginx.conf` file to be at least 5 seconds longer than this.
+    # "harakiri": 30,
```
aminvakil:
```suggestion
    # "harakiri": 25,
```

sentry/sentry.conf.example.py
```diff
@@ -230,6 +230,12 @@ def get_internal_network():
     "workers": 3,
     "threads": 4,
     "memory-report": False,
+    # The `harakiri` option terminates requests that take longer than the
+    # defined amount of time (in seconds) which can help avoid stuck workers
+    # caused by GIL issues or deadlocks.
+    # Modifying this value will also require you to modify `proxy_read_timeout`
```
aminvakil:
```suggestion
    # Ensure nginx `proxy_read_timeout` configuration (default: 30)
```

BYK:
@aldy505 would you mind adding something about this under the troubleshooting section in self-hosted docs?

PR #3793, opened 2025-07-08
test: run patch tests for optional modifications

aldy505:
Right now we don't know if it's broken, this tests will cover that issue

### DIFF
--- a/.github/workflows/optional-modifications.yml
+++ b/.github/workflows/optional-modifications.yml
@@ -1,0 +1,69 @@
+name: Optional Modifications
+
+on:
+  # NOTE(reinaldy): This `pull_request` trigger should be removed once we have it tested on the PR.
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 */7 * *"
+
+concurrency:
+  group: ${{ github.ref_name || github.sha }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  test-apply-patch:
+    name: Test Apply Patch
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Apply patches
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+
+          # For each directory in `optional-modifications/patches/`, we will try to execute:
+          # `patch < optional-modifications/patches/<dir>/<file>.patch`
+          # with `<file>` being the name of the file in the directory that we loop through.
+          #
+          # If the patch fails, we will open a GitHub issue with the failing patch file.
+          # To prevent so many issues created, we will check first whether there is already
+          # an issue open with the same patch `<dir>` (without the file name).
+
+          for patch_dir in optional-modifications/patches/*; do
+            echo "::group::Checking for existing issue for $patch_dir"
+
+            issue_title="[Optional Modification Failure] $patch_dir"
+            issue_body=("Failure during applying patches for $patch_dir:")
+
+            for patch_file in "$patch_dir"/*.patch; do
+              echo "::debug::Checking for existing issue for $patch_file"
+              patch_base_file_name=$(basename "$patch_file" .patch)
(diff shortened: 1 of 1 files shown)


aldy505:
> What do you think about running this workflow on pull requests? This does not run on pull requests as it is now, but we do not want to open an issue on a broken patch in a PR.

I don't want to make the person who made the patch broken to be responsible to fixing the patch when they might not have understand how the patch should work, therefore I don't want to make the PR's CI red and blocking the merge process.

> What is GITHUB_TOKEN secret value in PRs? If it just fails it's fine IMO and we can add `on pull_requests`.

The same with `${{ github.token }}`. It can open issues.

aminvakil:
So maintainer should test patches locally when someone opens a PR?

aldy505:
> So maintainer should test patches locally when someone opens a PR?

Not really, no. I would prefer that the author (or maintainer) of the patch be notified when their patch can't be applied.

aminvakil:
I meant new PRs.

Let's say user1 opens a PR which wants to add a patch, how should someone review it?

aldy505:
Ah! You got a point. Let me fix that.

aldy505:
There :laughing: @aminvakil

aminvakil:
I think I'm multitasking badly right now and get everything mixed :)

Sorry, I thought pull request was not activated on this workflow.

.github/workflows/optional-modifications.yml
```diff
@@ -0,0 +1,75 @@
...
+          # an issue open with the same patch `<dir>` (without the file name).
+
+          for patch_dir in optional-modifications/patches/*; do
+            echo "::group::Checking for existing issue for $patch_dir"
+
+            issue_title="[Optional Modification Failure] $patch_dir"
+            issue_body=("Failure during applying patches for $patch_dir:")
+
+            for patch_file in "$patch_dir"/*.patch; do
+              echo "Checking for existing issue for $patch_file"
```
BYK, line 52: Misleading message?

.github/workflows/optional-modifications.yml
```diff
@@ -0,0 +1,75 @@
...
+
+            # If the length of `issue_body` array is greater than 1,
+            # then we have a problem and we need to open an issue.
+            if [ "${#issue_body[@]}" -gt 1 ]; then
+              issue_exists=$(gh issue list --search "$issue_title in:title" --state open --json title --jq 'if length == 0 then empty end')
+              if [ -z "$issue_exists" ]; then
+                echo "Creating issue for $patch_dir"
+                gh issue create --title "$issue_title" --body "${issue_body[*]}"
+              fi
+            fi
```
BYK, line 69: I'd say check for the issue first and if there's an existing issue skip the apply/test step too.

.github/workflows/optional-modifications.yml
```diff
@@ -0,0 +1,75 @@
...
+
+            echo "::debug::Reset the repository state"
+            git reset --hard
```
BYK, line 72: These seem like cheap tests, only testing if the patch would apply cleanly. I think we can run these on PR for information and on master to create a new issue.

The "heavy" tests would be the ones which apply the patches _and_ test if they actually work. This might be tricky.

At this point I'm wondering if we should spin off these patches into their own repos (for externals too) and find a way to apply these at install step easily. This would also offload the maintenance burden to whoever created the patches.

aldy505: > These seem like cheap tests, only testing if the patch would apply cleanly. I think we can run these on PR for information and on master to create a new issue.
> 
> The "heavy" tests would be the ones which apply the patches _and_ test if they actually work. This might be tricky.

It's not just tricky. If it's something like this external Kafka with auth + TLS support, you'd need to spawn an actual external Kafka with those configurations. Yet, I do think it is too heavy to do, and this minimal tests should work fine.

BYK: >If it's something like this external Kafka with auth + TLS support, you'd need to spawn an actual external Kafka with those configurations.

That's also why I want these in their own repos so those repo owners would be thinking about these, not us 😅

aldy505:
Going nowhere

PR #3794, opened 2025-07-08
docs: encourage community patches

aldy505:
Hopefully this will provide a better guide on how to create patches.


optional-modifications/README.md
```diff
@@ -14,28 +13,38 @@ the desired goal.
...
 ## How to use patches
 
-The patches are designed mostly to help modify the existing
-configuration files. You will need to run the `install.sh` script
-afterwards.
+The patches are designed mostly to help modify the existing configuration files. You will need to run the `install.sh` script afterwards.
 
-They should be run from the root directory. For example, the
-`external-kafka` patches should be run as:
+They should be run from the root directory. For example, the `external-kafka` patches should be run as:
```
aminvakil, line 18: Can we ensure patches work fine without conflicts as we change `docker-compose.yml` over time?

Maybe a GHA or something like that?

I'll whip up a PR if you're OK with it.

aminvakil: Oops! I just saw https://github.com/getsentry/self-hosted/pull/3793 :)

aldy505: Done that here https://github.com/getsentry/self-hosted/pull/3793

aldy505: Oh, anyway, I was thinking of mentioning that there will be a CI job that runs weekly, if a patch is broken, it will create a new github issue, and I'll alert the original patch author.

BYK (approved):
The patch looks good. Two notes:

1. We may want to invest in an actual plug-in system as patch files won't be easy to maintain
2. These docs are getting longer so might be deserving their own spot on self-hosted docs now?

aldy505:
> 1. We may want to invest in an actual plug-in system as patch files won't be easy to maintain

I'd prefer to go with this current route first, then if we have enough interest, we may move into a plugin system. Currently I have no idea on how it looks like if you need to modify the docker compose and everything.

> 2. These docs are getting longer so might be deserving their own spot on self-hosted docs now?

Honestly it's harder for me to get things merged on the docs rather than on self-hosted.

BYK:
>Honestly it's harder for me to get things merged on the docs rather than on self-hosted.

Not if you ping me for this.

PR #3817, opened 2025-07-18
feat: inspect docker compose failure on self-hosted e2e action

aldy505:
It's hard to debug docker compose failure on other repositories since they can't see the `docker compose ps` and `docker compose logs`. One problem occurred here: https://github.com/getsentry/relay/pull/4940

This PR aims to provide both commands if failure happens.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -171,6 +171,7 @@ runs:
 
     - name: Inspect failure
       if: failure()
+      shell: bash
       run: |
         echo "::group::Inspect failure - docker compose ps"
         docker compose ps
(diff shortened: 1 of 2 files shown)


action.yaml
```diff
@@ -168,3 +168,13 @@ runs:
       with:
         directory: ${{ github.action_path }}
         token: ${{ inputs.CODECOV_TOKEN }}
+
+    - name: Inspect failure
+      if: failure()
+      run: |
```
aldy505:
```suggestion
    - name: Inspect failure
      if: failure()
      shell: bash
      run: |
```

hubertdeng123 (approved):

BYK (approved):

PR #3829, opened 2025-07-23
docs: clearly state that `system.internal-url-prefix` shouldn't be changed

aldy505:
Happened today on Discord where a user changed internal-url-prefix and broke their instance. I think this should be stated clearer for new users and for non-native English speakers.

### DIFF
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -52,7 +52,7 @@ mail.host: 'smtp'
 
 # Most of the time, this should NOT be changed. It's used for communication
 # between containers. `web` is the container's name, and `9000` is the
-# default port exposed by the Sentry image. If you want to change the
+# default port opened by the Sentry backend (this is NOT the public port). If you want to change the
 # publicly exposed domain or port, you should probably change
 # `system.url-prefix` instead.
 system.internal-url-prefix: 'http://web:9000'
--- a/sentry/config.example.yml
+++ b/sentry/config.example.yml
@@ -52,9 +52,11 @@ mail.host: 'smtp'
 
 # Most of the time, this should NOT be changed. It's used for communication
 # between containers. `web` is the container's name, and `9000` is the
-# default port opened by the Sentry backend (this is NOT the public port). If you want to change the
-# publicly exposed domain or port, you should probably change
-# `system.url-prefix` instead.
+# default port opened by the Sentry backend (this is NOT the public port).
+#
+# If you want to change the publicly exposed domain or port, you should change
+# `system.url-prefix` above instead, along with `SENTRY_BIND` in `.env` file.
+# Also see https://develop.sentry.dev/self-hosted/#productionalizing.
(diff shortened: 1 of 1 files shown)


aldy505:
Ah, this explains why I had dejavu. I made the exact same PR a few years back. https://github.com/getsentry/self-hosted/pull/3100

sentry/config.example.yml
```diff
@@ -45,8 +45,16 @@ mail.host: 'smtp'
...
 
-# The URL prefix in which Sentry is accessible
+# This is the main URL prefix where Sentry can be accessed.
+# Sentry will use this to create links to its different parts of the user interface.
+# This is most helpful if you are using an external reverse proxy.
 # system.url-prefix: https://example.sentry.com
+
+# Most of the time, this should NOT be changed. It's used for communication
+# between containers. `web` is the container's name, and `9000` is the
+# default port exposed by the Sentry image. If you want to change the
```
BYK:
```suggestion
# default port opened by the Sentry backend (this is NOT the public port). If you want to change the
```

sentry/config.example.yml
```diff
@@ -45,8 +45,16 @@ mail.host: 'smtp'
...
+# publicly exposed domain or port, you should probably change
```
BYK:
```suggestion
# publicly exposed domain or port, you should change
# `system.url-prefix` above instead
# along with SENTRY_BIND in `.env` file.
# Also see https://develop.sentry.dev/self-hosted/#productionalizing
```

aldy505: Ughhh remind me a bit later

sentry/config.example.yml
```diff
@@ -45,8 +45,16 @@ mail.host: 'smtp'
 # System Settings #
 ###################
 
-# The URL prefix in which Sentry is accessible
+# This is the main URL prefix where Sentry can be accessed.
+# Sentry will use this to create links to its different parts of the user interface.
```
aminvakil:
```suggestion
# Sentry will use this to create links to its different parts of the self-hosted interface.
```

BYK: Don't think this makes it any clear. For me it is more confusing as I don't know what "self-hosted interface" is. What we mean here is the web UI, maybe we should just say that?

sentry/config.example.yml
BYK: Why? 😀

sentry/config.example.yml
aldy505:
```suggestion
# Sentry will use this to create links to its different parts of the web UI.
```

What about this?

aldy505: Oh missed burak's comment.

sentry/config.example.yml
aldy505: I'm juggling so many things at work :joy:

BYK (approved):
Unblocking

Issue #3526, opened 2025-01-13
--no-report-self-hosted-issues flag doesn't work in backup command


ingria:
### Self-Hosted Version

24.12.1

### CPU Architecture

x86_64

### Docker Version

27.4.1

### Docker Compose Version

v2.32.1

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

1. read the restore.sh script suggestion:
```
Thirty (30) day retention. No marketing. Privacy policy at sentry.io/privacy.

y or n? n

Understood. To avoid this prompt in the future, use one of these flags:

  --report-self-hosted-issues
  --no-report-self-hosted-issues
```

2. run ./scripts/restore.sh --no-report-self-hosted-issues

### Expected Result

everything works fine

### Actual Result

```
▶ Initializing Docker Compose ...

▶ Detecting Docker platform
Detected Docker platform is linux/amd64

▶ Setting up error handling ...

.....

Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
Usage: sentry import [OPTIONS] COMMAND [ARGS]...
Try 'sentry import --help' for help.

Error: No such option: --no-report-self-hosted-issues
cat: unrecognized option '--no-report-self-hosted-issues_log-2025-01-13_12-07-36.txt'
Try 'cat --help' for more information.
```

aldy505:
JFYI @hubertdeng123

klemen-df:
I think this is still not working

aldy505:
> I think this is still not working

@klemen-df Sorry, we don't have the resource & time to prioritize this yet. You may need to be patient.

aldy505:
I'll allocate time after 25.7.0 for this.

PR #3831, opened 2025-07-25
fix(scripts): every known flags should be shifted before executing the sentry <foo> command

aldy505:
Fixes https://github.com/getsentry/self-hosted/issues/3526

Previously, if this was executed:
```bash
./scripts/backup.sh --no-report-self-hosted-issues
```

The final backup command that will be executed is:
```bash
docker compose run -v "${PWD}/sentry:/sentry-data/backup" --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export --no-report-self-hosted-issues /sentry-data/backup/backup.json
```

Which is invalid. This PR strips all known flags specific to self-hosted before passing it onto Sentry.

One other option is to enable "ignore unknown options" on click (the CLI dependency) on Sentry, but I don't want to go to that route.

### DIFF
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -1,4 +1,17 @@
 #!/usr/bin/env bash
+MINIMIZE_DOWNTIME="${MINIMIZE_DOWNTIME:-}"
+REPORT_SELF_HOSTED_ISSUES="${REPORT_SELF_HOSTED_ISSUES:-}"
+
+while (($#)); do
+  case "$1" in
+  --report-self-hosted-issues) REPORT_SELF_HOSTED_ISSUES=1 ;;
+  --no-report-self-hosted-issues) REPORT_SELF_HOSTED_ISSUES=0 ;;
+  --minimize-downtime) MINIMIZE_DOWNTIME=1 ;;
+  *) version=$1 ;;
+  esac
+  shift
+done
+
 cmd="backup $1"
 source scripts/_lib.sh
 $cmd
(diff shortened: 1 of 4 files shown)


scripts/backup.sh
```diff
@@ -1,4 +1,17 @@
 #!/usr/bin/env bash
+MINIMIZE_DOWNTIME="${MINIMIZE_DOWNTIME:-}"
+REPORT_SELF_HOSTED_ISSUES="${REPORT_SELF_HOSTED_ISSUES:-}"
+
+while (($#)); do
```
hubertdeng123, line 5: Why not just add `--minimize-downtime` to the `scripts/_lib.sh`? That way we don't need to maintain 3 areas of similar logic

aldy505: Because it won't be shifted by Bash and would be included on the Docker command. The problem was this:
```bash
cmd="backup $1"
source scripts/_lib.sh
$cmd
```

If you do `./scripts/backup.sh --minimize-downtime`, the resulting `cmd` would be `backup --minimize-downtime`. If you want to add another known flag for the backup script, like `--silent`, you would have to do `backup --sillent --minimize-downtime`. It will be hard to make users aware that they are not supposed to do `backup --minimize-downtime --silent`.

So if you omit those known flags before passing the command, it will work as expected. This also would open the possibility of doing `cmd="backup $@"` (I don't know if this correct, but I hope you get the right idea) to allow more than one parameter.

PR #3845, opened 2025-08-01
feat: Continue using celery in self-hosted for now

markstory:
Disable taskworkers for self-hosted until the documentation has been published.  What I'm not clear on is how to define this option for _existing_ self-hosted installs.

Refs getsentry/sentry#96966


aldy505:
> What I'm not clear on is how to define this option for _existing_ self-hosted installs.

You can't. Users must update their config file themselves.

aldy505 (approved):

markstory:
> You can't. Users must update their config file themselves.

Good to know. The 25.8.0 release will still use celery then. For the 25.9 release I'll enable taskworkers by default, which will give us a couple releases that give folks the option to run either set of workers.

Issue #3838, opened 2025-07-29
Metric Alerts do not trigger


mzglinski:
### Self-Hosted Version

25.7.0

### CPU Architecture

x86_64

### Docker Version

28.3.2

### Docker Compose Version

2.38.2

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

1. Install fresh 25.7.0 release
2. Set up a project, start sending events, for example, using this script
```bash
~ cat ./events.sh
#!/bin/bash

DSN_PUBLIC="<FIRST PART OF DSN, BEFORE @>"
SENTRY_HOST="http://<SENTRY HOST>"
PROJECT_ID="<NUMBER AT THE END OF DSN>"

curl -X POST --data '{ "exception": [{ "type": "Sentry curl test request", "value": "This is a test request to sentry sent by curl" }] }' \
        -H "Content-Type: application/json" \
        -H "X-Sentry-Auth: Sentry sentry_version=7, sentry_key=$DSN_PUBLIC, sentry_client=raven-bash/0.1" \
        $SENTRY_HOST/api/$PROJECT_ID/store/

# Timestamps 2 s apart
START=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
END=$(date -u -d "+2 seconds" +"%Y-%m-%dT%H:%M:%SZ")

# IDs
EVENT_ID=$(cat /proc/sys/kernel/random/uuid | tr -d '-')
TRACE_ID=$(cat /proc/sys/kernel/random/uuid | tr -d '-')
SPAN_ID=$(openssl rand -hex 8)

# Build JSON payload
PAYLOAD=$(printf '{
  "event_id":"%s",
  "type":"transaction",
  "transaction":"curl-test-transaction",
  "start_timestamp":"%s",
  "timestamp":"%s",
  "contexts":{
    "trace":{
      "trace_id":"%s",
      "span_id":"%s",
      "op":"http",
      "status":"ok"
    }
  },
  "spans":[],
  "platform":"other"
}' "$EVENT_ID" "$START" "$END" "$TRACE_ID" "$SPAN_ID")

# Calculate byte length **without** trailing newline
LENGTH=$(printf %s "$PAYLOAD" | wc -c)

# Assemble envelope
ENVELOPE=$(printf '%s\n%s\n%s' \
[...]

PR #3847, opened 2025-08-04
fix: add schedulers for generic metrics subscriptions

mzglinski:
Fix for https://github.com/getsentry/self-hosted/issues/3838. By analyzing the setup in Snuba [devserver](https://github.com/getsentry/snuba/blob/master/snuba/cli/devserver.py), I've confirmed that in the current setup, self-hosted never acts on a generic metric subscription. I cannot be certain if this is a complete solution to the problem, but I do know it fixed the metric alerts for me. 

Proof:

image

### Legal Boilerplate


zKoz210:
It work for me, thanks!

aldy505 (approved):

aldy505:
Hi @mzglinski, thank you so much for taking your time to figure out that those 4 consumers are missing. I'm checking in internally with the Snuba team if this is the right way forward.

phacops (approved):

PR #3855, opened 2025-08-06
fix(enhancement): ensure correct ownership check before setting permissions of profiles

### LINKED ISSUE #3841: install.sh taking a really long time
### Self-Hosted Version

25.7.0

### CPU Architecture

x86_64

### Docker Version

28.1.1

### Docker Compose Version

2.35.1

### Machine Specification

### Steps to Reproduce

I have had my Sentry 25.6.0 running on docker for the past month or two and I upgraded it to 25.7.0 by downloading the repo and running the `install.sh`. The problem is that since Sentry has already been running for a long while, it created hundreds of millions of profiles and, when I run install.sh, it has a step that does a recursive chown for the vroom profiles, and it takes a really long time. (hours..)

### Expected Result

It would be nice if the install script skipped this process of chown vroom profiles if Sentry has already been running, which would already have the proper permissions set up.

### Actual Result

▶ Ensuring correct permissions on profiles directory ...

This runs forever if you have a ton of profiles.

In the permission script it executes:
```
$dcr --no-deps --entrypoint /bin/bash --user root vroom -c 'chown -R vroom:vroom /var/vroom/sentry-profiles && chmod -R o+rwx /var/vroom/sentry-profiles'
```

### Event ID

_No response_

LvckyAPI:
Fixes #3841 

Check if the parent directory of /var/vroom/sentry-profiles is already owned by vroom:vroom.
- If it is, do nothing.
- If owner is incorrect, correct it

### Legal Boilerplate

### DIFF
--- a/install/ensure-correct-permissions-profiles-dir.sh
+++ b/install/ensure-correct-permissions-profiles-dir.sh
@@ -4,6 +4,11 @@
 
 echo "${_group}Ensuring correct permissions on profiles directory ..."
 
-$dcr --no-deps --entrypoint /bin/bash --user root vroom -c 'chown -R vroom:vroom /var/vroom/sentry-profiles && chmod -R o+rwx /var/vroom/sentry-profiles'
+# Check if the parent directory of /var/vroom/sentry-profiles is already owned by vroom:vroom
+if [ "$(stat -c '%U:%G' /var/vroom)" = "vroom:vroom" ]; then
+  echo "Ownership of /var/vroom is already set to vroom:vroom. Skipping chown."
+else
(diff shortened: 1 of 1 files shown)


LvckyAPI:
> Nice & thanks!

You're welcome

aminvakil:
This is an improvement alright, but do we have any idea why it took so long at the first place?

LvckyAPI:
I guess there was something in the issue #3841? I don't know

aminvakil:
> I guess there was something in the issue #3841? I don't know

Yes, I've seen that, it was taking too long to exec chown which this PR is doing as well (in case it was parent directory was not owned by vroom:vroom)

Did this took too long because it was already owned by vroom:vroom ? This PR fixes this problem (if there is).

PR #3866, opened 2025-08-12
fix(eap): Fix dataset parameter to target spans

phacops:
We're going to need https://github.com/getsentry/relay/pull/5055 in order for this to work.


aldy505:
@phacops What about the dataset migration? Would the self-hosted users able to read their old spans/transactions data on the web UI?

phacops:
> @phacops What about the dataset migration? Would the self-hosted users able to read their old spans/transactions data on the web UI?

I'm working on it, it will come in a different PR. This is what needs to be done to have a proper release.

hubertdeng123 (approved):

PR #3868, opened 2025-08-13
Replace Redis with Valkey

aldy505:
> [!CAUTION]
> This is **me** playing around with Valkey as a Redis replacement, since so many people from the community mentioned Valkey already.
> There is no guarantee that this will be merged at all.


hubertdeng123:
It's nice to experiment with this, but could cause some pain for us to maintain self-hosted since prod uses redis

frederikspang:
While valkey is a great replacement - I think Redis 8 should be the natural next-step - We're running `redis:8.0.2-alpine` in prod; And we're seeing no problems there. We could try tuning for io-threads for throughput? 

I haven't measured how much throughput is actually in Redis, by default (vs kafka vs postgresql for instance).

But if Celery + rate limits + something-else is running there, there must be *some*.

I can test out 8.2 with increased io-threads over the next few weeks in our prod.

BYK:
I think Redis 8's license is okay to use in this repo? @gavin-zee? -- https://redis.io/legal/licenses/

aldy505:
> I think Redis 8's license is okay to use in this repo? @gavin-zee? -- https://redis.io/legal/licenses/

@BYK nevertheless, aren't we currently running Redis 6 on SaaS?

BYK:
@aldy505 good to be prepared. Also I didn't look _who_ wrote the `redis-8` message and now I realize it was @frederikspang who is not part of Sentry 😅

frederikspang:
> @aldy505 good to be prepared. Also I didn't look _who_ wrote the `redis-8` message and now I realize it was @frederikspang who is not part of Sentry 😅

Yeah - That’s my bad!
We’re running 8.2 on self-hosted!😂

LordSimal:
Would something like https://github.com/google/cadvisor help measure the throughput of the different containers?

docker-compose.yml
```diff
@@ -115,10 +115,10 @@ services:
       test: echo stats | nc 127.0.0.1 11211
   redis:
     <<: *restart_policy
-    image: "redis:6.2.14-alpine"
+    image: "docker.io/valkey/valkey:8.1.3-alpine"
```
williamdes, line 118: I had `image: eqalpha/keydb:x86_64_v6.3.4` running for months

Thanks for this PR !

aldy505:
Closing this PR, won't merge.

PR #3859, opened 2025-08-07
feat: healthchecks for sentry components

mzglinski:
Resolves  https://github.com/getsentry/self-hosted/issues/3853

Few services still are missing the healthcheck, here is why:

- relay - does not have any binaries in the distroless image
- taskbroker, taskworker, taskscheduler, uptime-checker - I was not able to find any healthcheck-related options; they probably do not support that at the moment
- snuba-replacer - same as the above, no support
- all types of cron and cleanups - I do not have an idea on how to test this, does not seem necessary. 

Proof:

```
admin@ip-172-31-23-18:~$ docker ps --format "{{.ID}}  {{.Status}}  {{.Names}}"
495e2eab6069  Up 13 minutes (healthy)  sentry-self-hosted-post-process-forwarder-errors-1
69d517b4b935  Up 13 minutes (healthy)  sentry-self-hosted-attachments-consumer-1
bd5b91f08f51  Up 13 minutes (healthy)  sentry-self-hosted-monitors-clock-tasks-1
2d61e4c79ad6  Up 13 minutes (healthy)  sentry-self-hosted-ingest-replay-recordings-1
b46278c643a0  Up 13 minutes (healthy)  sentry-self-hosted-transactions-consumer-1
4237ef96f4ee  Up 13 minutes (healthy)  sentry-self-hosted-ingest-monitors-1
5b2c5ead49f4  Up 13 minutes (healthy)  sentry-self-hosted-ingest-occurrences-1
6cc8c31d2cb4  Up 13 minutes (healthy)  sentry-self-hosted-subscription-consumer-generic-metrics-1
b33990f43127  Up 13 minutes (healthy)  sentry-self-hosted-subscription-consumer-metrics-1
c5411608bbfa  Up 13 minutes (healthy)  sentry-self-hosted-ingest-feedback-events-1
2bacdce7feea  Up 13 minutes (healthy)  sentry-self-hosted-process-segments-1
dad0562aed81  Up 13 minutes (healthy)  sentry-self-hosted-subscription-consumer-transactions-1
e465753d8531  Up 13 minutes (healthy)  sentry-self-hosted-billing-metrics-consumer-1
3c1609b9491b  Up 13 minutes (healthy)  sentry-self-hosted-worker-1
90e191aec337  Up 13 minutes (healthy)  sentry-self-hosted-post-process-forwarder-issue-platform-1
4bdca9ba81e5  Up 13 minutes  sentry-self-hosted-taskscheduler-1
[...]

PR #3874, opened 2025-08-14
fix: adjust file healthcheck durations

mzglinski:
### Legal Boilerplate


aminvakil:
What is the necessity for adding swap? AFAICS it already has 16GB memory which is above self-hosted limit (14GB)
https://github.com/getsentry/self-hosted/actions/runs/16963036969/job/48080155328?pr=3874#step:4:4255
```
               total        used        free      shared  buff/cache   available
Mem:            15Gi       1.6Gi       4.3Gi        48Mi        10Gi        13Gi
```

aminvakil:
Sorry, I've just seen https://github.com/getsentry/self-hosted/pull/3859#issuecomment-3187929781

I think there is another problem here if it wants to use more than 16GB.

mzglinski:
There are 72 containers, when they start (almost all) at once, the RAM requirements are much higher than 16GB, it settles later, but during startup, swap is required.

mzglinski:
Adding healthcheck just exposed that problem, before `--wait` finished without error, even if containers without healthcheck were killed or not started fully.

aminvakil:
> There are 72 containers, when they start (almost all) at once, the RAM requirements are much higher than 16GB, it settles later, but during startup, swap is required.

We should increase memory requirement if that's the case, there are many scenarios which swap cannot be added, my environments are one of them:)

mzglinski:
> We should increase memory requirement 

That is not a bad idea; however, we cannot do much about GitHub-hosted runners, having the amount of resources they have. There are [large runners](https://docs.github.com/en/actions/concepts/runners/larger-runners), but those are not free to use.

aminvakil:
> Adding healthcheck just exposed that problem, before `--wait` finished without error, even if containers without healthcheck were killed or not started fully.

Oops! Has there been any other complaint about this in issues or discord? @aldy505 



> > We should increase memory requirement
> 
> That is not a bad idea; however, we cannot do much about GitHub-hosted runners, having the amount of resources they have. There are [large runners](https://docs.github.com/en/actions/concepts/runners/larger-runners), but those are not free to use.

Yes, you're right. We have no other option aside from adding swap in GHA.

aldy505:
> > There are 72 containers, when they start (almost all) at once, the RAM requirements are much higher than 16GB, it settles later, but during startup, swap is required.
> 
> 
> 
> We should increase memory requirement if that's the case, there are many scenarios which swap cannot be added, my environments are one of them:)

No, please don't do that. I already recommend people to add some swapfile to battle with the insane amount of RAM usage. I'd prefer adding more info regarding swap.

aldy505:
> > Adding healthcheck just exposed that problem, before `--wait` finished without error, even if containers without healthcheck were killed or not started fully.
> 
> 
> 
> Oops! Has there been any other complaint about this in issues or discord? @aldy505 

I was the one who complained 😆. SH e2e tests are failing everywhere. And I don't want to merge the revert PR.

aldy505:
@aminvakil Burak is on vacation and I don't wanna wait until Hubert is awake. If you're ok with this, I'll go merge this.

aminvakil (approved):
I'm fine with adding swap on GHA.

aminvakil:
> > > Adding healthcheck just exposed that problem, before `--wait` finished without error, even if containers without healthcheck were killed or not started fully.
> > 
> > 
> > Oops! Has there been any other complaint about this in issues or discord? @aldy505
> 
> I was the one who complained 😆. SH e2e tests are failing everywhere. And I don't want to merge the revert PR.

I still cannot understand how integration tests passed before, it only needs more than 16GB upon starting up?

mzglinski:
`docker compose up --wait` stops observing a service without a defined healthcheck once it reaches the running state. After that, it may restart or be killed, and `up --wait` will succeed. Once `up --wait` finishes, nothing is really happening on the machine, so failed or killed services restart quickly.

However, with healthchecks defined, those services not only have to report a healthy state, but also stay in that state for the entire duration of the wait. It was still possible for the test to fail before, but I suspect the odds were quite small for that to happen.

I also did not check what exactly the integration tests are testing, it might be that they would pass even with some snuba or sentry consumers not running ;)

aminvakil:
> I also did not check what exactly the integration tests are testing, it might be that they would pass even with some snuba or sentry consumers not running ;)

I think most probably this was the case.

aldy505:
> I also did not check what exactly the integration tests are testing, it might be that they would pass even with some snuba or sentry consumers not running ;) 

Some of it test user creation, login, and some basic interaction with the API. Other than that, it sends events through the Python SDK: error, transaction, and profile. Then, query to the Sentry API to see if there are any submitted event from the SDK.

PR #3875, opened 2025-08-14
feat: Relay healthcheck

aldy505:
Continuation of #3859


docker-compose.yml
```diff
@@ -674,6 +674,9 @@ services:
         <<: *depends_on-healthy
       web:
         <<: *depends_on-healthy
+    healthcheck:
```
hubertdeng123, line 677: If this passes CI, seems fine to me

PR #3876, opened 2025-08-14
fix: setup swapfile only if runner architecture is X64 or X86

aldy505:
Using matrix.os only works on self-hosted. SH e2e test on other repositories would still fail.

Refer to https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -144,7 +144,7 @@ runs:
 
     - name: Setup swapfile
       shell: bash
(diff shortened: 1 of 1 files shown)


aldy505:
> Out of curiosity, does sentry gets tested on x86?

No, but just in case.

I'm okay with removing it.

PR #3878, opened 2025-08-16
chore(deps): bump clickhouse to 25.3

aldy505:
See https://github.com/getsentry/snuba/pull/7361:

> We are already running `25.3.6.10034` in `s4s` and some `de` clusters

ClickHouse upgrade PR on snuba: https://github.com/getsentry/snuba/pull/7339

### DIFF
--- a/clickhouse/default-password.xml
+++ b/clickhouse/default-password.xml
@@ -1,0 +1,10 @@
+<clickhouse>
+  <users>
+    <default>
+      <password></password>
+      <networks>
+        <ip>::/0</ip>
+      </networks>
+    </default>
+  </users>
+</clickhouse>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -187,7 +187,7 @@ services:
     build:
       context: ./clickhouse
       args:
-        BASE_IMAGE: "altinity/clickhouse-server:23.8.11.29.altinitystable"
+        BASE_IMAGE: "altinity/clickhouse-server:25.3.6.10034.altinitystable"
     ulimits:
       nofile:
         soft: 262144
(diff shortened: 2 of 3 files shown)


aldy505:
So this works. Next thing is to spin up a relatively "enough" VPS with an old version, and try to upgrade the clickhouse.

aldy505:
image

image


Can confirm this works.

aldy505:
> I repeat my plea to make these hard-stops machine-readable somewhere as I think it would help with some other things too.

I think it's going to be useful for self-hosted hard stops rather than ClickHouse hard stop, as we only use 2 image tags. 

Initially I thought of something like this (pseudocode, bash is a horrible language):

```
var current_ch_version = (docker compose exec clickhouse clickhouse-client 'select version()')
var ch_stops = ["23.11", "24.8", "24.11", "25.3"]
for (var ch_stop in ch_stops) do
  docker compose build clickhouse $ch_stop
  wait until healthy
  sleep 5s
end
```

BYK:
>I think it's going to be useful for self-hosted hard stops rather than ClickHouse hard stop, as we only use 2 image tags.

Yeah that's what I meant sorry. That code snippet you shared also looks fine to me.

aldy505:
> >I think it's going to be useful for self-hosted hard stops rather than ClickHouse hard stop, as we only use 2 image tags.
> 
> 
> 
> Yeah that's what I meant sorry. That code snippet you shared also looks fine to me.

I don't really like bash's `for` loop. Lets revisit that a bit later

aldy505:
@aminvakil Hey, can I ask your review on this one, making sure my bash is not wrong.

PR #3879, opened 2025-08-16
chore(deps): bump patches version

aldy505:
Except nginx, they don't seem to have any breaking change. See https://nginx.org/en/CHANGES

Other than nginx, patches version are available.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -115,7 +115,7 @@ services:
       test: echo stats | nc 127.0.0.1 11211
   redis:
     <<: *restart_policy
-    image: "redis:6.2.14-alpine"
+    image: "redis:6.2.19-alpine"
     healthcheck:
       <<: *healthcheck_defaults
       test: redis-cli ping | grep PONG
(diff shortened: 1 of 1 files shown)


BYK (approved):
Sure sure sure

docker-compose.yml
```diff
@@ -133,7 +133,7 @@ services:
   postgres:
     <<: *restart_policy
     # Using the same postgres version as Sentry dev for consistency purposes
-    image: "postgres:14.11"
+    image: "postgres:14.19-bookworm"
```
hubertdeng123, line 136: Any reason to specifically use `bookworm` here? Looks like 14.11 was on debian bookworm so not a big deal but I am scarred since siwtching to postgres alpine introduced a lot of errors for us

hubertdeng123 (approved):

docker-compose.yml
aldy505: 14.11 was indeed debian bookworm. The default for `14.19` is debian trixie (debian 13). I don't know if that would cause any issues, but I would say that it's best to minimize issues for now.

And yes, we won't be moving to Alpine images.

hubertdeng123: good call

PR #3880, opened 2025-08-16
test: integration test for user feedback

aldy505:


_integration-test/test_01_basics.py
```diff
@@ -407,6 +408,36 @@ def placeholder_fn():
...
+    poll_for_response(
+        f"{SENTRY_TEST_HOST}/api/0/organizations/sentry/issues/?query=issue.category%3Afeedback",
+        client,
+        lambda x: len(json.loads(x)) > 0,
+    )
+    poll_for_response(
+        f"{SENTRY_TEST_HOST}/api/0/organizations/sentry/events/?dataset=issuePlatform&field=message&field=title&field=timestamp&project=1&statsPeriod=1h",
+        client,
+        lambda x: len(json.loads(x)["data"]) > 0,
+    )
```
aldy505, line 440: @bruno-garcia @aliu39 do you know who maintain user feedback these days? Wondering if these are the right API endpoints to check user feedback events?

aliu39: these look right to me, the main one to check is the first, though the second should also return data if project is right

aldy505:
> Can we also test all the other stuff if we are using SDKs now? Like actual profiling data, source maps, traces etc.

We already have profiling test. Source map is a good idea.

> Also, check out that Seer error please _and_ make sure using the actual SDK doesn't increase test/build times and causes flakiness (as the version may change etc.)

Seer is drunk 😆, `npm ci` is being run on `actions.yaml`.

_integration-test/test_01_basics.py
```diff
@@ -407,6 +408,36 @@ def placeholder_fn():
...
 
 @pytest.mark.skipif(os.environ.get("COMPOSE_PROFILES") != "feature-complete", reason="Only run if feature-complete")
 def test_receive_logs_events(client_login):
```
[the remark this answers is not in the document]
aldy505: `npm ci` is being run by `actions.yaml` file.

PR #3884, opened 2025-08-18
Add pgbouncer

frederikspang:
Add an optional-modification for pgbouncer. This has been extracted from our internal setup after discussion on #self-hosted in Discord.

Regarding pgbouncer, see also https://github.com/getsentry/self-hosted/pull/2740

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -142,8 +142,6 @@ services:
       ["postgres", "-c", "max_connections=${POSTGRES_MAX_CONNECTIONS:-100}"]
     environment:
       POSTGRES_HOST_AUTH_METHOD: "trust"
-      POSTGRES_USER: "postgres"
-      POSTGRES_PASSWORD: "sentry"
     volumes:
       - "sentry-postgres:/var/lib/postgresql/data"
   pgbouncer:
@@ -151,15 +149,15 @@ services:
     healthcheck:
       <<: *healthcheck_defaults
       # Using default user "postgres" from sentry/sentry.conf.example.py or value of POSTGRES_USER if provided
-      test: ["CMD-SHELL", "PGPASSWORD=\"sentry\" psql -U postgres -p 5432 -h 127.0.0.1 -tA -c \"select 1;\" -d postgres >/dev/null"]
+      test: ["CMD-SHELL", "psql -U postgres -p 5432 -h 127.0.0.1 -tA -c \"select 1;\" -d postgres >/dev/null"]
     depends_on:
       postgres:
         <<: *depends_on-healthy
     environment:
       DB_USER: ${POSTGRES_USER:-postgres}
-      DB_PASSWORD: sentry
       DB_HOST: postgres
       DB_NAME: postgres
+      AUTH_TYPE: trust
       POOL_MODE: transaction
       ADMIN_USERS: postgres,sentry
       MAX_CLIENT_CONN: 10000
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -47,7 +47,7 @@ DATABASES = {
         "ENGINE": "sentry.db.postgres",
         "NAME": "postgres",
         "USER": "postgres",
-        "PASSWORD": "sentry",
+        "PASSWORD": "",
         "HOST": "pgbouncer",
         "PORT": "",
     }


BYK:
@aldy505 shall we make this part of the default setup?

lucas-zimerman:
I think it would be great if it was, tracing UI is leaving a lot of idle connections, breaking sentry due to maximum connections reached, pgcat should help with that.

frederikspang:
@BYK If so, are there any preferences for a pgcat config file? Just a new folder+file `/pgcat/config.toml` and mount into container? (To follow same procedure as `- "./geoip:/geoip:ro"` etc

aldy505:
I'm really torn between making this default vs not.

saibotk:
Additional insight: Our setup also now regularly reaches the connection limit and that freezes Sentry entirely.
Are there any downsides to have this as a default?

frederikspang:
The reason we added this to our setup, was the same. We saw a large amount of idle connections and errors - even with low traffic in ingress and a single user doing Internal demo of trace views, and issues views.

aldy505:
Alright. Lets make it as default.

aldy505:
> @BYK If so, are there any preferences for a pgcat config file? Just a new folder+file `/pgcat/config.toml` and mount into container? (To follow same procedure as `- "./geoip:/geoip:ro"` etc

/pgcat.toml is fine, no need for a directory. Just like the redis.conf and nginx.conf.

frederikspang:
Got it. Will push update later

frederikspang:
I am out of RAM on my laptop for a test-run, which is unfortunate. 

@aldy505 Does the integration tests boot a new installed environment, or should I just boot a VM at work, to test out from scratch-installation?

aldy505:
> @aldy505 Does the integration tests boot a new installed environment

Yep, they do

frederikspang:
@aldy505 Can you approve workflow runs, so I can do a quick look using those, before spinning up a VM? (I'm gonna do both, either way)

BYK:
@frederikspang looks like it's broken:

```
 time="2025-08-19T20:54:51Z" level=warning msg="The \"PGCAT_IMAGE\" variable is not set. Defaulting to a blank string."
```

docker-compose.yml
```diff
@@ -144,6 +146,17 @@ services:
       POSTGRES_HOST_AUTH_METHOD: "trust"
     volumes:
       - "sentry-postgres:/var/lib/postgresql/data"
+  pgcat:
+    image: "$PGCAT_IMAGE"
```
aldy505: Can we just hardcode the Docker image?

docker-compose.yml
```diff
@@ -144,6 +146,17 @@ services:
...
+    ports:
+      - "5432:5432"
```
aldy505: We need to remove this. Nothing should be exposed other than `$SENTRY_BIND:80` on `nginx` container.

docker-compose.yml
```diff
@@ -144,6 +146,17 @@ services:
...
+    depends_on:
+      - postgres
```
aldy505: depends on healthy?
```suggestion
    depends_on:
      postgres:
        <<: *depends_on-healthy
```

pgcat.toml
```diff
@@ -0,0 +1,21 @@
...
+host = "0.0.0.0"
+port = 6433
+admin_username = "sentry"
+admin_password = "sentry"
+
+[pools.sentry.users.0]
+username = "sentry"
+password = "sentry"
+pool_size = 100
+min_pool_size = 100
```
aldy505: Should we provide in-code docs here to describe what's the difference between this `pool_size` compared to postgres' pool size?

aldy505 (changes requested):
It'd be great if you can add healthcheck to `pgcat` container. See https://github.com/postgresml/pgcat/pull/607

aldy505:
Anyway I'll just...

docker-compose.yml
aldy505:
```suggestion
    image: "ghcr.io/postgresml/pgcat:v1.2.0"
```

aldy505:
```
'exc': 'OperationalError(\'could not translate host name "pgcat" to address: Temporary failure in name resolution\\n\')
```

docker-compose.yml
frederikspang: Left over from debugging in our setup - my bad! :)

aldy505: It's okay. I usually debug self-hosted Sentry using GitHub Actions

frederikspang:
@aldy505 Wanna go for the healthcheck command in the linked PR?
`PGPASSWORD="<some-password>" psql -U pgcat -p 6432 -h 127.0.0.1 -tA -c "show version;" -d pgcat >/dev/null`

Or just for pg_isready like the postgres container?

aldy505:
> @aldy505 Wanan go for the healthcheck command in the linked PR? `PGPASSWORD="<some-password>" psql -U pgcat -p 6432 -h 127.0.0.1 -tA -c "show version;" -d pgcat >/dev/null`
> 
> Or just for pg_isready like the postgres container?

Up to you. What we need is to move other containers to depends on pgcat rather than postgres.

frederikspang:
they depend on both right now, in the x-sentry-defaults; We can remove postgres, no problem.

aldy505:
> they depend on both right now, in the x-sentry-defaults; We can remove postgres, no problem.

Yep yep, let's make sure pgcat is healthy first before any other container starts.

frederikspang:
@BYK Can you approve workflow run? :)

aldy505:
done

docker-compose.yml
```diff
@@ -36,8 +36,8 @@ x-sentry-defaults: &sentry_defaults
       <<: *depends_on-healthy
     kafka:
       <<: *depends_on-healthy
-    postgres:
-      <<: *depends_on-healthy
+    pgcat:
+      <<: *depends_on-default
```
aldy505: Let's make this depends on healthy instead.

```suggestion
      <<: *depends_on-healthy
```

aldy505 (approved):
nice

aminvakil (changes requested):
https://github.com/postgresml/postgresml/issues/1688#issuecomment-3041057338
https://github.com/postgresml/pgcat/issues/931#issuecomment-2940127892

Seems like postgresml/pgcat development has been stopped.

aldy505:
Ah shoot I didn't know about that.

frederikspang:
Well - That's unfortunate😂.

Back to pgbouncer?

aldy505:
> Well - That's unfortunate😂.
> 
> Back to pgbouncer?

Yeah, what other option we have anyway? 🤣

frederikspang:
@aminvakil Not sure what you mean - It's changed in `"HOST": "pgbouncer",` - But we can't force existing installs to change hosts. `sentry/sentry.conf.py` is in .gitignore, so existing installs has to update their configuration according to changes in example files, right?

aminvakil:
> @aminvakil Not sure what you mean - It's changed in `"HOST": "pgbouncer",` - But we can't force existing installs to change hosts. `sentry/sentry.conf.py` is in .gitignore, so existing installs has to update their configuration according to changes in example files, right?

You're right. I don't know how I failed to see `sentry.conf.example.py` change in this PR.

@aldy505 We might want to mention this change in release notes, so current users who want to change, get noticed as well.

aldy505:
> > @aminvakil Not sure what you mean - It's changed in `"HOST": "pgbouncer",` - But we can't force existing installs to change hosts. `sentry/sentry.conf.py` is in .gitignore, so existing installs has to update their configuration according to changes in example files, right?
> 
> 
> 
> You're right. I don't know how I failed to see `sentry.conf.example.py` change in this PR.
> 
> 
> 
> @aldy505 We might want to mention this change in release notes, so current users who want to change, get noticed as well.

Yep! That will be on the release notes.

docker-compose.yml
```diff
@@ -142,8 +142,28 @@ services:
       ["postgres", "-c", "max_connections=${POSTGRES_MAX_CONNECTIONS:-100}"]
     environment:
       POSTGRES_HOST_AUTH_METHOD: "trust"
+      POSTGRES_USER: "postgres"
+      POSTGRES_PASSWORD: "sentry"
```
aldy505: Hmm I wonder if this will break anything.

aldy505 (approved):

BYK:
>Yep! That will be on the release notes.

I think we should actually migrate existing users if they haven't touched the setting. Otherwise we'd be running this new service for no good reason.

frederikspang:
> >Yep! That will be on the release notes.
> 
> 
> 
> I think we should actually migrate existing users if they haven't touched the setting. Otherwise we'd be running this new service for no good reason.

How do you suggest?

We could do something hacky like running pgbouncer as "postgres" and postgres as "db" in docker compose.
Would pipe existing configs through pgbouncer. We need to "fix" the auth method to TRUST then, to match 1:1 existing setup in the file (in case users haven't set a password themselves as we have)

aldy505:
> > >Yep! That will be on the release notes.
> > 
> > 
> > 
> > I think we should actually migrate existing users if they haven't touched the setting. Otherwise we'd be running this new service for no good reason.
> 
> How do you suggest?
> 
> We could do something hacky like running pgbouncer as "postgres" and postgres as "db" in docker compose.
> Would pipe existing configs through pgbouncer. We need to "fix" the auth method to TRUST then, to match 1:1 existing setup in the file (in case users haven't set a password themselves as we have)

Usually we put something on the install script to modify the sentry.conf.py. This is doable, might do this on a separate PR to not make this PR being too big.

aminvakil (approved):
I agree with Burak on changing for users if they're using the default database configurations already.

@frederikspang  Thank you for adding this!

aminvakil:
On a side note, I really didn't like https://github.com/getsentry/self-hosted/pull/2740 then :)

Let's revert that after merging this and changing `sentry.conf.py`, users which have changed it in `.env.custom` will be unaffected.

docker-compose.yml
frederikspang: It might. Let me try with auth_Type=trust as it is currently

aldy505:
I want to merge this first, then (in concurrent manner):
1. Create a PR that change user's `sentry.conf.py` file
2. Revert the PR that Amin mentioned.

aldy505:
Install script is here: https://github.com/getsentry/self-hosted/pull/3898

@aminvakil can you create the revert PR, so I can quickly approve & merge?

aminvakil:
> Install script is here: #3898
> 
> @aminvakil can you create the revert PR, so I can quickly approve & merge?

Done in #3899 .

aldy505:
Cool. Let me merge this one.

PR #3898, opened 2025-08-23
feat: install script to migrate sentry.conf.py config to use pgbouncer

aldy505:
This PR may only be merged after https://github.com/getsentry/self-hosted/pull/3884 has been merged

### DIFF
--- a/clickhouse/default-password.xml
+++ b/clickhouse/default-password.xml
@@ -1,0 +1,10 @@
+<clickhouse>
+  <users>
+    <default>
+      <password></password>
+      <networks>
+        <ip>::/0</ip>
+      </networks>
+    </default>
+  </users>
+</clickhouse>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -206,7 +206,7 @@ services:
     build:
       context: ./clickhouse
       args:
-        BASE_IMAGE: "altinity/clickhouse-server:23.8.11.29.altinitystable"
+        BASE_IMAGE: "altinity/clickhouse-server:25.3.6.10034.altinitystable"
     ulimits:
       nofile:
         soft: 262144
@@ -214,10 +214,14 @@ services:
     volumes:
       - "sentry-clickhouse:/var/lib/clickhouse"
       - "sentry-clickhouse-log:/var/log/clickhouse-server"
-      - type: bind
+      - type: "bind"
         read_only: true
-        source: ./clickhouse/config.xml
-        target: /etc/clickhouse-server/config.d/sentry.xml
+        source: "./clickhouse/config.xml"
+        target: "/etc/clickhouse-server/config.d/sentry.xml"
+      - type: "bind"
+        read_only: true
+        source: "./clickhouse/default-password.xml"
+        target: "/etc/clickhouse-server/users.d/default-password.xml"
     environment:
       # This limits Clickhouse's memory to 30% of the host memory
       # If you have high volume and your search return incomplete results
--- a/install/upgrade-clickhouse.sh
+++ b/install/upgrade-clickhouse.sh
@@ -14,14 +14,46 @@ if $ps_command | grep -q clickhouse; then
   # Start clickhouse if it is not already running
   start_service_and_wait_ready clickhouse
 
-  # In order to get to 23.8, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8
+  # In order to get to 25.3, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8 -> 24.8 -> 25.3
   version=$($dc exec clickhouse clickhouse-client -q 'SELECT version()')
   if [[ "$version" == "21.8.13.1.altinitystable" || "$version" == "21.8.12.29.altinitydev.arm" ]]; then
+    echo "Detected clickhouse version $version"
     $dc down clickhouse
+
+    echo "Upgrading clickhouse to 22.8"
     $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:22.8.15.25.altinitystable clickhouse
     start_service_and_wait_ready clickhouse
     $dc down clickhouse
+
+    echo "Upgrading clickhouse to 23.3"
     $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:23.3.19.33.altinitystable clickhouse
+    start_service_and_wait_ready clickhouse
+    $dc down clickhouse
+
+    echo "Upgrading clickhouse to 23.8"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:23.8.11.29.altinitystable clickhouse
+    start_service_and_wait_ready clickhouse
+    $dc down clickhouse
+
+    echo "Upgrading clickhouse to 24.8"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:24.8.14.10459.altinitystable clickhouse
+    start_service_and_wait_ready clickhouse
+    $dc down clickhouse
+
+    echo "Upgrading clickhouse to 25.3"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:25.3.6.10034.altinitystable clickhouse
+    start_service_and_wait_ready clickhouse
+  elif [[ "$version" == "23.8.11.29.altinitystable" ]]; then
+    echo "Detected clickhouse version $version"
+    $dc down clickhouse
+
+    echo "Upgrading clickhouse to 24.8"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:24.8.14.10459.altinitystable clickhouse
+    start_service_and_wait_ready clickhouse
+    $dc down clickhouse
+
+    echo "Upgrading clickhouse to 25.3"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:25.3.6.10034.altinitystable clickhouse
     start_service_and_wait_ready clickhouse
   else
     echo "Detected clickhouse version $version. Skipping upgrades!"


install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,42 @@
+# Guard this file behind SETUP_PGBOUNCER_MIGRATION
+if [ "$SETUP_PGBOUNCER_MIGRATION" == "1" ]; then
```
aldy505: I will need to remove this flag after https://github.com/getsentry/self-hosted/pull/3884 is merged.

_unit-test/migrate-pgbouncer-test.sh
```diff
@@ -0,0 +1,199 @@
...
+  echo "$expected_db_config"
+  echo "Actual:"
+  echo "$actual_db_config"
+  exit 1
+else
+  echo "DATABASES section is correct"
+fi
+
+# Remove the file
+rm $SENTRY_CONFIG_PY /tmp/sentry_conf_py
```
aldy505, line 195: Hopefully it's okay to not clean `sentry/config.yml` and `relay/config.yml`.

aldy505:
Test results are ok:

```
Test 1 (pre 25.9.0 release)
Migrating Postgres config to PGBouncer...
  Migrating sentry/sentry.conf.py to use PGBouncer
  Migrated sentry/sentry.conf.py to use PGBouncer
DATABASES section is correct
Test 2 (post 25.9.0 release)
Migrating Postgres config to PGBouncer...
  Found pgbouncer in sentry/sentry.conf.py, I'm assuming you're good! :)
DATABASES section is correct
Test 3 (custom postgres config)
Migrating Postgres config to PGBouncer...
  ⚠️ You don't have standard configuration for Postgres in sentry/sentry.conf.py, skipping pgbouncer migration. I'm assuming you know what you're doing.
DATABASES section is correct
migrate-pgbouncer-test.sh - Success 👍
```

frederikspang:
@aldy505 LGTM

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,39 @@
+echo "${_group}Migrating Postgres config to PGBouncer..."
```
hubertdeng123, line 1: We historically have not done a sed to modify the `sentry.conf.py` files, as we chose not to modify people's config files without permission. 

I wonder if this is something that we should have a very loud announcement about, and keep it at that and have users modify their config files themselves.

aldy505: Often times there are GitHub issues complaining why certain things are broken, and the problem is that they didn't change the config file. I believe it's fine for this one, since we're trying to make everyone's instance better. 

But, if we're still not wanting to go down this route, I'm fine with it.

hubertdeng123: @BYK what do you think here?

BYK: @hubertdeng123 I'm the one who pushed for this migration as I think it will make things quite a bit better. We do a grep-based check with a warning for memcached backend here: https://github.com/getsentry/self-hosted/blob/67c32e883dae0021f5be7e31fdbcefc3cff6c880/install/check-memcached-backend.sh#L4

My argument for automatically doing this is:

1. If you are running a custom/complex setup, you probably are not using the install script anyway
2. If you are running the default setup, then we not only will be helping your setup transparently, we would also avoid running `pgbouncer` in the background for no good reason.

So if we don't want this auto migration then we should:

1. Hide the `pgbouncer` container and relevant stuff behind a (possibly new) docker compose profile
2. We should at least add the `grep`-based check and warning we have memached and steer people in the right direction.

Which approach we take depends on our risk apetite and I cannot really force either way as I won't be the one who will be bearing the consequences 😅

aldy505: I vote for merging this :)

hubertdeng123: I hear all the arguments and there is no perfect solution here, as we'd want to make this as seamless as possible for folks. However, I'd still prefer if this wasn't merged given the reasons:

1. I wouldn't feel comfortable directly editing people's config files and setting precedent that it is fine to do in the future
2. Higher risk, as we have mentioned above

Dev-infra talked about reworking how the `sentry.conf.py` config file looks in sentry since the sentry.conf.py file contains a lot of defaults, not just custom settings. To me, it makes sense to have the defaults for self-hosted live in sentry in a separate file so we wouldn't need to do this. Then, people's custom `sentry.conf.py` can override those defaults. This would pave a future where these sort of upgrades are seamless, even though this wouldn't provide relief immediately.  

This would help us make the dev environment nicer and improve the self-hosted config story. Is this something you're interested in helping out with @aldy505?

aldy505: > This would help us make the dev environment nicer and improve the self-hosted config story. Is this something you're interested in helping out with @aldy505?

I don't have much experience with Django. I will leave this for dev-infra.

BYK: > To me, it makes sense to have the defaults for self-hosted live in sentry in a separate file so we wouldn't need to do this. Then, people's custom sentry.conf.py can override those defaults. This would pave a future where these sort of upgrades are seamless, even though this wouldn't provide relief immediately.

For this to work you will need to rip the band-aid at some point so people's custom files do not override our default settings.

We have thought about this problem in the past and for now the current situation seems to work.

My recommendadion is adding this swapping in and put it behind a prompt. If it is a "headless" install, assume the user said "no" otherwise, default to true?

aldy505: +1 for making this as a prompt

aldy505:
ping @BYK

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,80 @@
...
+#         "HOST": "pgbouncer",
+#         "PORT": "",
+#     }
+# }
+# ```
+
+if sed -n '/^DATABASES = {$/,/^}$/p' "$SENTRY_CONFIG_PY" | grep -q '"HOST": "postgres"'; then
+  if [[ -z "${APPLY_AUTOMATIC_CONFIG_UPDATES:-}" ]]; then
+    echo
+    echo "We want to add PGBouncer to your Compose stack, and that would mean"
```
BYK:
```suggestion
    echo "We added PGBouncer to the default Compose stack, and to use that"
```

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,80 @@
...
+    echo "you will need to modify your sentry.conf.py file contents."
+    echo "Do you want us to do it automatically for you?"
```
BYK:
```suggestion
    echo "Do you want us to make this change automatically for you?"
```

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,80 @@
...
+#         "NAME": "postgres",
+#         "USER": "postgres",
+#         "PASSWORD": "",
+#         "HOST": "pgbouncer",
+#         "PORT": "",
+#     }
+# }
+# ```
+
+if sed -n '/^DATABASES = {$/,/^}$/p' "$SENTRY_CONFIG_PY" | grep -q '"HOST": "postgres"'; then
```
BYK, line 29: I think we should either check `SKIP_USER_CREATION`, `SKIP_USER_PROMPT` env vars, check for an interactive TTY and/or add a `SENTRY_NON_INTERACTIVE_INSTALL` option to skip all prompts in a sane way.

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,80 @@
...
+      case $yn in
+      y | yes | 1)
+        export APPLY_AUTOMATIC_CONFIG_UPDATES=1
+        echo
+        echo -n "Thank you."
+        ;;
+      n | no | 0)
+        export APPLY_AUTOMATIC_CONFIG_UPDATES=0
+        echo
+        echo -n "Alright, you will need to update your sentry.conf.py file manually before running 'docker compose up'."
```
BYK:
```suggestion
        echo -n "Alright, you will need to update your sentry.conf.py file manually before running 'docker compose up' or remove the `pgbouncer` service if you don't want to use that."
```

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,80 @@
...
+
+  if [[ "$APPLY_AUTOMATIC_CONFIG_UPDATES" == 1 ]]; then
+    echo "Migrating $SENTRY_CONFIG_PY to use PGBouncer"
+    sed -i 's/"HOST": "postgres"/"HOST": "pgbouncer"/' "$SENTRY_CONFIG_PY"
+    echo "Migrated $SENTRY_CONFIG_PY to use PGBouncer"
+  fi
+elif sed -n '/^DATABASES = {$/,/^}$/p' "$SENTRY_CONFIG_PY" | grep -q '"HOST": "pgbouncer"'; then
+  echo "Found pgbouncer in $SENTRY_CONFIG_PY, I'm assuming you're good! :)"
+else
+  echo "⚠️ You don't have standard configuration for Postgres in $SENTRY_CONFIG_PY, skipping pgbouncer migration. I'm assuming you know what you're doing."
```
BYK, line 78: Maybe a link to the changelog or just some more context around the addition of PGBouncer with a link to relevant issue or PR?

aldy505: Ah yes. Right.

install/migrate-pgbouncer.sh
aldy505: I've re-read the codebase, practically the existing `SKIP_USER_PROMPT` is  `SENTRY_NON_INTERACTIVE_INSTALL`. I'll just add that.

install/parse-cli.sh
```diff
@@ -45,6 +45,7 @@ depwarn() {
 if [ ! -z "${SKIP_USER_PROMPT:-}" ]; then
   depwarn "SKIP_USER_PROMPT variable" "SKIP_USER_CREATION"
   SKIP_USER_CREATION="${SKIP_USER_PROMPT}"
+  APPLY_AUTOMATIC_CONFIG_UPDATES="${SKIP_USER_PROMPT}"
```
BYK, line 48: Slightly nervous about this.

BYK (approved):

install/parse-cli.sh
aldy505: Why

BYK: Because we will be trying the auto migration/config update without asking about it. Ideally this would be no action on non-interactive installs that said since you try to ensure it is unchanged, I'm fine. Just a little bit nervous in case we missed something.

install/migrate-pgbouncer.sh
```diff
@@ -0,0 +1,82 @@
...
+    echo "you will need to modify your sentry.conf.py file contents."
+    echo "Do you want us to make this change automatically for you?"
+    echo
+
+    yn=""
+    until [ ! -z "$yn" ]; do
+      read -p "y or n? " yn
+      case $yn in
+      y | yes | 1)
+        export apply_config_changes_pgbouncer=1
```
aldy505, line 43: FYI @BYK I made a change for this one, to not let it populate the global `APPLY_AUTOMATIC_CONFIG_UPDATES`. Hopefully this is the right approach.

_unit-test/migrate-pgbouncer-test.sh
BYK: Should be fine

install/migrate-pgbouncer.sh
BYK: Looks good to me. Not sure if you should be `export`ing it tho.

aldy505: I'm not sure about it either. Took the safest way about it.

PR #3910, opened 2025-08-29
test: run errors-only integration tests

aldy505:
Depends on #3911

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,8 @@ jobs:
       matrix:
         os: [ubuntu-24.04, ubuntu-24.04-arm]
         container_engine: ['docker'] # TODO: add 'podman' into the list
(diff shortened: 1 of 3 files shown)


aldy505:
```
  Error response from daemon: No such image: ghcr.io/getsentry/vroom:nightly
  Error in install/ensure-correct-permissions-profiles-dir.sh:11.
  '$dcr --no-deps --entrypoint /bin/bash --user root vroom -c 'chown -R vroom:vroom /var/vroom/sentry-profiles && chmod -R o+rwx /var/vroom/sentry-profiles'' exited with status 1
  yes: standard output: Broken pipe
  yes: standard output: Broken pipe
  -> ./install.sh:main:41
  --> install/ensure-correct-permissions-profiles-dir.sh:source:11
```

yep this is expected

aminvakil (commented):
Let's merge this after https://github.com/getsentry/self-hosted/pull/3911 .

aminvakil (approved):

PR #3911, opened 2025-08-30
fix: ensuring vroom permission should be skipped on errors-only

aldy505:


aldy505:
I don't know why the arm64 integration test is failing.

BYK (approved):

install/ensure-correct-permissions-profiles-dir.sh
```diff
@@ -2,6 +2,12 @@
 
 # TODO: Remove this after the next hard-stop
 
+# Should only run when `$COMPOSE_PROFILES` is set to `feature-complete`
+if [[ "$COMPOSE_PROFILES" == "errors-only" ]]; then
+  echo "Skipping permissions check for profiles directory."
+  exit 0
```
aminvakil:
```suggestion
  exit 0
```
This prevents the next files sourced in `install.sh` to be executed, is this fine?

aminvakil (changes requested):
This change affects other files in `install` as well.

install/ensure-correct-permissions-profiles-dir.sh
aldy505: Should I enclose this within the if clause instead of exiting?

aminvakil: Yes, do not `exit 0` whatever you do :)

aminvakil (approved):

aldy505:
Alright now, the FOSSA thingy. Burak said this:

> Reported to Gavin Zee -- we enabled some deeper search in the FOSSA check which now incorrectly flags test dependencies etc
> we should be able to disable it for now

PR #3933, opened 2025-09-12
feat(tasks): Remove taskworker option override and add worker healthcheck

markstory:
Along with getsentry/sentry#99374 and this change taskworkers will be enabled by default in self-hosted. I've left the celery workers active to smooth over any tasks that are in-flight during an upgrade.

Add a worker healtcheck as we have one now.

Refs STREAM-450

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -721,7 +721,9 @@ services:
     command: run taskworker-scheduler
   taskworker:
     <<: *sentry_defaults
-    command: run taskworker --concurrency=4 --rpc-host=taskbroker:50051
+    command: run taskworker --concurrency=4 --rpc-host=taskbroker:50051 --health-check-file-path=/tmp/health.txt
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -721,7 +721,9 @@ services:
     command: run taskworker-scheduler
   taskworker:
     <<: *sentry_defaults
-    command: run taskworker --concurrency=4 --rpc-host=taskbroker:50051
+    command: run taskworker --concurrency=4 --rpc-host=taskbroker:50051 --health-check-file-path=/tmp/health.txt
+    healthcheck:
+      <<: *file_healthcheck_defaults
```
aldy505, line 726: Interesting, does this uses `rust_arroyo` under the hood?

markstory: no, but we use a similar healthcheck file setup.

PR #3939, opened 2025-09-16
ref: Add span buffer consumers

untitaker:


docker-compose.yml
```diff
@@ -435,6 +435,12 @@ services:
     command: run consumer generic-metrics-subscription-results --consumer-group query-subscription-consumer
     profiles:
       - feature-complete
+  process-spans:
```
aminvakil, line 692:
```suggestion
   process-spans:
```

aminvakil (commented):
It's missing a white space :)

docker-compose.yml
untitaker: then the indentation level would be 3, that can't be right

untitaker:
seems like it was already done in https://github.com/getsentry/self-hosted/commit/e2ad04d564385f05e6dfbc62c7ff1c16d5a81297

PR #3944, opened 2025-09-19
fix: install behind a proxy

moroine:
Updated the s3cmd installation command to include proxy environment variables.

### Legal Boilerplate


BYK (approved):
Thank you!!!

aminvakil:
This reminds me of a problem which I've faced a while ago, setting HTTP_PROXY, HTTPS_PROXY does not work, it should be lowercase.

There are lots of packages which work this way (either lowercase or uppercase), but maybe we can support both, this should be left for another PR though.

aldy505:
Oh wow I missed that. Thank you!!

PR #3946, opened 2025-09-19
chore(tasks): Remove the worker and cron containers

markstory:
These are replaced by the `taskworker` and `taskscheduler` containers

Refs STREAM-433

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -142,25 +142,6 @@ SENTRY_OPTIONS["redis.clusters"] = {
 }
 
 #########
-# Queue #
-#########
-
-# See https://develop.sentry.dev/services/queue/ for more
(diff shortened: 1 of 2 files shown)


aldy505 (approved):

sentry/sentry.conf.example.py
```diff
@@ -141,25 +141,6 @@ def get_internal_network():
...
 
-#########
-# Queue #
-#########
-
-# See https://develop.sentry.dev/services/queue/ for more
-# information on configuring your queue broker and workers. Sentry relies
-# on a Python framework called Celery to manage queues.
-
-rabbitmq_host = None
```
BYK, line 152: @aldy505 @markstory would this require a migration for existing setups or would it just be a useless config definition for them until they delete?

aldy505: Useless config definition I believe

Issue #3958, opened 2025-09-23
Cannot install or run sentry with seaweedfs behind a proxy


SteppingHat:
### Self-Hosted Version

25.9.0

### CPU Architecture

x86_64

### Docker Version

28.4.0

### Docker Compose Version

2.39.2

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

1. Set the `http_proxy`, `https_proxy` and `no_proxy` environment as per the Sentry and docker configuration when configuring an instance behind a proxy[https://develop.sentry.dev/self-hosted/#installing-behind-a-proxy](https://develop.sentry.dev/self-hosted/#installing-behind-a-proxy)
2. Run the install script as per usual

This is on a clean system, fresh install of docker and OS and no prior Sentry installation.

### Expected Result

Sentry should be able to install with no issues.

### Actual Result

What will happen is that the health check for seaweedfs will fail after it's timeout and abort the installation.

```
Docker images built.

▶ Bootstrapping seaweedfs (node store)...
 Network sentry-self-hosted_default  Creating
 Network sentry-self-hosted_default  Created
 Container sentry-self-hosted-postgres-1  Creating
 Container sentry-self-hosted-seaweedfs-1  Creating
 Container sentry-self-hosted-postgres-1  Created
 Container sentry-self-hosted-seaweedfs-1  Created
 Container sentry-self-hosted-postgres-1  Starting
 Container sentry-self-hosted-seaweedfs-1  Starting
 Container sentry-self-hosted-postgres-1  Started
 Container sentry-self-hosted-seaweedfs-1  Started
 Container sentry-self-hosted-seaweedfs-1  Waiting
 Container sentry-self-hosted-postgres-1  Waiting
 Container sentry-self-hosted-postgres-1  Healthy
container sentry-self-hosted-seaweedfs-1 is unhealthy
Error in install/bootstrap-s3-nodestore.sh:3.
'$dc up --wait seaweedfs postgres' exited with status 1
-> ./install.sh:main:39
--> install/bootstrap-s3-nodestore.sh:source:3
```

This issue seems similar to https://github.com/getsentry/self-hosted/issues/1537 where the instance of wget used in the health check does not respect the `no_proxy` environment variable, and the requests made to `seaweedfs` are being passed through the proxy, causing the heath check to fail.

SteppingHat:
Opened up a PR to resolve this - see #3959

Issue #1537, opened 2022-06-22
Problems with installing self-hosted sentry over http proxy


goganchic:
### Version

22.7.0

### Steps to Reproduce

I have a server, which can access the Internet only through http proxy. To make Docker access the Internet I perform 2 steps.

**Step 1**

Create file `/etc/systemd/system/docker.service.d/http-proxy.conf` with such contents:

```
[Service]
Environment="HTTP_PROXY=http://proxy:3128"
Environment="HTTPS_PROXY=http://proxy:3128"
Environment="NO_PROXY=127.0.0.0/8"
```

Where `proxy` is my proxy host. Then I reload systemd with a command `systemctl daemon-reload` and restart docker: `systemctl restart docker.service`.

These actions allow docker to perform `docker pull` command using http proxy.

**Step2**

Create file `~/.docker/config.json` with such contents:

```json
{
 "proxies":
 {
   "default":
   {
     "httpProxy": "http://proxy:3128",
     "httpsProxy": "http://proxy:3128",
     "noProxy": "127.0.0.0/8"
   }
 }
}
```

Where `proxy` is my proxy host. It sets `http_proxy`, `https_proxy` and `no_proxy` env variables in the docker containers, so it makes it possible to execute `curl` and other commands that require Internet access.

After it I try to execute `install.sh`.

### Expected Result

Everything works as described in the README and the documentation.

### Actual Result

Installation fails.

`apt-get` fails for [cron](https://github.com/getsentry/self-hosted/blob/master/cron/Dockerfile) container. It does not use proxy env variables and requires to configure proxies in the `/etc/apt/apt.conf`. So it can be fixed with such apt.conf:

```
Acquire::http::proxy "http://proxy:3128/";
Acquire::https::proxy "http://proxy:3128/";
```

And [cron/Dockerfile](https://github.com/getsentry/self-hosted/blob/master/cron/Dockerfile) should be modified like this:

```
ARG BASE_IMAGE
FROM ${BASE_IMAGE}
USER 0
COPY apt.conf /etc/apt/apt.conf # <-- this line added
RUN apt-get update && apt-get install -y --no-install-recommends cron && \
    rm -r /var/lib/apt/lists/*
COPY entrypoint.sh /entrypoint.sh
ENTRYPOINT ["/entrypoint.sh"]
```

Even if I fix cron container problem - the installation fails again with such an error:

```
container for service "clickhouse" is unhealthy
An error occurred, caught SIGERR on line 3
```

[...]

PR #3959, opened 2025-09-23
Unset the proxy when performing the seaweedfs health check

SteppingHat:
This PR includes a fix for running Sentry behind a proxy #3958 where the health check for the seaweedfs container is routed through the proxy due to wget not respecting the `no_proxy` header.

This issue was seen earlier with the clickhouse container, and this fix mimics the solution used to resolve the issue.

By appending the health check command with `http_proxy=''`, requests to the seaweedfs endpoints are no longer routed through the proxy, allowing the Sentry stack to be installed behind a proxy as this caused the health check to fail otherwise, preventing the stack from being installed.

### Legal Boilerplate

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -42,6 +42,12 @@ runs:
           echo "COMPOSE_PROFILES=$COMPOSE_PROFILES" >> ${{ github.action_path }}/.env
         fi
 
+    - name: Cleanup runner image
+      shell: bash
+      run: |
+        ### Inspired by https://github.com/endersonmenezes/free-disk-space ###
+        sudo rm -rf /usr/local/.ghcup
+
     - name: Setup dev environment
       shell: bash
       run: |


docker-compose.yml
```diff
@@ -274,7 +274,14 @@ services:
...
       - "sentry-seaweedfs:/data"
     healthcheck:
-      test: ["CMD", "wget", "-q", "-O-", "http://seaweedfs:8080/healthz", "http://seaweedfs:9333/cluster/healthz", "http://seaweedfs:8333/healthz"]
+      test: [
+          "CMD-SHELL",
+          # Manually override any http_proxy envvar that might be set, because
+          # this wget does not support no_proxy. See:
+          # https://github.com/getsentry/self-hosted/issues/1537
+          # https://github.com/getsentry/self-hosted/issues/3958
+          "http_proxy='' wget http://seaweedfs:8080/healthz http://seaweedfs:9333/cluster/healthz http://seaweedfs:8333/healthz || exit 1",
```
aldy505: Curious on why you dropped `-q -O-`. But overall, looks solid!

aldy505:
Integration test error:
```
  container sentry-self-hosted-seaweedfs-1 is unhealthy
  Error in install/bootstrap-s3-nodestore.sh:3.
  '$dc up --wait seaweedfs postgres' exited with status 1
  yes: standard output: Broken pipe
  yes: standard output: Broken pipe
  -> ./install.sh:main:39
  --> install/bootstrap-s3-nodestore.sh:source:3
  
```

I assume wrong healthcheck command?

docker-compose.yml
SteppingHat: This is a copy paste from my env which I dropped for testing. I seem to have forgotten to put it back, but thanks for pointing it out! I'll put it back in

SteppingHat:
@aldy505 probably was due to the related comment with the missing `-q -O-`. I checked out in a separate repo, applied the change and built the stack and all seems to be good, the tests should pass this time round.

aldy505:
> @aldy505 probably was due to the related comment with the missing `-q -O-`. I checked out in a separate repo, applied the change and built the stack and all seems to be good, the tests should pass this time round.

Okay. Just approved the CI run, let's see.

docker-compose.yml
```diff
@@ -274,7 +274,14 @@ services:
     volumes:
       - "sentry-seaweedfs:/data"
     healthcheck:
-      test: ["CMD", "wget", "-q", "-O-", "http://seaweedfs:8080/healthz", "http://seaweedfs:9333/cluster/healthz", "http://seaweedfs:8333/healthz"]
+      test: [
+          "CMD-SHELL",
+          # Manually override any http_proxy envvar that might be set, because
+          # this wget does not support no_proxy. See:
+          # https://github.com/getsentry/self-hosted/issues/1537
```
aminvakil, line 281: I do not think these issue links are necessary to be here. Please remove both.

SteppingHat: I've copied the comment from the clickhouse health check. Would you prefer the issue removed from that comment too for consistency?

aminvakil: Right, I've just saw that, let's keep only #1537 here then like clickhouse.

aminvakil (approved):
Thanks for fixing this!

BYK:
There's an incident going on with Docker Hub which is the cause of all these failures: https://www.dockerstatus.com/pages/incident/533c6539221ae15e3f000031/68d47a2f93c09e05486d93a9

Will retry once that's fixed.

SteppingHat:
> There's an incident going on with Docker Hub which is the cause of all these failures: https://www.dockerstatus.com/pages/incident/533c6539221ae15e3f000031/68d47a2f93c09e05486d93a9
> 
> Will retry once that's fixed.

Looks like it's resolved 😄

PR #3965, opened 2025-09-25
fix: logic error in s3 install script

### LINKED ISSUE #3960: Nodestore migration script issue
### Self-Hosted Version

25.9.0

### CPU Architecture

irrelevant

### Docker Version

irrelevant

### Docker Compose Version

irrelevant

### Machine Specification

### Steps to Reproduce

Run `./install.sh` to update to `25.9.0` and when asked allow migration to S3/seaweedfs

### Expected Result

The script configures the nodestore fall back to the existing data in Postgres. 

### Actual Result

The example config is copied without the fallback options.

Pretty sure this `if` is the wrong way around

https://github.com/getsentry/self-hosted/blob/18b7a4d1cccbdbd6b7d278761383160e305e0765/install/bootstrap-s3-nodestore.sh#L56-L58

it only adds the `read_through` and `delete_through` options, if the `nodestore_node` table **does not** exist.

### Event ID

_No response_

kodebach:
Fix #3960 

### Legal Boilerplate


aldy505 (approved):

aminvakil (commented):
What about users already migrated to `seaweedfs`?

BYK:
>What about users already migrated to seaweedfs?

Then this should be a noop or they should not have the nodestore table?

PR #3967, opened 2025-09-26
ci: try different runner

aldy505:


aldy505:
Could it possibly be the version number of a package we depends on?

aminvakil:
Currently `ubuntu-latest` is `ubuntu-24.04` (https://github.com/actions/runner-images?tab=readme-ov-file#available-images).

And this has also happened on another PR (https://github.com/getsentry/self-hosted/pull/3965).

aldy505:
> Currently `ubuntu-latest` is `ubuntu-24.04` (https://github.com/actions/runner-images?tab=readme-ov-file#available-images).
> 
> And this has also happened on another PR (#3965).

Yeah I should've just set `ubuntu-22.04` :(

PR #3981, opened 2025-10-04
ref: sound SENTRY_DISALLOWED_IPS on the configuration file

aldy505:
To not mislead people and to prevent time spent scrolling on GitHub issues just to fix why requests to private IP addresses are not working.

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -103,6 +103,60 @@ SENTRY_SELF_HOSTED_ERRORS_ONLY = env("COMPOSE_PROFILES") != "feature-complete"
 # AI model prices from various public APIs.
 SENTRY_AIR_GAP = False
 
+# As of 25.9.0 (September 2025 release), Sentry enforces tighter restrictions
+# of allowed IP addresses for outgoing requests. This is to prevent
+# accidentally leaking sensitive information to third parties.
+# By default, Sentry will not allow requests to private IP addresses.
+# You can override this by configuring the allowed IP addresses here.
+SENTRY_DISALLOWED_IPS: tuple[str, ...] = (
+    # https://en.wikipedia.org/wiki/Reserved_IP_addresses#IPv4
+    "0.0.0.0/8",
+    "10.0.0.0/8",
+    "100.64.0.0/10",
+    "127.0.0.0/8",
+    "169.254.0.0/16",
+    "172.16.0.0/12",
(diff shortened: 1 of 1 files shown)


sentry/sentry.conf.example.py
```diff
@@ -103,6 +103,60 @@ def get_internal_network():
 # AI model prices from various public APIs.
 SENTRY_AIR_GAP = False
 
+# As of 25.9.0 (September 2025 release), Sentry enforces tighter restrictions
+# of allowed IP addresses for outgoing requests. This is to prevent
+# accidentally leaking sensitive information to third parties.
+# By default, Sentry will not allow requests to private IP addresses.
+# You can override this by configuring the allowed IP addresses here.
+SENTRY_DISALLOWED_IPS: tuple[str, ...] = (
```
aminvakil: Is this list maintained on sentry repository and once you remove an IP from here you can send a request to that removed IP?

`SENTRY_DISALLOWED_IPS` is a little misleading name IMHO.

aldy505: > Is this list maintained on sentry repository and once you remove an IP from here you can send a request to that removed IP?

Yes. 

> `SENTRY_DISALLOWED_IPS` is a little misleading name IMHO.

cc @oioki

aminvakil: Maybe it's better for self-hosted to maintain a `SENTRY_OUTGOING_ALLOWED_IPS` list and remove the items from getsentry/sentry list?

Not that IANA is going to change this list :) It has better readability for users to maintain allowed IPs.

aldy505: That makes sense.

aldy505: Just realized, I don't think it's possible since it's a tuple and we're gonna have a lot of utils function to figure out which IP address belongs to which IP subnet...

hubertdeng123:
Looks like in the issue you linked, the user needed to provide an override?

aldy505:
> Looks like in the issue you linked, the user needed to provide an override?

Yep, therefore it requires a bit more visibility. One Google SSO plugin made by Siemens is also broken due to not providing this info.

hubertdeng123:
Having this here forces us to maintain two sources of truth though which is more painful when `SENTRY_DISALLOWED_IPS` changes. IMO it is better to include this here but commented out for visibility purposes if that is what we care about

Issue #3953, opened 2025-09-21
install/geoip.sh wrong docs


fliespl:
### Self-Hosted Version

25.9.0

### CPU Architecture

x86_64

### Docker Version

newest

### Docker Compose Version

newest

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

Docs mention to run ./install/geoip.sh to manually update database. Yet, recent changes introducted dynamic variable "CONTAINER_ENGINE".

So currently, docs should mention something like that:
`CONTAINER_ENGINE=docker bash ./install/geoip.sh`

or CONTAINER_ENGINE should fallback to docker when executed in cli context (not during install)

### Expected Result

Working update without env variable or updated docs.

```
root@sentry:/opt/self-hosted# CONTAINER_ENGINE=docker bash ./install/geoip.sh
Setting up GeoIP integration ...
Setting up IP address geolocation ...
IP address geolocation database already exists.
IP address geolocation is configured for updates.
Updating IP address geolocation database ... 
Unable to find image 'ghcr.io/maxmind/geoipupdate:v6.1.0' locally
v6.1.0: Pulling from maxmind/geoipupdate
661ff4d9561e: Pull complete 
876d32bc0ee8: Pull complete 
4d282239778e: Pull complete 
05783ca06316: Pull complete 
e874a92abc5a: Pull complete 
d439d27a1e72: Pull complete 
Digest: sha256:cdd36d36c2e1d353a990a48e5a36d42ee75089d5d9064d80d36d90c147fd2606
Status: Downloaded newer image for ghcr.io/maxmind/geoipupdate:v6.1.0
Done updating IP address geolocation database.
Done setting up IP address geolocation.
```

### Actual Result

```
root@sentry:/opt/self-hosted# bash ./install/geoip.sh 
Setting up GeoIP integration ...
Setting up IP address geolocation ...
IP address geolocation database already exists.
IP address geolocation is configured for updates.
Updating IP address geolocation database ... 
./install/geoip.sh: line 24: run: command not found
Error updating IP address geolocation database.
Error setting up IP address geolocation.
```

aldy505:
Ah you're right! Will take a proper look this weekend.

aldy505:
PR is here: https://github.com/getsentry/self-hosted/pull/3982

PR #3982, opened 2025-10-04
fix: geoip standalone script should check on CONTAINER_ENGINE variable first

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/3953

### DIFF
--- a/install/geoip.sh
+++ b/install/geoip.sh
@@ -1,3 +1,15 @@
+# If `$CONTAINER_ENGINE` is not set, we assume that we are running this script independently
+# to update the geoip database as written on the documentation.
+# Therefore we need to `source _detect-container-engine.sh` to detect the container engine.
+if [[ -z "$CONTAINER_ENGINE" ]]; then
+  if [[ -f "install/_detect-container-engine.sh" ]]; then
+    source install/_detect-container-engine.sh
+  else
+    echo "Error: Cannot find install/_detect-container-engine.sh. Defaulting to docker."
+    export CONTAINER_ENGINE="docker"
+  fi
+fi
+
 echo "${_group}Setting up GeoIP integration ..."
 
 install_geoip() {


aminvakil (changes requested):
I'd suggest adding this after `echo "${_group}Setting up GeoIP integration ..."` line.

Also `geoip.sh` may get called from any other directory or using a systemd service, ... .

It's better to check for `_detect-container-engine.sh` next to script_dir directory.
With a command like this:
```
script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
```

The plan is to make `/path/to/install/geoip.sh` and `./geoip.sh` source `_detect-container-engine.sh` and make it work.

aldy505:
> I'd suggest adding this after `echo "${_group}Setting up GeoIP integration ..."` line.
> 
> Also `geoip.sh` may get called from any other directory or using a systemd service, ... .
> 
> It's better to check for `_detect-container-engine.sh` next to script_dir directory.
> With a command like this:
> ```
> script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
> ```
> 
> The plan is to make `/path/to/install/geoip.sh` and `./geoip.sh` source `_detect-container-engine.sh` and make it work.

@aminvakil I was initially thinking to make a separate bash script rather than to tell users to use geoip.sh directly. 

So they would execute "./scripts/update-geoip.sh"


What do you think?

aminvakil:
> @aminvakil I was initially thinking to make a separate bash script rather than to tell users to use geoip.sh directly.
> 
> So they would execute "./scripts/update-geoip.sh"
> 
> What do you think?

If users were using `geoip.sh` to update their IPs, they will to continue to do so for quite some time. I'd suggest fixing this anyway.

install/geoip.sh
```diff
@@ -1,5 +1,18 @@
 echo "${_group}Setting up GeoIP integration ..."
 
+# If `$CONTAINER_ENGINE` is not set, we assume that we are running this script independently
+# to update the geoip database as written on the documentation.
+# Therefore we need to `source _detect-container-engine.sh` to detect the container engine.
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
+if [[ -z "$CONTAINER_ENGINE" ]]; then
+  if [[ -f "$script_dir/install/_detect-container-engine.sh" ]]; then
```
aminvakil:
```suggestion
  if [[ -f "$script_dir/_detect-container-engine.sh" ]]; then
```

install/geoip.sh
```diff
@@ -1,5 +1,18 @@
...
+    source $script_dir/install/_detect-container-engine.sh
```
aminvakil:
```suggestion
    source $script_dir/_detect-container-engine.sh
```

install/geoip.sh
```diff
@@ -1,5 +1,18 @@
...
+  else
+    echo "Error: Cannot find install/_detect-container-engine.sh. Defaulting to docker."
```
aminvakil:
```suggestion
    echo "Error: Cannot find _detect-container-engine.sh. Defaulting to docker."
```

aminvakil (changes requested):
script_dir is where script directory is.

aminvakil (approved):

PR #3996, opened 2025-10-11
Wrong option for taskworker

hertell:
Taskworker does not have such  option as  --health-check-file-path=/tmp/health.txt The container fails to start with the following error

taskworker-1  | Error: No such option: --health-check-file-path taskworker-1  | Usage: sentry run taskworker [OPTIONS] taskworker-1  | Try 'sentry run taskworker --help' for help.

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -742,7 +742,7 @@ services:
     command: run taskworker-scheduler
   taskworker:
     <<: *sentry_defaults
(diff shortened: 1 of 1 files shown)


aldy505:
Taskworker does indeed has `--health-check-file-path` option. Are you running an older tag?

hertell:
You are absolutely right! I messed up my upgrade prcess, and had stuff wrong.

sorry for the false alarm!!

PR #3999, opened 2025-10-14
feat: Introduce upgrade lockfile mechanism for one-time exec fixes

aldy505:


aldy505:
Not going to move forward with this

PR #3991, opened 2025-10-08
fix: missing `-dir` flag for seaweedfs

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/3990

Data won't be persisted otherwise. Folks needs to re-create the volume.

PR #4000, opened 2025-10-14
fix: Move files to correct `/data` directory if any volume-related data exists in `/tmp` for SeaweedFS

aldy505:
This is a followup for https://github.com/getsentry/self-hosted/pull/3991 and an alternative of https://github.com/getsentry/self-hosted/pull/3999 that doesn't use a lock file.

### DIFF
--- a/install/turn-things-off.sh
+++ b/install/turn-things-off.sh
@@ -1,7 +1,7 @@
 echo "${_group}Turning things off ..."
 
 # Only execute this when `seaweedfs` container is running
-if ! $dc ps --quiet seaweedfs; then
+if [ -z "$($dc ps --quiet seaweedfs)" ]; then
   echo "SeaweedFS container is not running, skipping moving tmp data."
 else
   # Only execute this when we find `*.dat` and/or `*.vif` files in `/tmp`


aldy505:
25.10.0 is released. Too late / not relevant anymore.

BYK:
@aldy505 why? We can always make a patch release and tell people to skip this one?

aldy505:
> @aldy505 why? We can always make a patch release and tell people to skip this one? 

@byk Because I've already merged the fix in another PR. If I can say, folks should skip 25.9.0 instead.

BYK:
My suggestion was for folks who have already upgraded to 25.9 but not to 25.10 yet as if they do, they'll lose data, right?

aldy505:
> > > @aldy505 why? We can always make a patch release and tell people to skip this one?
> > 
> > @BYK Because I've already merged the fix in another PR. If I can say, folks should skip 25.9.0 instead.
>
> My suggestion was for folks who have already upgraded to 25.9 but not to 25.10 yet as if they do, they'll lose data, right?

Yeah, that's true.

gertvdijk:
Got hit by this and wanted to say it may be worth to enable read-only root fs for the containers by default; it will reject writes to non-volume paths ie prevent writes data to volatile paths like this. For actual temp data paths one can use the tmpfs volume type in Docker to explicitly declare such volatile paths and allow writes there. Much less error prone. 😃

PR #4031, opened 2025-11-11
feat: statsd configuration through environment variables

aldy505:
Because of https://github.com/getsentry/sentry/pull/102988, we want to start telling people to use statsd to monitor their self-hosted sentry. This is another part of it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -69,7 +69,7 @@ x-sentry-defaults: &sentry_defaults
     SENTRY_EVENT_RETENTION_DAYS:
     SENTRY_MAIL_HOST:
     SENTRY_MAX_EXTERNAL_SOURCEMAP_SIZE:
-    SENTRY_STATSD_ADDR: ${STATSD_ADDR:-127.0.0.1:8125}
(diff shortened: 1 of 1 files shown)


sentry/sentry.conf.example.py
```diff
@@ -463,10 +463,14 @@ def get_internal_network():
 #
 # To start, uncomment the following line and adjust the options as needed.
 
-# SENTRY_METRICS_BACKEND = 'sentry.metrics.statsd.StatsdMetricsBackend'
-# SENTRY_METRICS_OPTIONS: dict[str, Any] = {
-#     'host': '100.100.123.123', # It is recommended to use IP address instead of domain name
-#     'port': 8125,
-# }
+SENTRY_STATSD_ADDR = env("STATSD_ADDR")
```
markstory: Should this use the `SENTRY_STATSD_ADDR` env var?

aldy505: Yes, good catch!

```suggestion
SENTRY_STATSD_ADDR = env("SENTRY_STATSD_ADDR")
```

PR #4040, opened 2025-11-14
fix: ensure seaweedfs lifecycle policy is set correctly

### LINKED ISSUE #4028: Changing RETENTION_DAYS with nodestore
### Self-Hosted Version

25.10.0

### CPU Architecture

x86_64

### Docker Version

28.3.2

### Docker Compose Version

28.3.2

### Machine Specification

### Steps to Reproduce

1. Run `./install.sh`
2. Change `SENTRY_EVENT_RETENTION_DAYS` in `.env`
3. Re-run `./install.sh`

### Expected Result

The new `SENTRY_EVENT_RETENTION_DAYS` set in `.env` is used.

### Actual Result

AFAICT with the new `seaweedfs` container the lifecycle policy is only set during initial setup in `install/bootstrap-s3-nodestore.sh`

The setup for the lifecycle policy is behind this `if` 
https://github.com/getsentry/self-hosted/blob/976e06525ba2049fff8bd6c3a0bdc2853e36f2bf/install/bootstrap-s3-nodestore.sh#L10-L12

and the `else` branch is just a no-op

https://github.com/getsentry/self-hosted/blob/976e06525ba2049fff8bd6c3a0bdc2853e36f2bf/install/bootstrap-s3-nodestore.sh#L66-L89

This is probably a bit of an edge case, and some additional documentation about changing the retention period might be enough. I just noticed the issue while looking at the script, while manually upgrading to 25.10.0 after skipping the nodestore change for 25.9.0.

### Event ID

_No response_

kodebach:
Fixes #4028

### Legal Boilerplate

### DIFF
--- a/install/bootstrap-s3-nodestore.sh
+++ b/install/bootstrap-s3-nodestore.sh
@@ -62,7 +62,11 @@ if [[ $(echo "$bucket_list" | tail -1 | awk '{print $3}') != 's3://nodestore' ]]
 
   $dc exec seaweedfs mkdir -p /data/idx/
   $s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' mb s3://nodestore
+else
+  echo "Node store already exists, skipping creation..."
+fi
 
+if [[ -z "${APPLY_AUTOMATIC_CONFIG_UPDATES:-}" || "$APPLY_AUTOMATIC_CONFIG_UPDATES" == 1 ]]; then
   # XXX(aldy505): Should we refactor this?
   lifecycle_policy=$(
     cat <<EOF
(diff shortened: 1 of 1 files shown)


install/bootstrap-s3-nodestore.sh
```diff
@@ -62,7 +62,11 @@ if [[ $(echo "$bucket_list" | tail -1 | awk '{print $3}') != 's3://nodestore' ]]
 
   $dc exec seaweedfs mkdir -p /data/idx/
   $s3cmd --access_key=sentry --secret_key=sentry --no-ssl --region=us-east-1 --host=localhost:8333 --host-bucket='localhost:8333/%(bucket)' mb s3://nodestore
+else
+  echo "Node store already exists, skipping creation..."
+fi
 
+if [[ -z "${APPLY_AUTOMATIC_CONFIG_UPDATES:-}" || "$APPLY_AUTOMATIC_CONFIG_UPDATES" == 1 ]]; then
```
kodebach, line 69: I added this if for users who manually changed their lifecycle policy, but this should probably be a separate CLI option or environment variable.

Ideally, we would check the existing policy matches what we expect or explicitly just update the `<Days>` part in our `<Rule>`. But that would require parsing XML and I don't think there is currently a setup to do this

aldy505: > Ideally, we would check the existing policy matches what we expect or explicitly just update the `<Days>` part in our `<Rule>`. But that would require parsing XML and I don't think there is currently a setup to do this

Yeah I don't think that's possible

BYK: There's this tool: https://github.com/TomWright/dasel

aldy505:
why did I approve it again...

aldy505:
I'm gonna merge this.

Issue #4053, opened 2025-11-19
Uptime check failure


rcampos87:
### Self-Hosted Version

25.10.0

### CPU Architecture

x86_64

### Docker Version

28.1.1

### Docker Compose Version

v2.35.1

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

Im trying to setup uptime checks for some of my internal DNS which resolves to 192.168.100.xxx ip range but im getting destination restricted.

I have UPTIME_CHECKER_ALLOW_INTERNAL_IPS enabled in my .env file

When doing curl, i can resolve to my DNS fine

### Expected Result

Uptime check to succeed

### Actual Result

[...]

PR #4054, opened 2025-11-19
Add UPTIME_CHECKER_ALLOW_INTERNAL_IPS on .env

ronaldorcampos:
This PR allows `UPTIME_CHECKER_ALLOW_INTERNAL_IPS` to be added on env file and used on docker compose uptime service.

Fixes https://github.com/getsentry/self-hosted/issues/4053

### Legal Boilerplate


aldy505:
No sorry, I don't want to go down this path. See my comment here: https://github.com/getsentry/self-hosted/issues/4053#issuecomment-3558154747

ronaldorcampos:
ok

PR #4061, opened 2025-11-25
ref: migrate to uv

aldy505:
Why I want to migrate to uv:
- More modern tooling
- Making it easy for new contributors who don't really understand Python well enough (like me) to easily setup their Python virtual environment and everything
- Although you can't run (or actually, you shouldn't) run the integration test on your local machine, one benefit is you get the typings when developing the tests locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## 25.11.1
+
+### Build / dependencies / internal 🔧
+
+- chore: remove some more unused directories by @aldy505 in [#4046](https://github.com/getsentry/self-hosted/pull/4046)
+- build(deps): bump actions/create-github-app-token from 2.1.4 to 2.2.0 by @dependabot in [#4058](https://github.com/getsentry/self-hosted/pull/4058)
+- build(deps): bump actions/checkout from 5 to 6 by @dependabot in [#4059](https://github.com/getsentry/self-hosted/pull/4059)
+- chore: uptime checker is missing on the release issue by @aldy505 in [#4047](https://github.com/getsentry/self-hosted/pull/4047)
+
+### Bug Fixes 🐛
+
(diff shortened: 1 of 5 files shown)


action.yaml
```diff
@@ -51,11 +51,17 @@ runs:
         sudo rm -rf /usr/share/dotnet /usr/share/doc/dotnet-*
         df -h
 
+    - name: Setup uv
+      uses: astral-sh/setup-uv@1e862dfacbd1d6d858c55d9b792c756523627244 # v7.1.4
+      with:
```
joshuarli, line 56: you might want to pin the `uv` version to something to avoid surprises, i'm seeing `Could not determine uv version from uv.toml or pyproject.toml. Falling back to latest.`

i also see it's using whatever system python is available (not good) `Using CPython 3.12.3 interpreter at: /usr/bin/python3` so it'd be preferable to pin the python version by writing a `.python-version` and using action/setup-python

you can achieve those points with the following which caches the entire virtualenv (even faster):

and you can also drop your `uv run pytest` since action-setup-venv activates the virtualenv as well

```
    - uses: astral-sh/setup-uv@884ad927a57e558e7a70b92f2bccf9198a4be546 # v6
      with:
        version: "0.8.2"
        # we just cache the venv-dir directly in action-setup-venv
        enable-cache: false

    - uses: getsentry/action-setup-venv@v3.0.0
      with:
        cache-dependency-path: uv.lock
        install-cmd: uv sync --frozen --active
```

action.yaml
```diff
@@ -51,11 +51,28 @@ runs:
         sudo rm -rf /usr/share/dotnet /usr/share/doc/dotnet-*
         df -h
 
+    - uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
```
joshuarli: you can remove setup-python; setup-venv does this for you and infers the version from `.python-version`

Issue #3894, opened 2025-08-21
Nginx keeps using old upstream IP of Relay after Relay restart


jerbob92:
### Self-Hosted Version

25.7.0

### CPU Architecture

x86_64

### Docker Version

28.1.1

### Docker Compose Version

2.35.1

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

Sometimes (I'm not quite sure what triggers this yet), our Relay gets recreated/restarted and gets a new IP. Afterwards the Relay becomes unavailable because nginx is still using the old upstream IP. The only way to resolve this is to restart the nginx service. 

This is probably because the nginx config does not have the `resolve` directive for the upstreams: https://nginx.org/en/docs/http/ngx_http_upstream_module.html#resolve

By default, nginx resolvers the IPs of the upstreams once and caches them forever.

So perhaps changing the upstreams to:

```
	upstream relay {
		server relay:3000 resolve;
		keepalive 2;
	}

	upstream sentry {
		server web:9000 resolve;
		keepalive 2;
	}
```

Would already fix it.

This does require the resolver to be configured: https://nginx.org/en/docs/http/ngx_http_upstream_module.html#resolver

So probably something like this:

```
    resolver 127.0.0.11 valid=5s;
```

5s might be a bit short, the default TTL is 600, which seems a bit long, maybe 60?

I already read somewhere that adding just the resolver directive on root level could be enough.

### Expected Result

Nginx should pick up the new IP address.

### Actual Result

```
nginx-1                                         | 2025/08/21 07:29:40 [error] 32#32: *1270951 connect() failed (111: Connection refused) while connecting to upstream, client: xxxx, server: , request: "POST /api/38/envelope/ HTTP/1.0", upstream: "http://10.201.8.65:3000/api/38/envelope/", host: "xxxx"
```

aldy505:
@BYK @aminvakil Do you think adding `resolver 127.0.0.11 valid=5s;` be a good option moving forward? I personally say yes.

aminvakil:
@jerbob92 What's the reason for restarting relay? Restarting container would not change the IP though, does it get recreated and why?

jerbob92:
> [@jerbob92](https://github.com/jerbob92) What's the reason for restarting relay? Restarting container would not change the IP though, does it get recreated and why?

[...]

PR #4079, opened 2025-12-04
Add resolver and zone configuration to nginx.conf

inoa-jboliveira:
This adds a DNS resolver for the web and relay container names so that Nginx can find them if they are restarted. This solves issues when these containers start before nginx eventually too.

The resolver uses the IP of the DNS of the docker daemon to be able to locate the hostnames.

Requires [Nginx 1.27.3](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#resolve) or later. Currently this project runs 1.29, so we are fine.

### Legal Boilerplate

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -74,7 +74,7 @@ http {
 	}
 
 	upstream sentry {
-        zone lb_zone 64k;
+        zone lb_zone_web 64k;
 		server web:9000 resolve;
 		keepalive 2;
 	}


nginx.conf
```diff
@@ -65,13 +65,17 @@ http {
 	proxy_read_timeout 30s;
 	proxy_send_timeout 5s;
 
+    resolver 127.0.0.11 valid=30s;
+
 	upstream relay {
-		server relay:3000;
+        zone lb_zone 64k;
```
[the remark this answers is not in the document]
inoa-jboliveira: They could have separated shared memory blocks although irrelevant here. I will change anyway

inoa-jboliveira:
```suggestion
        zone lb_zone_relay 64k;
```

nginx.conf
```diff
@@ -65,13 +65,17 @@ http {
...
+		server relay:3000 resolve;
 		keepalive 2;
 	}
 
 	upstream sentry {
-		server web:9000;
+        zone lb_zone 64k;
```
inoa-jboliveira:
```suggestion
        zone lb_zone_web 64k;
```

nginx.conf
```diff
@@ -65,13 +65,17 @@ http {
 	proxy_read_timeout 30s;
 	proxy_send_timeout 5s;
 
+    resolver 127.0.0.11 valid=30s;
```
aminvakil, line 68: Is podman DNS daemon `127.0.0.11` as well?
This may also break users who have changed their docker daemon IP address to become something else.

inoa-jboliveira: I am not familiar with podman, but I understand 127.11 is not a standard, just something docker chose. 

The solution would be to read /etc/resolv.conf and inject resolver into nginx.conf via a template.

There is a similar solution here using a custom entrypoint for nginx
https://github.com/penpot/penpot/issues/5861
https://github.com/penpot/penpot/pull/5967/changes

doc-sheet:
Just a reminder that nginx resolves things at startup. You need to reload it on IP change or use variables.
I don't know if it works with upstream block but for a simple proxy_pass you have to use tricks like
```
set $domain web;
proxy_pass http://$domain;
```

doc-sheet:
Oh, I didn't see 1.27.3 change. Nice.

inoa-jboliveira:
@doc-sheet 
The resolver change require an explicit DNS host which for docker is what I implemented.

It does not automatically get the system one (/etc/resolve.conf) but you could parse that as suggested above.

The name must be able to resolve post load of nginx. This change is very useful for me as it won't need a full restart when the web crashes

aminvakil:
I still do not find this useful, `nginx` container shall be restarted in case `relay` or `web` gets restarted after https://github.com/getsentry/self-hosted/pull/3914 got merged and it will use new IP address (if it has been changed) anyway.

aminvakil:
> This change is very useful for me as it won't need a full restart when the web crashes

With this change, `nginx` gets restarted anyway.

inoa-jboliveira:
> With this change, `nginx` gets restarted anyway.

No, it doesn't. This will happen in next install which must restart nginx. Then it can be on forever while the other containers can crash or be restarted if problematic (trust me, I have to do it from time to time)

inoa-jboliveira:
> I still do not find this useful, `nginx` container shall be restarted in case `relay` or `web` gets restarted after https://github.com/getsentry/self-hosted/pull/3914 got merged and it will use new IP address (if it has been changed) anyway.

@aminvakil I haven't seen this. It is not good practice to restart nginx, but I guess it might solve the issue sometimes...

inoa-jboliveira:
@aminvakil 
Yeah, looking at it you may create way more problems restarting nginx if relay or web die. Because it may drop healthy connections. 

I don't thing yours is a good fix for #3894

Reading the last comment there you even acknowledge that your fix isn't really going to work (my guess is the restart mechanism of docker don't work well based on healthy status). I continue getting the error months after the "fix". Plus you may cause the connection drop

inoa-jboliveira:
> I still do not find this useful, `nginx` container shall be restarted in case `relay` or `web` gets restarted after https://github.com/getsentry/self-hosted/pull/3914 got merged and it will use new IP address (if it has been changed) anyway.

But that is not really even working. I don't trust the restart mechanism of docker. 

If it did work, I wouldn't open this PR months later

inoa-jboliveira:
Well. Anyway. I will leave you guys to it.
I fixed on my local install and I'm happy.

Please feel free to close this if you don't intend to pursue this route.

Happy new year guys! Thank you for sentry!

aminvakil:
> > With this change, `nginx` gets restarted anyway.
> 
> No, it doesn't. This will happen in next install which must restart nginx. Then it can be on forever while the other containers can crash or be restarted if problematic (trust me, I have to do it from time to time)

https://docs.docker.com/compose/how-tos/startup-order/
https://github.com/getsentry/self-hosted/pull/3914#issuecomment-3246824802



> > I still do not find this useful, `nginx` container shall be restarted in case `relay` or `web` gets restarted after #3914 got merged and it will use new IP address (if it has been changed) anyway.
> 
> @aminvakil I haven't seen this. It is not good practice to restart nginx, but I guess it might solve the issue sometimes...

> @aminvakil Yeah, looking at it you may create way more problems restarting nginx if relay or web die. Because it may drop healthy connections.
> 
> I don't thing yours is a good fix for #3894

You're correct, Quoting myself from https://github.com/getsentry/self-hosted/issues/3894#issuecomment-3246821289:

> https://github.com/getsentry/self-hosted/pull/3914 ensures `nginx` gets restarted in case `relay` and / or `web` get restarted.
> 
> It's an improvement IMO, but it does not fix the real problem.

When `relay` or `web` container is down, self-hosted does not work anyway, so there is no healthy connection, but it should not happen at all, if it does there is another critical problem which should not be fixed with `nginx` getting restarted or adding `resolver` to its config, and so on.

@jerbob92 Was https://github.com/getsentry/self-hosted/pull/3914 helpful in your self-hosted instance?

aminvakil:
Oops! You had two other comments while I was writing mine :)

We had this discussion before and decided not to go with `resolver`, as OP's local installation is working and happy with it, I'm going to close this.

PR #4096, opened 2025-12-09
fix: Provide useful info on permission errors

BYK:
Prevents the very confusing 'Unsupported Docker Architecture' error when we actually had a permission error running docker.

### DIFF
--- a/install/detect-platform.sh
+++ b/install/detect-platform.sh
@@ -19,7 +19,22 @@ if [[ $CONTAINER_ENGINE == "podman" ]]; then
   FORMAT="{{.Host.Arch}}"
 fi
 
-export DOCKER_ARCH=$($CONTAINER_ENGINE info --format "$FORMAT")
+DOCKER_ARCH_OUTPUT=$($CONTAINER_ENGINE info --format "$FORMAT" 2>&1)
+DOCKER_INFO_EXIT_CODE=$?
+
+if [[ $DOCKER_INFO_EXIT_CODE -ne 0 ]]; then
+  echo "FAIL: Unable to get $CONTAINER_ENGINE architecture information."
+  echo "$DOCKER_ARCH_OUTPUT"
+  if [[ "$DOCKER_ARCH_OUTPUT" == *"permission denied"* ]]; then
+    echo ""
+    echo "You may need to add your user to the docker group:"
+    echo "  sudo usermod -aG docker \$USER"
+    echo "Then log out and log back in, or run: newgrp docker"
+  fi
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


install/detect-platform.sh
```diff
@@ -19,7 +19,22 @@ if [[ $CONTAINER_ENGINE == "podman" ]]; then
...
+  if [[ "$DOCKER_ARCH_OUTPUT" == *"permission denied"* ]]; then
+    echo ""
+    echo "You may need to add your user to the docker group:"
+    echo "  sudo usermod -aG docker \$USER"
+    echo "Then log out and log back in, or run: newgrp docker"
+  fi
+  exit 1
+fi
+
+export DOCKER_ARCH="$DOCKER_ARCH_OUTPUT"
```
Dav1dde, line 37:
```suggestion
export DOCKER_ARCH
```

Dav1dde (approved):
Technically correct info, but I am blind (pls ignore).

~~I think that should do:~~

```
dav1d@colima:xy$ FOO="123"
dav1d@colima:xy$ export FOO
dav1d@colima:xy$ bash
dav1d@colima:xy$ echo $FOO
123
dav1d@colima:xy$ exit
dav1d@colima:xy$ unset FOO
dav1d@colima:xy$ FOO="123"
dav1d@colima:xy$ bash
dav1d@colima:xy$ echo $FOO

dav1d@colima:xy$
```

install/detect-platform.sh
Dav1dde: Nevermind foget that, it's late and you changed the variable name here.

install/detect-platform.sh
```diff
@@ -19,7 +19,22 @@ if [[ $CONTAINER_ENGINE == "podman" ]]; then
...
+
+if [[ $DOCKER_INFO_EXIT_CODE -ne 0 ]]; then
+  echo "FAIL: Unable to get $CONTAINER_ENGINE architecture information."
+  echo "$DOCKER_ARCH_OUTPUT"
+  if [[ "$DOCKER_ARCH_OUTPUT" == *"permission denied"* ]]; then
+    echo ""
+    echo "You may need to add your user to the docker group:"
+    echo "  sudo usermod -aG docker \$USER"
+    echo "Then log out and log back in, or run: newgrp docker"
+  fi
```
aldy505, line 33: There are two options from here, actually:
1. Include yourself into the docker group
2. Run the install.sh script with sudo

BYK: Just got this auto-merged so happy if you wanna change :D

aminvakil: >     2. Run the install.sh script with sudo

Let's not suggest bad practices to users :)

PR #4149, opened 2026-01-30
Allow configuring Sentry taskworker concurrency via env

madest92:
It replaces the hardcoded `--concurrency` value in `docker-compose.yml` with `SENTRY_TASKWORKER_CONCURRENCY`, improving flexibility for different deployment sizes while preserving the existing default.

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -1,6 +1,6 @@
 name: Changelog Preview
 on:
-  pull_request:
+  pull_request_target:
     types:
     - opened
     - synchronize
--- a/.github/workflows/fast-revert.yml
+++ b/.github/workflows/fast-revert.yml
@@ -19,7 +19,7 @@ jobs:
     if: |
       github.event_name == 'workflow_dispatch' || github.event.label.name == 'Trigger: Revert'
     steps:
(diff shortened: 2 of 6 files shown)


.env
```diff
@@ -9,6 +9,8 @@ SENTRY_EVENT_RETENTION_DAYS=90
 SENTRY_BIND=9000
 # Set SENTRY_MAIL_HOST to a valid FQDN (host/domain name) to be able to send emails!
 # SENTRY_MAIL_HOST=example.com
+# Set SENTRY_TASKWORKER_CONCURRENCY to control the number of parallel taskworker processes
+SENTRY_TASKWORKER_CONCURRENCY=4
```
markstory: Is it worth stating that going above 32 is not advised as memory per instance goes way up, and the workers can start hitting contention on multiprocessing pipes?

markstory (approved):
Makes sense to me.

.env
madest92: I'm not sure many users will change this setting. I actually want to reduce memory usage. But I can add a comment if you think it's necessary.

madest92:
```suggestion
# Parallel taskworker processes (higher values increase memory usage; >32 not recommended)
SENTRY_TASKWORKER_CONCURRENCY=4
```

Ok?

aminvakil: This seems good to me.

madest92: @aminvakil done

PR #4191, opened 2026-02-25
fix(release): restore version bumping for releases

### LINKED ISSUE #4181: released .env file refers to nightly docker tags
### Self-Hosted Version

26.2.0

### CPU Architecture

...

### Docker Version

...

### Docker Compose Version

...

### Machine Specification

### Installation Type

Upgrade from 26.1 to 26.2

### Steps to Reproduce

read https://github.com/getsentry/self-hosted/blame/master/.env#L18

### Expected Result

should probably use the docker tag 26.2.0?

### Actual Result

`:nightly` is used

### Event ID

_No response_

BYK:
Fixes #4181

## Root cause

Bumping `minVersion` to `2.21.6` in `.craft.yml` (commit f28e0bc) silently broke the release version bump. Craft 2.21.0+ skips running `scripts/bump-version.sh` by default and instead auto-bumps based on configured targets — but the only target here is `github`, which doesn't support auto-bumping. The result: `.env` shipped with `:nightly` image tags in the 26.2.0 release.

## Changes

- **`.craft.yml`**: add `preReleaseCommand: bash scripts/bump-version.sh` to explicitly opt out of auto-bump and run the custom script
- **`scripts/bump-version.sh`**: read `CRAFT_OLD_VERSION`/`CRAFT_NEW_VERSION` env vars (Craft's preferred interface), with positional-arg fallback so `post-release.sh` continues to work unchanged; fix the README `sed` pattern that has been a no-op since the README was rewritten in Sep 2024
- **`.github/workflows/release.yml`**: remove `calver: true` — not a valid input for the Craft action (CalVer is configured via `versioning.policy` in `.craft.yml`)


.craft.yml
```diff
@@ -1,5 +1,6 @@
 minVersion: 2.21.6
 changelogPolicy: auto
+preReleaseCommand: bash scripts/bump-version.sh
```
hubertdeng123, line 3: We should need post-release script too right? 
https://github.com/getsentry/self-hosted/blob/master/scripts/post-release.sh

BYK: I think that one is automatic

BYK: `post-release.sh` should already work without changes:

1. It calls `./scripts/bump-version.sh '' 'nightly'` with **positional args**, and the updated `bump-version.sh` falls back to positional args when env vars are not set (`${CRAFT_NEW_VERSION:-${2:-}}`), so this path is preserved.
2. Craft's `postReleaseCommand` defaults to `bash scripts/post-release.sh` and does **not** have the same auto-bump override that `preReleaseCommand` has with `minVersion >= 2.21.0` — that behavior only applies to the pre-release step. So we don't need to explicitly set `postReleaseCommand` in `.craft.yml`.
3. In practice, `post-release.sh` is run manually (not via `craft publish`), so the Craft config doesn't even come into play.

PR #4203, opened 2026-03-04
feat(ci): upgrade test from latest release to nightly

aldy505:
We don't have an existing test for upgrades, we only test against new installations. Through this, I hope we can ensure that upgrade works.


BYK:
We currently do something similar to this thanks to the cached docker database stuff. Unless there's an urgent need you can point to, we should not do this as it takes too long and wastes resources.

PR #4204, opened 2026-03-04
fix: manual image tags rollback to nightly

aldy505:


aminvakil (approved):
Sorry, I didn't quite catch up what happened with 26.2.0 and what was wrong in GHA responsible for creating and fixing release tags.

But this PR is fine anyway.

aldy505:
> Sorry, I didn't quite catch up what happened with 26.2.0 and what was wrong in GHA responsible for creating and fixing release tags.
> 
> But this PR is fine anyway.

@aminvakil we merged the Craft update PR to perform new releases via Craft. On 26.2.0, the resulting docker tags in .env file on self-hosted are still "nightly" where it should be 26.2.0. Then we released 26.2.1, which fix the first problem, but causes another problem of the tags were not reverted back to "nightly".

aminvakil:
> > Sorry, I didn't quite catch up what happened with 26.2.0 and what was wrong in GHA responsible for creating and fixing release tags.
> > But this PR is fine anyway.
> 
> @aminvakil we merged the Craft update PR to perform new releases via Craft. On 26.2.0, the resulting docker tags in .env file on self-hosted are still "nightly" where it should be 26.2.0. Then we released 26.2.1, which fix the first problem, but causes another problem of the tags were not reverted back to "nightly".

@aldy505 Thanks!

PR #4235, opened 2026-03-21
ref: swap pre-commit with prek

aldy505:
Better compatiblity when developing self-hosted in all platforms, pre-commit really sucks when I'm on Windows.

Luckily, prek is a simple drop-in replacement and we need to change nothing.

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -15,5 +15,5 @@ permissions:
 
 jobs:
   changelog-preview:
-    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@f4889d04564e47311038ecb6b910fef6b6cf1363 # 2.25.0
     secrets: inherit
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
       - name: Get auth token
         id: token
-        uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2.2.1
+        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
         with:
           app-id: ${{ vars.SENTRY_RELEASE_BOT_CLIENT_ID }}
           private-key: ${{ secrets.SENTRY_RELEASE_BOT_PRIVATE_KEY }}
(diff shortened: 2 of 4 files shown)


aldy505:
Blocked by https://github.com/j178/prek/issues/1838?reload=1

aldy505:
CI is green!

aminvakil (approved):

.github/workflows/changelog-preview.yml
```diff
@@ -15,5 +15,5 @@ permissions:
 
 jobs:
   changelog-preview:
-    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@f4889d04564e47311038ecb6b910fef6b6cf1363 # 2.25.0
```
BYK, line 18: Keep this dynamic.

```suggestion
    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
```

.github/workflows/release.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     steps:
       - name: Get auth token
         id: token
-        uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2.2.1
+        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
```
BYK, line 26: This one too.

```suggestion
        uses: actions/create-github-app-token@v3
```

BYK (approved):

.github/workflows/release.yml
aldy505: Why shouldnt we pin this one?

.github/workflows/changelog-preview.yml
aldy505: I apply regular updates from dependanbot, so it's fine?

.github/workflows/release.yml
BYK: Because it's from GitHub

.github/workflows/changelog-preview.yml
BYK: And keep the churn and useless PRs and commits and CI time?

aldy505: Fair

.github/workflows/release.yml
aldy505: Well tbf we do pin actions/checkout

PR #4239, opened 2026-03-23
chore: pin GitHub Actions to full-length commit SHAs

joshuarli:
## Summary

- Pin all GitHub Actions references in `.github/` workflow files to full-length commit SHAs

Generated by `devenv pin_gha`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -15,5 +15,5 @@ permissions:
 
 jobs:
   changelog-preview:
-    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@f4889d04564e47311038ecb6b910fef6b6cf1363 # v2
     secrets: inherit
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -8,7 +8,7 @@ jobs:
     if: github.repository_owner == 'getsentry'
     runs-on: ubuntu-latest
     steps:
-      - uses: getsentry/forked-action-lock-threads@master
+      - uses: getsentry/forked-action-lock-threads@486f7380c15596f92b724e4260e4981c68d6bde6 # master
         with:
           github-token: ${{ github.token }}
           issue-lock-inactive-days: 15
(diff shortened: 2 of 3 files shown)


geoffg-sentry (approved):

hubertdeng123 (approved):

.github/workflows/changelog-preview.yml
```diff
@@ -15,5 +15,5 @@ permissions:
 
 jobs:
   changelog-preview:
-    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@f4889d04564e47311038ecb6b910fef6b6cf1363 # v2
```
BYK, line 18:
```suggestion
    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
```

BYK (changes requested):
changelog preview should stay on the tag

.github/workflows/lock.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     if: github.repository_owner == 'getsentry'
     runs-on: ubuntu-latest
     steps:
-      - uses: getsentry/forked-action-lock-threads@master
+      - uses: getsentry/forked-action-lock-threads@486f7380c15596f92b724e4260e4981c68d6bde6 # master
```
BYK, line 11: It's also a bit silly to pin this to a SHA. We should lock down its permissions instead.

aminvakil: Why not do both?

aldy505:
@aminvakil FYI internally, there is a Slack thread that's disagreeing on this whole GitHub Actions pinning organization-wide, specifically for actions from `getsentry` org. @BYK argues that he finds it okay with pinning actions made by external parties other than `getsentry` and `actions` github org.

I agree with what Burak said and want to change that for self-hosted. Majority of it because of noisy Dependabot PRs. Here's the Slack convo AI summary for the thread:

image

aldy505:
I'll drop this PR in 8 hours...

aminvakil:
Gotcha.

PR #4246, opened 2026-03-27
fix(install): gracefully handle missing containers in minimize-downtime

### LINKED ISSUE #3954: minimize-downtime not working
### Self-Hosted Version

25.9.0

### CPU Architecture

x86_64

### Docker Version

28.4.0

### Docker Compose Version

2.39.4

### Machine Specification

### Steps to Reproduce

Run upgrade with minimize-downtime

### Expected Result

No error

### Actual Result

```
nginx is missing dependency relay Error in install/wrap-up.sh:6. '$dc restart relay' exited with status 1 -> ./install.sh:main:46 --> install/wrap-up.sh:source:6
```

### Event ID

_No response_

shameemkpofficial-git:
Fixes #3954. 

When [install.sh](cci:7://file:///Users/shameem/Documents/self-hosted/install.sh:0:0-0:0) is run with `MINIMIZE_DOWNTIME=1`, the script blindly attempts to run `docker compose restart relay` and `docker compose exec nginx nginx -s reload`. If the `relay` container was removed or was completely shut down prior to running the installer, docker compose aggressively exits with code 1, which causes the bash script to crash. 

This commit checks whether the relevant containers actually exist via `docker compose ps` before executing the restart/reload, completely resolving the uncaught crash scenario while continuing to natively recreate the missing containers correctly at the end of the script using `start_service_and_wait_ready`.

### DIFF
--- a/install/wrap-up.sh
+++ b/install/wrap-up.sh
@@ -3,8 +3,18 @@ if [[ "$MINIMIZE_DOWNTIME" ]]; then
 
   # Start the whole setup, except nginx and relay.
   start_service_and_wait_ready --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay)$')
-  $dc restart relay
-  $dc exec -T nginx nginx -s reload
+
+  if [ -n "$($dc ps -q relay 2>/dev/null)" ]; then
+    $dc restart relay
+  else
+    echo "Relay container not found, skipping restart."
+  fi
+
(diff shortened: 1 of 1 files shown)


aldy505 (approved):
Woah, this is good.

aminvakil (approved):
Thanks! This is definitely an improvement!

Although I do not know why you should use `minimize-downtime` when your services are down :P 

Either way this should get fixed, thanks!

Issue #3476, opened 2024-12-18
Project setup > failed to solve: sentry-self-hosted-local: pull access denied


bobvandevijver:
### Self-Hosted Version

24.12.0

### CPU Architecture

x86_64

### Docker Version

27.3.1

### Docker Compose Version

2.29.7

### Machine Specification

- [x] My system meets the minimum system requirements of Sentry

### Steps to Reproduce

I'm resubmitting #2641 as I believe I have a reproducer, even though I am not absolutely sure it is related or even a sentry issue.

We are in the process of disabling the default `bridge` network to force all containers to have a dedicated network, and for that reason we needed to create a local builder on a dedicated `build` network.

```sh
V4_PREFIX=10.100.10
V6_PREFIX=fd00:foo:100:10

docker network create --subnet $V4_PREFIX.0/24 --gateway $V4_PREFIX.1 --ip-range $V4_PREFIX.128/25 \
        --ipv6 --subnet $V6_PREFIX::/64 --gateway $V6_PREFIX::1 --ip-range $V6_PREFIX:128::/80 build
docker buildx create --name local-builder --driver docker-container --driver-opt network=build --use
```

Docker compose automatically uses this container for its builds, but the sentry install procedure seems to have issues resolving a local image.
I tried changing `dbuild` to `docker buildx build $proxy_args` in `install/dc-detect-version.sh`, but that did not solve the issue.

### Expected Result

I expect to be able to get sentry running, even with the custom local builder.

### Actual Result

```
#0 building with "local-builder" instance using docker-container driver

#1 [sentry-cleanup internal] load build definition from Dockerfile
#1 transferring dockerfile: 469B done
#1 WARN: InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE} results in empty or invalid base image name (line 2)
#1 DONE 0.0s

#2 [sentry-cleanup internal] load metadata for docker.io/library/sentry-self-hosted-local:latest
#2 ERROR: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
------
 > [sentry-cleanup internal] load metadata for docker.io/library/sentry-self-hosted-local:latest:
------
failed to solve: sentry-self-hosted-local: failed to resolve source metadata for docker.io/library/sentry-self-hosted-local:latest: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
Error in install/build-docker-images.sh:9.
[...]

Issue #2557, opened 2023-11-08
Unable to pull `sentry-self-hosted-local` Docker image "pull access denied"


nguyenbry:
### Self-Hosted Version

23.10.1

### CPU Architecture

arm64

### Docker Version

20.10.23

### Docker Compose Version

2.15.1

### Steps to Reproduce

1. Install zip from `https://github.com/getsentry/self-hosted/releases/tag/23.10.1`
2. Extract
3. cd into folder and `./install.sh`

### Expected Result

Install scrip should complete without error and the docker image should be able to be pulled.

### Actual Result

```

#2 [internal] load metadata for docker.io/library/sentry-self-hosted-local:latest
#2 ...

#3 [auth] library/sentry-self-hosted-local:pull token for registry-1.docker.io
#3 DONE 0.0s

#2 [internal] load metadata for docker.io/library/sentry-self-hosted-local:latest
#2 ERROR: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
------
 > [internal] load metadata for docker.io/library/sentry-self-hosted-local:latest:
------
failed to solve: sentry-self-hosted-local: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
Error in install/build-docker-images.sh:8.
'$dcb --force-rm "$service"' exited with status 17
-> ./install.sh:main:29
--> install/build-docker-images.sh:source:8

Cleaning up...

```

azaslavsky:
This is almost certainly an issue with either your local docker settings, or a firewall/VPN you are behind. I just ran the install script with no issues, so I don't think its down remotely.

barthez:
I had the same issue. I managed to build Sentry self-hosted by disabling the docker buildkit:
```
export DOCKER_BUILDKIT=0
```
However, I don't think it is the permanent solution since I can now see a deprecation warning:
```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.
```

Issue #2302, opened 2023-07-26
sentry-self-hosted-local: pull access denied


nikserg:
### Self-Hosted Version

23.7.1

### CPU Architecture

x86_64

### Docker Version

20.10.21

### Docker Compose Version

2.13.0

### Steps to Reproduce

Run `./install.sh`

### Expected Result

Sentry is installed

### Actual Result

See an error after some time of installation:

```
failed to solve: sentry-self-hosted-local: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
Error in install/build-docker-images.sh:9.
'$dcb --force-rm "$service"' exited with status 17
-> ./install.sh:main:29
--> install/build-docker-images.sh:source:9
```

hubertdeng123:
Is this a fresh install of Sentry? Did you happen to change any variables in your `.env` file? Wondering if you're referencing the wrong docker images

Issue #2248, opened 2023-06-30
Install script is failing with docker compose build, pulling from docker.io/library even tho the image is locally available.


macnibblet:
### Self-Hosted Version

23.6.1

### CPU Architecture

x86_64

### Docker Version

  24.0.2

### Docker Compose Version

2.18.1

### Steps to Reproduce

Upgrading from 22.9.0

./install.sh

### Expected Result

Success?

### Actual Result

```++ [[ -f .env.custom ]]
++ _ENV=.env
+++ mktemp
++ t=/tmp/tmp.CsAYRRhWJM
++ export -p
++ set -a
++ . .env
+++ COMPOSE_PROJECT_NAME=sentry-self-hosted
+++ SENTRY_EVENT_RETENTION_DAYS=90
+++ SENTRY_BIND=9000
+++ SENTRY_IMAGE=getsentry/sentry:23.6.1
+++ SNUBA_IMAGE=getsentry/snuba:23.6.1
+++ RELAY_IMAGE=getsentry/relay:23.6.1
+++ SYMBOLICATOR_IMAGE=getsentry/symbolicator:23.6.1
+++ VROOM_IMAGE=getsentry/vroom:23.6.1
+++ WAL2JSON_VERSION=latest
+++ HEALTHCHECK_INTERVAL=30s
+++ HEALTHCHECK_TIMEOUT=60s
+++ HEALTHCHECK_RETRIES=10
++ set +a
++ . /tmp/tmp.CsAYRRhWJM
+++ declare -x HOME=/home/ansible
+++ declare -x LANG=C.UTF-8
+++ declare -x LC_CTYPE=C.UTF-8
+++ declare -x 'LESSCLOSE=/usr/bin/lesspipe %s %s'
+++ declare -x 'LESSOPEN=| /usr/bin/lesspipe %s'
+++ declare -x LOGNAME=root
[...]

PR #4250, opened 2026-03-28
fix: use default buildx builder to resolve local images during build

maiqigh:
## Problem

When the active buildx builder uses the `docker-container` driver (which is increasingly common as Docker defaults to BuildKit), services like `sentry-cleanup` that depend on locally built images (`sentry-self-hosted-local`) via `FROM ${BASE_IMAGE}` fail with:

```
failed to solve: sentry-self-hosted-local: failed to resolve source metadata for
docker.io/library/sentry-self-hosted-local:latest: pull access denied
```

The `docker-container` driver runs in isolation and cannot access host-local images, so BuildKit attempts to pull from Docker Hub instead.

This has been reported multiple times: #3476, #2557, #2302, #2248, #1781.

## Solution

Set `BUILDX_BUILDER=default` in `install/build-docker-images.sh` before building images. The `default` builder uses the `docker` driver, which has direct access to the local image store.

- Only applied when `CONTAINER_ENGINE=docker` (no effect on Podman)
- Scoped to the build script only
- Does not use the deprecated `DOCKER_BUILDKIT=0` flag
- References upstream issue: https://github.com/moby/buildkit/issues/4162

### DIFF
--- a/install/build-docker-images.sh
+++ b/install/build-docker-images.sh
@@ -1,6 +1,16 @@
 echo "${_group}Building and tagging Docker images ..."
 
 echo ""
+
+# Use the default buildx builder to ensure locally built images
+# (e.g., sentry-self-hosted-local) are accessible during the build.
+# Non-default builders using the docker-container driver run in isolation
+# and cannot resolve local images, causing "pull access denied" errors.
+# See: https://github.com/moby/buildkit/issues/4162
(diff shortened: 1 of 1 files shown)


aminvakil (approved):
This is a nice default to have indeed.

Thanks!

aldy505 (approved):
Nice

PR #4267, opened 2026-04-07
feat: Add launchpad taskworker container

NicoHinderling:
Adds a `launchpad-taskworker` service to the self-hosted docker-compose setup. This allows Launchpad (preprod artifact processing) to run alongside Sentry in self-hosted deployments.

The launchpad taskworker shares the existing taskbroker and Kafka infrastructure with the sentry taskworker. Because each application registers with a different `application` value (`"launchpad"` vs `"sentry"`), workers only fetch tasks from their own application — no cross-contamination.

**Changes:**
- `.env`: Add `LAUNCHPAD_IMAGE` and `LAUNCHPAD_TASKWORKER_CONCURRENCY` variables
- `docker-compose.yml`: Add `launchpad-taskworker` service with taskbroker, kafka, and web dependencies

**Before merging, these need to be resolved:**
- `LAUNCHPAD_IMAGE` — currently set to `nightly`; should be pinned to a release tag
- `LAUNCHPAD_RPC_SHARED_SECRET` — left empty; needs the shared secret for authenticating API calls back to sentry
- `OBJECTSTORE_URL` — not included (app icon processing is skipped without it); may need `http://seaweedfs:8333` if required

## Testing

1. Pull the launchpad image:
   ```bash
   docker pull ghcr.io/getsentry/launchpad:nightly
   ```

2. Start the new service:
   ```bash
   docker compose up -d launchpad-taskworker
   ```

3. Verify the container starts and connects to the taskbroker:
   ```bash
   docker compose logs -f launchpad-taskworker
   ```
   Expected: container starts, connects to `taskbroker:50051`, and waits for tasks.

4. Verify health check passes:
   ```bash
   docker compose ps launchpad-taskworker
   ```
   Should show `healthy` status after the start period.

Ref: the sentry taskworker definition at L771-L775 was the template for this, per Mark Story's guidance.


aldy505 (approved):

action.yaml
```diff
@@ -215,12 +215,26 @@ runs:
       with:
         node-version: "22.x"
 
+    - name: Setup Sentry CLI
+      shell: bash
+      run: curl -sL https://sentry.io/get-cli/ | SENTRY_CLI_VERSION="3.3.5" bash
```
runningcode: We should update to 3.4.1 otherwise the object store authentication won't work.

runningcode (approved):

PR #4269, opened 2026-04-08
fix: disable a few more new clickhouse tables

### LINKED ISSUE #4266: Clickhouse config seems to be missing some remove directives
### Self-Hosted Version

26.3.1

### CPU Architecture

x86_64

### Docker Version

29.3.1

### Docker Compose Version

5.1.1

### Machine Specification

### Installation Type

Very old install, but running fine

### Steps to Reproduce

1. Run an instance for some time I guess ?

### Expected Result

Normal cpu usage

I noticed that clickhouse config.xml seems to not include `error_log`, `latency_log`, `query_views_log`, `processors_profile_log`, which all cause a LOT more CPU usage than needed. I recently disabled them for my own install and noticed much improved performance. Did they slip through when they got introduced recently ? Seems clickhouse config.xml hasn't seen an update in quite some time, but I'm asking since there may be a reason to this.

### Actual Result

Big cpu usage

### Event ID

_No response_

alkanna:
Closes #4266

### Legal Boilerplate

### DIFF
--- a/clickhouse/config.xml
+++ b/clickhouse/config.xml
@@ -15,6 +15,10 @@
     <!-- Update: Required for newer versions of Clickhouse -->
     <session_log remove="remove"/>
     <part_log remove="remove"/>
+    <query_views_log remove="remove"/>
+    <processors_profile_log remove="remove"/>
+    <error_log remove="remove"/>
+    <latency_log remove="remove"/>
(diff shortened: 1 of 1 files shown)


aldy505 (approved):
why not

PR #4295, opened 2026-04-21
fix(nginx): add resolver and upstream resolve to prevent stale IP rou…

ilias-115:
## Why

In the default self-hosted Docker Compose deployment, nginx proxies traffic to
the `relay` and `web` services by hostname.

When those containers are recreated and receive a new container IP, nginx may
continue using a stale upstream address until nginx itself is reloaded or
restarted.

Observed symptom:
- `connect() failed (...) while connecting to upstream`
- nginx attempts to reach a stale upstream container IP

## What changed

- Added a DNS resolver in `nginx.conf`
- Enabled runtime DNS re-resolution for upstream backends:
  - `relay:3000 resolve`
  - `web:9000 resolve`
- Added `zone` directives required by nginx for dynamic upstream resolution

## Config details

- `resolver 127.0.0.11 valid=30s;`
- `resolver_timeout 5s;`
- `zone relay 64k;`
- `zone sentry 64k;`

## Why these values

- `127.0.0.11`: Docker embedded DNS in the default self-hosted container network
- `valid=30s`: balances recovery time and DNS lookup overhead
- `resolver_timeout=5s`: avoids long stalls on DNS issues
- `64k zone`: sufficient for these small upstream groups

## Test plan

- [x] nginx starts successfully with the updated config
- [x] traffic is routed correctly to `relay` and `web`
- [x] after recreating `relay`, nginx no longer keeps using a stale upstream IP
- [x] manual local verification of continued ingestion after backend IP change

## Notes

This change is intended to make nginx more resilient to backend container IP
changes in the default self-hosted Docker Compose setup.

### Legal Boilerplate

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -65,13 +65,20 @@ http {
 	proxy_read_timeout 30s;
 	proxy_send_timeout 5s;
 
+	# Resolve container hostnames through Docker DNS so upstream IP changes
+	# (after container recreation) are picked up without requiring nginx restart.
+	resolver 127.0.0.11 valid=30s;
+	resolver_timeout 5s;
+
 	upstream relay {
-		server relay:3000;
+		zone relay 64k;
+		server relay:3000 resolve;
 		keepalive 2;
 	}
 
 	upstream sentry {
-		server web:9000;
+		zone sentry 64k;
+		server web:9000 resolve;
 		keepalive 2;
 	}
 


aminvakil:
This has been discussed in https://github.com/getsentry/self-hosted/pull/4079 and has been decided not to add this then.

moroine:
I see the workaround with the `depends_on` but somehow it didn't worked on our case. I think it's not restarted when web or relay are recreated after a crash.

I get the point with the DNS server which can vary depending on host, IMO we have 2 ways:
- retrieve from `/etc/resolv.conf`
- use the old fashion way:
```
set $upstream "http://relay:3000";
proxy_pass $upstream;
```

ilias-115:
> This has been discussed in #4079 and has been decided not to add this then.

This PR comes from a real production incident on our side, not from a theoretical optimization.

According to Docker docs, `depends_on` only guarantees dependent restart on **explicit Compose operations** (e.g. docker compose restart), not all runtime lifecycle events.
For crash/OOM/runtime restarts, behavior is controlled by Docker restart policies (always, unless-stopped, on-failure).

For DNS portability, we can avoid hardcoding by retrieving resolver IPs from `/etc/resolv.conf`.


_References:_
https://docs.docker.com/compose/how-tos/startup-order/
-> “`restart: true` ensures that if db is updated or restarted due to **an explicit Compose operation**, for example `docker compose restart`"
https://github.com/docker/compose/issues/12477#issuecomment-2735544977
-> `systemctl restart docker` is unrelated to docker compose.

aminvakil:
Sorry, I've mentioned the wrong pull request.
Here is the issue:
https://github.com/getsentry/self-hosted/issues/3894

> I see the workaround with the `depends_on` but somehow it didn't worked on our case. I think it's not restarted when web or relay are recreated after a crash.
> 
> I get the point with the DNS server which can vary depending on host, IMO we have 2 ways:
> 
>     * retrieve from `/etc/resolv.conf`
> 
>     * use the old fashion way:
> 

Using DNS servers from `/etc/resolv.conf` does not work in compose, your upstream DNS server in `/etc/resolv.conf` do not know what is `web` and `relay`.
> 
> ```
> set $upstream "http://relay:3000";
> proxy_pass $upstream;
> ```

Simply bad practice, you should use `resolver` when you need this behaviour, do not push random configurations to push your PR.

> > This has been discussed in #4079 and has been decided not to add this then.
> 
> This PR comes from a real production incident on our side, not from a theoretical optimization.
> 
> According to Docker docs, `depends_on` only guarantees dependent restart on **explicit Compose operations** (e.g. docker compose restart), not all runtime lifecycle events. For crash/OOM/runtime restarts, behavior is controlled by Docker restart policies (always, unless-stopped, on-failure).
> 

Agreed, but as I've stated in https://github.com/getsentry/self-hosted/issues/3894, the real fix would be to understand why `relay` has been crashed, not pushing latency to all users in self-hosted.

> For DNS portability, we can avoid hardcoding by retrieving resolver IPs from `/etc/resolv.conf`.

Answered above, it does not work.

> 
> _References:_ https://docs.docker.com/compose/how-tos/startup-order/ -> “`restart: true` ensures that if db is updated or restarted due to **an explicit Compose operation**, for example `docker compose restart`" [docker/compose#12477 (comment)](https://github.com/docker/compose/issues/12477#issuecomment-2735544977) -> `systemctl restart docker` is unrelated to docker compose.

You're correct. Current `startup-order` does not handle all the problems which may arise in `self-hosted`, but I still do not think we should use `resolver` in all self-hosted installations of all users.

But I'm not a maintainer of this project and I only state my opinions here, my word is not final :)

@aldy505 What do you think?

ilias-115:
> Sorry, I've mentioned the wrong pull request. Here is the issue: #3894
> 
> > I see the workaround with the `depends_on` but somehow it didn't worked on our case. I think it's not restarted when web or relay are recreated after a crash.
> > I get the point with the DNS server which can vary depending on host, IMO we have 2 ways:
> > ```
> > * retrieve from `/etc/resolv.conf`
> > 
> > * use the old fashion way:
> > ```
> 
> Using DNS servers from `/etc/resolv.conf` does not work in compose, your upstream DNS server in `/etc/resolv.conf` do not know what is `web` and `relay`.
> 
> > ```
> > set $upstream "http://relay:3000";
> > proxy_pass $upstream;
> > ```
> 
> Simply bad practice, you should use `resolver` when you need this behaviour, do not push random configurations to push your PR.
> 
> > > This has been discussed in #4079 and has been decided not to add this then.
> > 
> > 
> > This PR comes from a real production incident on our side, not from a theoretical optimization.
> > According to Docker docs, `depends_on` only guarantees dependent restart on **explicit Compose operations** (e.g. docker compose restart), not all runtime lifecycle events. For crash/OOM/runtime restarts, behavior is controlled by Docker restart policies (always, unless-stopped, on-failure).
> 
> Agreed, but as I've stated in #3894, the real fix would be to understand why `relay` has been crashed, not pushing latency to all users in self-hosted.
> 
> > For DNS portability, we can avoid hardcoding by retrieving resolver IPs from `/etc/resolv.conf`.
> 
> Answered above, it does not work.
> 
> > _References:_ https://docs.docker.com/compose/how-tos/startup-order/ -> “`restart: true` ensures that if db is updated or restarted due to **an explicit Compose operation**, for example `docker compose restart`" [docker/compose#12477 (comment)](https://github.com/docker/compose/issues/12477#issuecomment-2735544977) -> `systemctl restart docker` is unrelated to docker compose.
> 
> You're correct. Current `startup-order` does not handle all the problems which may arise in `self-hosted`, but I still do not think we should use `resolver` in all self-hosted installations of all users.
> 
> But I'm not a maintainer of this project and I only state my opinions here, my word is not final :)
> 
> @aldy505 What do you think?

Thanks for the detailed feedback — this makes sense.

Just to clarify one point: when I mentioned retrieving DNS from `/etc/resolv.conf`, I meant the file inside the nginx container, not on the host. In Docker user-defined networks this is typically Docker’s embedded DNS, which resolves `web` and `relay` service names.

Also, we hit this not only after crashes, but during normal operations too (VM reboot / Docker daemon restart). So this is not only a “fix crash root cause” topic — it is also an upstream DNS resilience issue for expected runtime events where container IPs can change.

I agree root-cause work is still needed, but it does not remove the need for robust upstream re-resolution in nginx.

If changing default behavior is out of scope, I’m happy to re-scope this as:
- an opt-in behavior
- documentation-backed workaround for users affected by stale upstream IPs.

aminvakil:
> Just to clarify one point: when I mentioned retrieving DNS from `/etc/resolv.conf`, I meant the file inside the nginx container, not on the host. In Docker user-defined networks this is typically Docker’s embedded DNS, which resolves `web` and `relay` service names.
> 
Thanks for clarification, that makes sense.

> Also, we hit this not only after crashes, but during normal operations too (VM reboot / Docker daemon restart). So this is not only a “fix crash root cause” topic — it is also an upstream DNS resilience issue for expected runtime events where container IPs can change.
> 
That's highly unusual. I'm not saying this is impossible and containers created by `docker-compose` IPs never get changed, but as long as they have not been recreated, they remain the same usually, also on recreating them `nginx` container uses new `relay` and `web` IP as it depends on them getting healthy.

> I agree root-cause work is still needed, but it does not remove the need for robust upstream re-resolution in nginx.
> 
> If changing default behavior is out of scope, I’m happy to re-scope this as:
> 
>     * an opt-in behavior
> 
>     * documentation-backed workaround for users affected by stale upstream IPs.
I agree that this could be an opt-in behvaiour, but I'll be happy to review your PR regarding that, but I think in current scope of project and `nginx.conf` being tracked by git, you should probably write a install script which checks an env or something like that, but that's only my solution which is not good at all and I do not recommend it at all :)

But maybe you can fix the git tracking issue of `nginx.conf` somehow :shrug: 

Better solution would be documenting it as there have been more than a couple users claiming this has fixed their problems, but I still do not agree that setting `resolver` was the solution.

spawnia:
Hi @ilias-115 — we hit exactly this in production and have opened #4492, which carries your change with you as co-author.

It is the same two-line fix; what is new is a response to the objections raised here and in #3894:

- The DNS cost is a fixed 0.2 queries/s against Docker's in-process embedded DNS, driven by a per-upstream timer rather than by requests, so it does not scale with load — which was @BYK's concern about "heavy load".
- Hardcoding `127.0.0.11` is deliberate and now carries a comment: on a Compose network it is the only resolver that knows the names `relay` and `web`, so there is no portable alternative to substitute for it.
- Why #3914 does not close the gap: it covers `docker compose restart`, not a host-level Docker restart, which is how ours broke.

Happy for whichever of the two lands. Feel free to close this in favour of #4492, or say the word and we will close ours instead.

ilias-115:
Hi @spawnia 
Thanks for picking this up and keeping me as co-author. #4492  makes the case better, so closing in favor of it.

spawnia:
@ilias-115 Correcting something I told you here: I claimed #3914 does not cover the failure because ours came from a host-level Docker restart. It did not. Our `docker inspect` records show a plain `docker compose down` / `up -d`, with relay never moving — nginx cached a wrong address 12 s before relay started. That is the case `restart: true` already covers, and we hit it only because our pin predates #3914.

So the reason I gave for closing this in favour of #4492 was wrong, and I am sorry for that. #4492 was then closed as too much change to the default config. #4498 offers the same thing as an opt-in patch under `optional-modifications/` and has changes requested on maintainability grounds.

PR #4298, opened 2026-04-23
fix(docker): install custom packages into runtime venv

KimSeoYeon23:
## Summary

  This installs custom Python packages into Sentry's runtime virtualenv instead of the
  system Python.

  `self-hosted` runs Sentry from `/.venv`, but `sentry/Dockerfile` was installing the
  bundled nodestore extension with `pip`, which targets the system interpreter. That can
  leave custom packages unavailable at runtime and cause the `web` service to fail during
  startup.

  ## Root cause

  The image build used:

  ```dockerfile
  RUN pip install ...
```

  while the actual Sentry runtime uses /.venv.

  In my case, this caused the nodestore backend package to be installed outside the runtime
  environment, which led to repeated web worker exits and an unhealthy web container.

  ## Fix

  Use /.venv/bin/pip for custom package installation in sentry/Dockerfile.

  This updates:

  - the nodestore extension install step
  - the optional requirements.txt install step

  ## Validation

  After rebuilding the image and restarting the stack:

  - the custom package is installed into /.venv
  - web workers stop exiting during startup
  - sentry-self-hosted-web-1 becomes healthy

  

  ### Legal Boilerplate

  Look, I get it. The entity doing business as "Sentry" was incorporated in the State of
  Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in
  order to utilize my contributions in this here PR. So here's the deal: I retain all
  rights, title and interest in and to my contributions, and by keeping this boilerplate
  intact I confirm that Sentry can use, modify, copy, and redistribute my contributions,
  under Sentry's choice of terms.zhem

### DIFF
--- a/sentry/Dockerfile
+++ b/sentry/Dockerfile
@@ -1,7 +1,7 @@
 ARG SENTRY_IMAGE
 FROM ${SENTRY_IMAGE}
 
-RUN pip install https://github.com/getsentry/sentry-nodestore-s3/archive/main.zip
+RUN /.venv/bin/pip install https://github.com/getsentry/sentry-nodestore-s3/archive/main.zip
 
 COPY . /usr/src/sentry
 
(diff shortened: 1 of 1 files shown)


aldy505 (approved):
Looks good to me, let's see if the integration tests passed, and I'll merge

aminvakil (changes requested):
I do not think this should be merged.

https://github.com/getsentry/sentry/blob/3fb85962b4d7295920f27d687500089cb49beba2/self-hosted/Dockerfile#L42-L50

upstream `sentry` Dockerfile uses `/.venv/bin/` right now and `sentry/Dockerfile` in this repository already adapts that by executing `pip` as it is.

Although this does not harm right now, but we should keep up with `pip` in upstream sentry Dockerfile if we change this.

Right now there is no problem to be solved AFAICS.

PR #4311, opened 2026-04-29
fix: force disable HTTP(S) proxy for seaweedfs services

### LINKED ISSUE #4300: SeaweedFS unhealthy after upgrading to 26.4.1 behind a corporate proxy
### Self-Hosted Version

26.4.1

### CPU Architecture

x86_64

### Docker Version

29.1.3

### Docker Compose Version

2.33.1

### Machine Specification

### Installation Type

Upgrade from 26.3.1

### Steps to Reproduce

1. Prerequisite: Sentry host behind an explicit corporate proxy (you must have HTTP(S)_PROXY vars set to go online)
2. Bump the repo from 26.3.1 to 26.4.1
3. `./install.sh`
4. Notice the installation fails as SeaweedFS reports itself as unhealthy.

### Expected Result

Installing Sentry works.

### Actual Result

Alas, it does not.

Everything including the no_proxy configuration from the self-hosted docs index page is set correctly, yet SeaweedFS is unable to start. The logs suggest something inside of SeaweedFS is trying to connect to itself via the proxy:

```
I0423 12:48:00.090499 masterclient.go:454 .master masterClient reconnection attempt #102
I0423 12:48:01.385576 volume_grpc_client_to_master.go:44 checkWithMaster 172.18.0.2:9333: get master 172.18.0.2:9333 configuration: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: failed to do connect handshake, response: \"<response from the proxy>\""
I0423 12:48:01.799599 masterclient.go:454 .master masterClient reconnection attempt #103
I0423 12:48:03.490824 masterclient.go:454 .master masterClient reconnection attempt #104
...
```

I added a `docker-compose.override.yml` file with:

```
services:
  seaweedfs:
    environment:
      HTTP_PROXY: ""
      HTTPS_PROXY: ""
      http_proxy: ""
      https_proxy: ""
```

And now it installs and starts up correctly. The health check contains an explicit attempt to unset `http_proxy` too. I'm not sure if SeaweedFS ever has any business going online for self-hosted instances without replication etc (doesn't seem like it), so maybe these overrides could be set for the service globally?

### Event ID

_No response_

DragoonAethis:
This is an attempt to fix #4300 where SeaweedFS is unable to start successfully if HTTP(S)_PROXY vars are set.

If the host container runtime sets these variables, for example behind a corporate proxy, SeaweedFS services are unable to talk to each other by default. Since they have no business talking to the outside world, just force clear these variables.

This is not the first time those vars cause a weird problem, so I opted to put them in an anchored block for easy inclusion into however many services these might be required in. (Maybe it would be a good way to simplify the whole no_proxy config in the README too?)

Ping @aldy505 for attention

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,11 @@ x-file-healthcheck: &file_healthcheck_defaults
   timeout: "$HEALTHCHECK_FILE_TIMEOUT"
   retries: $HEALTHCHECK_FILE_RETRIES
   start_period: "$HEALTHCHECK_FILE_START_PERIOD"
+x-force-no-proxy: &force_no_proxy
+  HTTP_PROXY: ""
+  HTTPS_PROXY: ""
+  http_proxy: ""
+  https_proxy: ""
 x-sentry-defaults: &sentry_defaults
   <<: [*restart_policy, *pull_policy]
   image: sentry-self-hosted-local
@@ -275,6 +280,7 @@ services:
         -ip.bind=0.0.0.0
         -webdav=false
     environment:
+      <<: *force_no_proxy
       AWS_ACCESS_KEY_ID: sentry
       AWS_SECRET_ACCESS_KEY: sentry
     volumes:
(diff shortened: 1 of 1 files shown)


aldy505 (approved):
Nice. Thank you!

PR #4312, opened 2026-04-29
feat: enable Metrics

### LINKED ISSUE #3993: Add Trace Metrics
### Problem Statement

This is a different feature from https://github.com/getsentry/self-hosted/issues/2698 -- this is called "Metrics" on the UI, but it's under the codename of "tracemetrics".

### Solution Brainstorm

Soon

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/3993

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,5 +15,5 @@ dev = [
     "pytest-cov>=4.1.0",
     "pytest-rerunfailures>=11.0",
     "pytest-sentry>=0.1.11",
-    "sentry-sdk>=2.4.0,<3.0.0",
+    "sentry-sdk==2.58.0",
 ]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -15,5 +15,5 @@ dev = [
     "pytest-cov>=4.1.0",
     "pytest-rerunfailures>=11.0",
     "pytest-sentry>=0.1.11",
-    "sentry-sdk>=2.4.0,<3.0.0",
+    "sentry-sdk==2.58.0",
```
aminvakil, line 18: Why `sentry-sdk` version has been pinned to `2.58.0`? Why not `>=2.58.0` ?

aldy505: I need to update the `uv.lock` file to get the one version that has metrics support. I'm not doing `>=2.58.0` so I can intentionally update the SDK, and understand what changed. I don't use the Python SDK as much, so I believe this is one way to do so.

aminvakil: I think we should have another GHA like `dependabot` (can this be added to it?) to look for newer versions.

That can be added later though.

aminvakil (approved):

BYK (approved):

PR #4313, opened 2026-04-29
fix(ci): Make integration tests on Linux arm64 not flaky

aldy505:
Closes SELF-93


BYK (approved):

sentry/sentry.conf.example.py
```diff
@@ -221,7 +221,17 @@ def get_internal_network():
 DEFAULT_KAFKA_OPTIONS = {
     "bootstrap.servers": "kafka:9092",
     "message.max.bytes": 50000000,
-    "socket.timeout.ms": 1000,
+    "socket.timeout.ms": 10000,         # Timeout for individual socket operations (send/recv)
+    "request.timeout.ms": 30000,        # Max time to wait for a broker response before failing
```
aminvakil: From logs:
```
  uptime-results-1                         | %4|1778036634.157|SESSTMOUT|rdkafka#consumer-1| [thrd:main]: Consumer group session timed out (in join-state steady) after 30400 ms without a successful response from the group coordinator (broker 1001, last error was Success): revoking assignment and rejoining group
```
searching "Consumer group session timed out" shows a couple more kafka timeouts after 30s as well, maybe we should increase kafka timeout even further?

aldy505: @aminvakil holy sh*t you're right

aldy505:
image

I'm happy already, but let me rerun the test one more time

aldy505:
re-running it here: https://github.com/getsentry/self-hosted/actions/runs/26268331023

aldy505:
image

sigh, still flaky

sentry/sentry.conf.example.py
aminvakil: To be honest, I didn't catch this myself, it's been a while since I **also** use AI to review stuff :)

aldy505: But the problem persist. I have no idea other to blame github's arm64 runners.

aminvakil: Yeah, they are too slow, this happens to me as well in other projects.

I searched "Consumer group session timed out" in latest run:
```
  process-spans-1                            | %4|1779437870.346|SESSTMOUT|rdkafka#consumer-1| [thrd:main]: Consumer group session timed out (in join-state steady) after 31691 ms without a successful response from the group coordinator (broker -1, last error was Success): revoking assignment and rejoining group
```

Seems like they still have 30 seconds timeout.

aldy505: This is according to AI:

  Root cause summary:

   1. Silent config drop (Bug 1): KAFKA_CLUSTERS["default"] = DEFAULT_KAFKA_OPTIONS uses the legacy flat format. get_kafka_consumer_cluster_options() detects this and, because it's called with only_bootstrap=True, extracts only bootstrap.servers — all other settings including 
  session.timeout.ms and heartbeat.interval.ms are silently discarded.
   2. HACK override (Bug 2): Even if the settings survived, build_consumer_config in consumers/__init__.py unconditionally overrides session.timeout.ms to match max_poll_interval_ms when that value is < 45000ms. Since --max-poll-interval-ms defaults to 30000 in run.py, 
  session.timeout.ms is force-set to 30s — which is exactly what the ~31.7s timeout error reflects.

  Recommended fix (no code change required): Pass --max-poll-interval-ms 300000 (or any value ≥ 45000) to the sentry run consumer process-spans command in the self-hosted docker-compose.yml. This disables the HACK and allows the session timeout to stay at the safe default of
  45s.

Let me try to mess things up again

aminvakil: Seems like it wasn't fixed, right?

aldy505: It wasn't. But the first try works. I wonder if it's the cache, but I could be wrong

aldy505:
FYI I'm waiting for https://github.com/getsentry/snuba/pull/7961 to get merged

aldy505:
WOAH

aldy505:
Sigh... it is still flaky

aminvakil:
line 5134
```
  generic-metrics-consumer-1               | %4|1780382932.097|SESSTMOUT|rdkafka#consumer-1| [thrd:main]: Consumer group session timed out (in join-state steady) after 45290 ms without a successful response from the group coordinator (broker 1001, last error was Success): revoking assignment and rejoining group
```
It has timed out after 45s

aminvakil:
This is related to https://github.com/getsentry/sentry/pull/56219.

`session.timeout.ms` does not get set to 300 seconds right now.

https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md shows
`session.timeout.ms` is `45000` by default.

I don't know if it's a correct fix to change sentry, so you can raise `session.timeout.ms` to more than 45 seconds, I mean this seems like a XY problem.

aldy505:
@aminvakil so you're saying I have to increase `session.timeout.ms` as well?

aminvakil:
> @aminvakil so you're saying I have to increase `session.timeout.ms` as well?

Yes, currently it's capped to 45 seconds.

aldy505:
I'm going to split this PR into 2:
1. Introduce `SENTRY_KAFKA_MAX_POLL_INTERVAL_MS`
2. Put a `timeout-minutes` on CI

aldy505:
image

after 3 retries, it's not flaky as before

aminvakil (approved):
We're already changing `ulimits` to `262144` for clickhouse, changing it for `kafka` should not arise another problem for users.

Aside from that everything LGTM, thanks!

PR #4333, opened 2026-05-18
feat: optional chartcuterie service & profile

aldy505:
Chartcuterie has been a part of Sentry for a long time, and it's never a part of self-hosted package due to... no interest I guess?

I can see some (very little, tbh) self-hosted users are using Slack, so it's a perfect thing to add.


aldy505:
why do I have this on draft?

PR #4348, opened 2026-05-25
fix: prevent argument too long error on self-hosted error reports due to breadcrumbs too long

aldy505:
Previously some users reported that they are facing errors of 'docker argument list too long', it's annoying and hard to debug for us maintainers. This should prevent that.

### Legal Boilerplate

### DIFF
--- a/install/error-handling.sh
+++ b/install/error-handling.sh
@@ -19,6 +19,8 @@ generate_breadcrumb_json() {
   $jq -R -c '{"message": ., "category": "log", "level": "info"}'
 }
 
+# Create the breadcrumb payload now before stacktrace is printed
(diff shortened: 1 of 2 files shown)


install/error-handling.sh
```diff
@@ -182,7 +190,7 @@ cleanup() {
     # Create the breadcrumb payload now before stacktrace is printed
     # https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
     # Use sed to remove the last line, that is reported through the error message
```
aminvakil: I think we should move this comment to `collect_breadcrumbs` function now.

aminvakil (approved):

BYK (approved):

PR #4365, opened 2026-06-04
Adjust ClickHouse configuration to disable query_views_log

LordSimal:
image

As you can see in the screenshot above clickhouse doesn't cleanup its views_log tables automatically, see https://clickhouse.com/docs/operations/system-tables/query_views_log

> ClickHouse does not delete data from the table automatically. 

I don't think this is necessary in a "normal" self-hosted environment unless someone needs to debug clickhouse.

This of course does not truncate the existing tables, that maybe needs to be added to the install script.
@aldy505 what do you think?

### Legal Boilerplate

### DIFF
--- a/clickhouse/config.xml
+++ b/clickhouse/config.xml
@@ -7,6 +7,7 @@
     </logger>
     <query_thread_log remove="remove"/>
     <query_log remove="remove"/>
+    <query_views_log remove="remove"/>
     <text_log remove="remove"/>
     <trace_log remove="remove"/>
     <metric_log remove="remove"/>
--- a/install.sh
+++ b/install.sh
@@ -28,6 +28,7 @@ source install/check-minimum-requirements.sh
 # Upgrading clickhouse needs to come first before turning things off, since we need the old clickhouse image
 # in order to determine whether or not the clickhouse version needs to be upgraded.
 source install/upgrade-clickhouse.sh
+source install/cleanup-clickhouse.sh
 source install/update-docker-images.sh
 source install/turn-things-off.sh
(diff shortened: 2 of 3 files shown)


LordSimal:
The `_0` table was created with the https://github.com/getsentry/self-hosted/releases/tag/25.9.0 update and therefore a clickhouse version upgrade. Maybe a cleanup script should be added on install to remove all old, unnecessary clickhouse tables.

Primarily for me this is about `query_log_0` and `query_views_log_0`

aldy505:
Yeah sure, let's include some deletion queries on the install script. ClickHouse have lightweight deletes, so I think it'll be fine: https://clickhouse.com/docs/guides/developer/lightweight-delete

LordSimal:
To get this working/testing properly via `./install/cleanup-clickhouse.sh` i had to add
```
source install/dc-detect-version.sh
```
at the top of that file but after that it seems to have worked.
```
root@scarecrow:~/sentry# ./install/cleanup-clickhouse.sh
Cleaning up ClickHouse query log tables ...
▶ Initializing Docker|Podman Compose ...
docker compose --ansi never run --pull=never --rm

 Container sentry-self-hosted-clickhouse-1 Starting
 Container sentry-self-hosted-clickhouse-1 Started
 Container sentry-self-hosted-clickhouse-1 Waiting
 Container sentry-self-hosted-clickhouse-1 Healthy
Dropping system.query_log_0 ...
Dropping system.query_views_log_0 ...
Truncating system.query_log and system.query_views_log ...
```

aldy505 (approved):
Nice

PR #4374, opened 2026-06-09
feat(ci): run tests every 6 hours on the weekdays

aldy505:
Currently we have the tests running every 12 hours everyday, it's hard to keep track of issues that happen during the weekdays. Therefore, I separate the schedule to run on the weekdays for every 6 hours and on the weekends every 12 hours.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,9 @@ on:
       - "release/**"
   pull_request:
   schedule:
-    - cron: "0 0,12 * * *"
+    # Timezone is UTC by default
+    - cron: "0 0,6,12,18 * * 1-5" # weekdays
+    - cron: "0 0,12 * * 6,0" # weekends
 
 concurrency:
   group: ${{ github.ref_name || github.sha }}


aldy505:
Let me know if this doesn't make sense

aminvakil (changes requested):
I thought this was once every day :)

It was getting run every hour and now every 12 hours since https://github.com/getsentry/self-hosted/pull/2383.

How often has this caught a bug which has been fixed in 6 hours though?

This should be run every night IMHO.

aldy505:
> How often has this caught a bug which has been fixed in 6 hours though?

The last time sentry release breaks, usually Hubert, David Hertberth and I (and sometimes Burak) would brainstorm on Slack to figure out which commit might be causing it, and since the integration tests are within 12 hours interval, I need to test each commit one by one, which was a pain in the butt. With Hubert gone, I need some way to make it shorter.

aminvakil (approved):

Dav1dde (approved):

PR #4376, opened 2026-06-13
feat: set --max-poll-interval-ms flags to Sentry/Snuba consumers

aldy505:
Separated from https://github.com/getsentry/self-hosted/pull/4313/

Also introducint a healthcheck for snuba-replacer

### DIFF
--- a/.env
+++ b/.env
@@ -30,7 +30,7 @@ HEALTHCHECK_FILE_RETRIES=3
 HEALTHCHECK_FILE_START_PERIOD=600s
 # Set SETUP_JS_SDK_ASSETS to 1 to enable the setup of JS SDK assets
 # SETUP_JS_SDK_ASSETS=1
-#
+
 # Set SETUP_CUSTOM_CA_CERTIFICATE to 1 to generate trust stores that inject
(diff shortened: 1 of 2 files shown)


oioki (approved):

jap:
Hi, I was reviewing the changes before upgrading our sentry setup and had the same thought as the clanker in https://github.com/getsentry/self-hosted/pull/4376#pullrequestreview-4490980970 ; I was wondering why there is a fallback of 300_000 ms all over the place in the compose yaml, but the documentation says that the default is 30 seconds?

aldy505:
> Hi, I was reviewing the changes before upgrading our sentry setup and had the same thought as the clanker in [#4376 (review)](https://github.com/getsentry/self-hosted/pull/4376#pullrequestreview-4490980970) ; I was wondering why there is a fallback of 300_000 ms all over the place in the compose yaml, but the documentation says that the default is 30 seconds?

@jap Hey! I noticed that too, but I was thinking 30 seconds for most people is... too quick? Feel free to submit a PR to any change, whether update the documentation or adjust the default value.

PR #4403, opened 2026-07-09
feat: update feature flags through bash script

### LINKED ISSUE #2533: Upgrading self-hosted should automatically enable new features
New features are enabled by editing the `sentry.conf.example.py`. However, during the upgrade process, if a user already has a `sentry.conf.py` file, the new features won't be added by default. This is unpleasant and we should make this more smooth.

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/2533


install/update-feature-flags.sh
```diff
@@ -0,0 +1,55 @@
...
+    else
+      # Replace the existing SENTRY_FEATURES block with the canonical one.
+      # Find the start and end lines of the existing block.
+      _start_line=$(grep -n '^SENTRY_FEATURES' "$SENTRY_CONFIG_PY" | head -1 | cut -d: -f1)
+
+      # Find the closing `)`, the first line after _start_line that is exactly `)`.
+      _end_line=$(awk -v start="$_start_line" 'NR > start && /^\)$/ { print NR; exit }' "$SENTRY_CONFIG_PY")
+
+      if [[ -z "$_end_line" ]]; then
+        # No closing `)` found before EOF,  replace from start to EOF.
```
aminvakil, line 35:
```suggestion
        echo "Could not find end of SENTRY_FEATURES.update block; skipping automatic feature flag update."
```

install/update-feature-flags.sh
```diff
@@ -0,0 +1,55 @@
...
+        _end_line=$(wc -l <"$SENTRY_CONFIG_PY")
```
aminvakil, line 36:
```suggestion
        return 0
```

aminvakil (commented):
There is a corner case which this PR breaks self-hosted users flow.

Let's say a user has something like this for example:

```
SENTRY_FEATURES["projects:sample-events"] = False

GEOIP_PATH_MMDB = "/geoip/GeoLite2-City.mmdb"
CSRF_TRUSTED_ORIGINS = ["https://example.com"]
```
I mean when they do not have a `SENTRY_FEATURES.update` just after `SENTRY_FEATURES` for some reason, this PR with current commits would just remove everything the user has written afterwards (GEOIP, CSRF in this example) with `SENTRY_FEATURES.update`.

From a point of view, this may be ok for a user which has explictly stated `APPLY_AUTOMATIC_CONFIG_UPDATES=1`, but I still rather just skip and do not replace their whole script if `SENTRY_FEATURES.update` does not get found.

What do you think?

aldy505:
I'm not confident with my approach. If anyone want to continue the idea of this, feel free to do so.

PR #4419, opened 2026-07-17
fix: add accepted outcomes consumer

mzglinski:
Related to: https://github.com/getsentry/snuba/pull/7781

Without this, on v26.7.0 (maybe earlier, as I updated from 26.5.0), the stats & usage page does not show data for logs, application metrics, and dynamic sampling results.

### Legal Boilerplate

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -168,7 +168,8 @@ services:
       <<: *healthcheck_defaults
       # Using default user "postgres" from sentry/sentry.conf.example.py or value of POSTGRES_USER if provided
       test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-postgres}"]
-    command: ["postgres"]
+    command:
+      ["postgres"]
     environment:
       POSTGRES_HOST_AUTH_METHOD: "trust"
     volumes:
@@ -180,11 +181,7 @@ services:
     healthcheck:
       <<: *healthcheck_defaults
       # Using default user "postgres" from sentry/sentry.conf.example.py or value of POSTGRES_USER if provided
-      test:
-        [
-          "CMD-SHELL",
-          'psql -U postgres -p 5432 -h 127.0.0.1 -tA -c "select 1;" -d postgres >/dev/null',
-        ]
+      test: ["CMD-SHELL", "psql -U postgres -p 5432 -h 127.0.0.1 -tA -c \"select 1;\" -d postgres >/dev/null"]
     depends_on:
       postgres:
         <<: *depends_on-healthy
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -323,6 +326,11 @@ services:
     command: rust-consumer --storage outcomes_raw --consumer-group snuba-consumers --auto-offset-reset=earliest --max-batch-time-ms 750 --no-strict-offset-reset --health-check-file /tmp/health.txt --max-poll-interval-ms ${SENTRY_KAFKA_MAX_POLL_INTERVAL_MS:-300000}
     healthcheck:
       <<: *file_healthcheck_defaults
+  snuba-outcomes-accepted-consumer:
+    <<: *snuba_defaults
+    command: accepted-outcomes-consumer --storage eap_items --consumer-group accepted_outcomes --auto-offset-reset=earliest --max-batch-time-ms 750 --no-strict-offset-reset --accepted-outcomes-topic outcomes-billing --max-poll-interval-ms ${SENTRY_KAFKA_MAX_POLL_INTERVAL_MS:-300000}
+    profiles:
+      - feature-complete
```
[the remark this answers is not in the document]
mzglinski: Had to remove it, as this entrypoint does not support it.

aldy505 (approved):

docker-compose.yml
```diff
@@ -323,6 +323,11 @@ services:
     command: rust-consumer --storage outcomes_raw --consumer-group snuba-consumers --auto-offset-reset=earliest --max-batch-time-ms 750 --no-strict-offset-reset --health-check-file /tmp/health.txt --max-poll-interval-ms ${SENTRY_KAFKA_MAX_POLL_INTERVAL_MS:-300000}
     healthcheck:
       <<: *file_healthcheck_defaults
+  snuba-outcomes-accepted-consumer:
+    <<: *snuba_defaults
+    command: accepted-outcomes-consumer --storage eap_items --consumer-group accepted_outcomes --auto-offset-reset=earliest --max-batch-time-ms 750 --no-strict-offset-reset --accepted-outcomes-topic outcomes-billing --max-poll-interval-ms ${SENTRY_KAFKA_MAX_POLL_INTERVAL_MS:-300000}
```
aminvakil, line 328: Is 750 in `--max-batch-time-ms 750` correct?

mzglinski: I do not know the reasoning behind the proposed arguments in the PR I mentioned, which is why I decided to use the default of other consumers in self-hosted. This might need changing, but someone from Sentry would have to pitch in.

aminvakil: Yeah, some are 1000, I don't know which one is correct here.

mzglinski: I do not think 750 vs 1000 makes a big difference here, as the PR mentioned 120,000. Which is a lot, in my opinion, those are the types of nitpicks that you adjust when you start digging deeper into the data flow and the load on ClickHouse.

phacops: We always aim to insert at about once per second in all our production clusters but we have usually only one set of consumers inserting into a given cluster.

Now, in self-hosted, you usually insert to the same ClickHouse cluster so maybe setting  it to even higher is useful.

Higher means we insert less often into ClickHouse, meaning fewer parts created meaning less work (fewer merges to do) for ClickHouse, at the expense of having to wait a bit longer to see the data.

Back to the self-hosted use case, we have many consumers inserting into the same cluster, usually at low throughput per consumer, so I would probably set them to 30s at least to buffer as much data as possible and insert as little as possible (so less parts created meaning less work for ClickHouse). It means consumers will use more memory (since they buffer more) and data will come every 30s as opposed to every 1s so you'll see it a little later, but the system should be working less.

Make sure the max batch size is tuned accordingly so it hits the max batch time every time.

aldy505: Tried this on my instance, I set it to 2000 lol. With 750ms, the number of processed events are 30-90, but when I set it to 2000, it does not really affect memory much on high throughput.

aldy505:
@mzglinski do you wanna update anything before I merge?

phacops:
One more thing I might add: I don't know how useful this consumer is in self-hosted.

It's meant to derive outcomes from items pushed to EAP and push them to one of our billing topic. We use outcomes to know what to bill people and also to downsample queries occasionally.

I would think those 2 concerns do not exist in self-hosted which means we don't need this consumer.

mzglinski:
@aldy505 nothing on my side.

@phacops as mentioned in the description of the PR, without it the Stats & Usage page does not show values for certain data types. This might be a feature not needed by some, but for other (like me) that like to analyze the data it is :)

phacops:
> @phacops as mentioned in the description of the PR, without it the Stats & Usage page does not show values for certain data types. This might be a feature not needed by some, but for other (like me) that like to analyze the data it is :)

Makes sense, then that's definitely useful here.

PR #4441, opened 2026-07-31
Fix snuba-api healthcheck bypassing proxy for local requests

oddballfr:
Fix snuba-api healthcheck bypassing proxy for local requests (urllib NO_PROXY doesn't support CIDR ranges)

**Problem**

The snuba-api healthcheck (api_healthcheck.py) was failing in environments with an HTTP(S) proxy configured (HTTP_PROXY/HTTPS_PROXY). Squid logs confirmed the local request was being routed through the proxy and rejected:
TCP_DENIED/403 ... GET http://127.0.0.1:1218/health

**Root cause**

urllib.request does not support CIDR notation (127.0.0.0/8) in NO_PROXY — it only matches exact hostnames or domain suffixes. A NO_PROXY value that works for curl/wget can therefore still leave this local request routed through the configured proxy.

**Fix**

self-hosted/snuba/api_healthcheck.py now uses a dedicated urllib.request.build_opener(urllib.request.ProxyHandler({})) opener, explicitly forcing no proxy for this local request instead of relying on NO_PROXY parsing.

**Testing**

snuba-api healthcheck passes again after container rebuild, in an environment with HTTP_PROXY/HTTPS_PROXY configured.

**Legal Boilerplate**

### DIFF
--- a/snuba/api_healthcheck.py
+++ b/snuba/api_healthcheck.py
@@ -25,10 +25,17 @@ import urllib.request
 URL = os.environ.get("SNUBA_API_HEALTHCHECK_URL") or "http://127.0.0.1:1218/health"
 TIMEOUT = float(os.environ.get("SNUBA_API_HEALTHCHECK_TIMEOUT") or 2)
 
+# This is a local, in-container request. urllib's NO_PROXY matching doesn't
+# support CIDR ranges (e.g. 127.0.0.0/8), only exact hosts/domain suffixes,
+# so a NO_PROXY set up for other tools (curl, wget) can still leave this
+# request routed through HTTP(S)_PROXY. Force no proxy explicitly instead
+# of relying on NO_PROXY parsing.
+_NO_PROXY_OPENER = urllib.request.build_opener(urllib.request.ProxyHandler({}))
+
 
 def main() -> int:
     try:
-        body = urllib.request.urlopen(URL, timeout=TIMEOUT).read().decode()
+        body = _NO_PROXY_OPENER.open(URL, timeout=TIMEOUT).read().decode()
     except urllib.error.HTTPError as exc:
         print(f"snuba api returned HTTP {exc.code} from {URL}", file=sys.stderr)
         return 1


aldy505 (approved):

aminvakil (changes requested):
I think it's better to set NO_PROXY to 127.0.0.1 and there is no need for this change.

This should be documented in container scope proxy setting in docs though, current docs only say 127.0.0.0/8.

oddballfr:
Thanks for the review.
I actually tried adding 127.0.0.1 to NO_PROXY but the healthcheck kept failing.
This fix is what unblocked me, both on v25.5.2 (when the Python healthcheck was introduced) and again on v26.7.0.

aldy505 (approved):

aldy505:
@aminvakil are you okay if I merge this?

aminvakil:
@aldy505 Honestly I do not like this hack, and I've just tested and confirmed, setting `NO_PROXY` to `127.0.0.1` works, it's better to be mentioned in docs IMHO instead of hacking `urllib`.

```
$ docker compose exec snuba-api bash
snuba@a8a261436a93:/usr/src/snuba$ python api_healthcheck.py 
snuba@a8a261436a93:/usr/src/snuba$ echo $?
0
snuba@a8a261436a93:/usr/src/snuba$ export HTTP_PROXY="example.com:3128"
snuba@a8a261436a93:/usr/src/snuba$ python api_healthcheck.py 
snuba api unreachable at http://127.0.0.1:1218/health: [Errno 101] Network is unreachable
snuba@a8a261436a93:/usr/src/snuba$ export NO_PROXY="127.0.0.1"
snuba@a8a261436a93:/usr/src/snuba$ python api_healthcheck.py 
snuba@a8a261436a93:/usr/src/snuba$ echo $?
0
```

aldy505:
Coming back to this, why don't you set `SNUBA_API_HEALTHCHECK_URL` to `http://snuba:1218/health` instead? That would skip the proxy for `127.0.0.1`.

PR #4454, opened 2026-08-07
fix: stop the right containers for minimize-downtime flag

### LINKED ISSUE #4192: Reorder install.sh to update docker images before shutting down
### Problem Statement

The install.sh script shuts down Sentry before running the 'Fetching and updating docker images' step.

On slow internet connections, this can cause quite a long downtime for Sentry, when the script is doing something unrelated.

### Solution Brainstorm

Could the docker images be pulled before Sentry has been shutdown?

aldy505:
The `--minimize-downtime` should not remove all databases (postgres, kafka, clickhouse, etc) on `turn-things-off`. Tested this on my instance with `:nightly`. Works well as long as you don't have any wrongdoings on `docker-compose.override.yml` (learned this the hard way)

Closes https://github.com/getsentry/self-hosted/issues/4192

### DIFF
--- a/install/turn-things-off.sh
+++ b/install/turn-things-off.sh
@@ -1,8 +1,8 @@
 echo "${_group}Turning things off ..."
 
 if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
-  # Stop everything but relay and nginx
-  $dc rm -fsv $($dc config --services | grep -v -E '^(nginx|relay)$')
+  # Stop everything unless databases, relay, web, and nginx
+  $dc down -fsv $($dc config --services | grep -v -E '^(nginx|web|relay|smtp|memcached|redis|postgres|pgbouncer|kafka|clickhouse|seaweedfs)$')
 else
   # Clean up old stuff and ensure nothing is working while we install/update
   if [ "$CONTAINER_ENGINE" = "podman" ]; then
--- a/install/turn-things-off.sh
+++ b/install/turn-things-off.sh
@@ -2,7 +2,7 @@ echo "${_group}Turning things off ..."
 
 if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
   # Stop everything unless databases, relay, web, and nginx
-  $dc down -fsv $($dc config --services | grep -v -E '^(nginx|web|relay|smtp|memcached|redis|postgres|pgbouncer|kafka|clickhouse|seaweedfs)$')
+  $dc down $($dc config --services | grep -v -E '^(nginx|web|relay|smtp|memcached|redis|postgres|pgbouncer|kafka|clickhouse|seaweedfs)$')
 else
   # Clean up old stuff and ensure nothing is working while we install/update
   if [ "$CONTAINER_ENGINE" = "podman" ]; then
--- a/install/wrap-up.sh
+++ b/install/wrap-up.sh
@@ -2,13 +2,12 @@ if [[ "$MINIMIZE_DOWNTIME" ]]; then
   echo "${_group}Waiting for Sentry to start ..."
 
   # Start the whole setup, except nginx, relay, and web.
-  start_service_and_wait_ready --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay|web)$')
+  start_service_and_wait_ready --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay)$')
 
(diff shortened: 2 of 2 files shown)


aminvakil:
@aldy505 Would you please address issues by bots? They're right, `docker compose down` does not have `-f` flag nor `-s`.

aldy505:
> @aldy505 Would you please address issues by bots? They're right, `docker compose down` does not have `-f` flag nor `-s`.

I'm a bit... confused between the behavior of `docker compose rm` and `docker compose down` and `docker compose stop`. I would say `docker compose down` is the right  thing to do for that specific moment. Not sure what are those `-fse` flag even means.

aminvakil:
Force and stop are not necessary for `down`.
`down` always stops containers, and force stops them after a time anyway.

What's the reason for switching to `down` though here?

aldy505:
> Force and stop are not necessary for `down`. `down` always stops containers, and force stops them after a time anyway.
> 
> What's the reason for switching to `down` though here?

`down` is a modern command that I believe would cleanly remove the container, resulting in no orphan process. That's the only reason.

EDIT: oh right, and we need to reload the image tag, therefore dc stop and dc rm does not suffice

install/turn-things-off.sh
```diff
@@ -1,8 +1,8 @@
 echo "${_group}Turning things off ..."
 
 if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
-  # Stop everything but relay and nginx
-  $dc rm -fsv $($dc config --services | grep -v -E '^(nginx|relay)$')
+  # Stop everything unless databases, relay, web, and nginx
+  $dc down -fsv $($dc config --services | grep -v -E '^(nginx|web|relay|smtp|memcached|redis|postgres|pgbouncer|kafka|clickhouse|seaweedfs)$')
```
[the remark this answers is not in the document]
aldy505:
```suggestion
  $dc down $($dc config --services | grep -v -E '^(nginx|web|relay|smtp|memcached|redis|postgres|pgbouncer|kafka|clickhouse|seaweedfs)$')
```

install/wrap-up.sh
```diff
@@ -1,19 +1,14 @@
...
-  else
-    echo "Relay container not found, skipping restart."
-  fi
+  # Start the whole setup, except nginx, relay, and web.
+  start_service_and_wait_ready --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay|web)$')
 
   if [ -n "$($dc ps -q nginx 2>/dev/null)" ]; then
-    $dc exec -T nginx nginx -s reload || true
+    # relay and web will be restarted as well, if we restart relay.
+    $dc restart nginx
```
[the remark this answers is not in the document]
aminvakil: Yes, I agree with this comment.
What was the reason for this change?

install/wrap-up.sh
```diff
@@ -1,19 +1,14 @@
...
-  # Start the whole setup, except nginx and relay.
-  start_service_and_wait_ready --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay)$')
-
-  if [ -n "$($dc ps -q relay 2>/dev/null)" ]; then
-    $dc restart relay
-  else
-    echo "Relay container not found, skipping restart."
-  fi
+  # Start the whole setup, except nginx, relay, and web.
+  start_service_and_wait_ready --remove-orphans $($dc config --services | grep -v -E '^(nginx|relay|web)$')
```
aminvakil: Correct me if I'm wrong, but `web` container should always be recreated with new image after an upgrade, why does this has been added to exclusions?

install/turn-things-off.sh
```diff
@@ -1,8 +1,8 @@
 echo "${_group}Turning things off ..."
 
 if [[ -n "$MINIMIZE_DOWNTIME" ]]; then
-  # Stop everything but relay and nginx
-  $dc rm -fsv $($dc config --services | grep -v -E '^(nginx|relay)$')
+  # Stop everything unless databases, relay, web, and nginx
+  $dc down $($dc config --services | grep -v -E '^(nginx|web|relay|smtp|memcached|redis|postgres|pgbouncer|kafka|clickhouse|seaweedfs)$')
```
aminvakil, line 5: `podman` down, stops and removes dependencies like `postgres` as well.

aldy505: Sigh...

PR #4483, opened 2026-08-21
feat: bump clickhouse to 25.8

aldy505:
Snuba requires this now, see https://github.com/getsentry/snuba/pull/8359

### DIFF
--- a/install/upgrade-clickhouse.sh
+++ b/install/upgrade-clickhouse.sh
@@ -23,7 +23,6 @@ if $ps_command | grep -q clickhouse; then
     echo "Upgrading clickhouse to 25.8"
     $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:25.8.28.10001.altinitystable clickhouse
     start_service_and_wait_ready clickhouse
-    $dc down clickhouse
   else
     echo "Detected clickhouse version $version. Skipping upgrades!"
   fi
--- a/install/upgrade-clickhouse.sh
+++ b/install/upgrade-clickhouse.sh
@@ -14,7 +14,7 @@ if $ps_command | grep -q clickhouse; then
   # Start clickhouse if it is not already running
   start_service_and_wait_ready clickhouse
 
-  # In order to get to 25.3, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8 -> 24.8 -> 25.3
+  # In order to get to 25.8, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8 -> 24.8 -> 25.3 -> 25.8
   version=$($dc exec clickhouse clickhouse-client -q 'SELECT version()')
   if [[ "$version" == "25.3.6.10034.altinitystable" ]]; then
     echo "Detected clickhouse version $version"
--- a/install/upgrade-clickhouse.sh
+++ b/install/upgrade-clickhouse.sh
@@ -18,52 +18,12 @@ if $ps_command | grep -q clickhouse; then
   version=$($dc exec clickhouse clickhouse-client -q 'SELECT version()')
   if [[ "$version" == "25.3.6.10034.altinitystable" ]]; then
     echo "Detected clickhouse version $version"
-    start_service_and_wait_ready clickhouse
     $dc down clickhouse
 
     echo "Upgrading clickhouse to 25.8"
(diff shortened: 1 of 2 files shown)


install/upgrade-clickhouse.sh
```diff
@@ -16,7 +16,16 @@ if $ps_command | grep -q clickhouse; then
 
   # In order to get to 25.3, we need to first upgrade go from 21.8 -> 22.8 -> 23.3 -> 23.8 -> 24.8 -> 25.3
   version=$($dc exec clickhouse clickhouse-client -q 'SELECT version()')
-  if [[ "$version" == "21.8.13.1.altinitystable" || "$version" == "21.8.12.29.altinitydev.arm" ]]; then
+  if [[ "$version" == "25.3.6.10034.altinitystable clickhouse" ]]; then
+    echo "Detected clickhouse version $version"
+    start_service_and_wait_ready clickhouse
```
phacops: Do we need to start it before we down and upgrade it?

install/upgrade-clickhouse.sh
```diff
@@ -16,7 +16,16 @@ if $ps_command | grep -q clickhouse; then
...
+    $dc down clickhouse
+
+    echo "Upgrading clickhouse to 25.8"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:25.8.28.10001.altinitystable clickhouse
+    start_service_and_wait_ready clickhouse
+    $dc down clickhouse
+  elif [[ "$version" == "21.8.13.1.altinitystable" || "$version" == "21.8.12.29.altinitydev.arm" ]]; then
```
phacops: I think we should invest in figuring out if we're below the target version and upgrade to the latest. This logic is redundant.

install/upgrade-clickhouse.sh
aldy505: Not sure, that's a pattern since a very long time ago. I'd rather stick with it.

install/upgrade-clickhouse.sh
aldy505: I'm thinking of removing it, as people would follow the hard stop anyway, and recently 26.7.0 is a hard stop (https://develop.sentry.dev/self-hosted/releases/#hard-stops)

install/upgrade-clickhouse.sh
```diff
@@ -14,47 +14,16 @@ if $ps_command | grep -q clickhouse; then
...
+  if [[ "$version" == "25.3.6.10034.altinitystable" ]]; then
     echo "Detected clickhouse version $version"
     $dc down clickhouse
 
-    echo "Upgrading clickhouse to 22.8"
-    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:22.8.15.25.altinitystable clickhouse
+    echo "Upgrading clickhouse to 25.8"
+    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:25.8.28.10001.altinitystable clickhouse
     start_service_and_wait_ready clickhouse
     $dc down clickhouse
```
phacops: We don't want this, I think we want ClickHouse to stay up after the upgrade.

aldy505: Why?

phacops: Don't we want ClickHouse to stay up so the service can run? The previous upgrade loop would not down it at the end.

install/upgrade-clickhouse.sh
```diff
@@ -14,47 +14,16 @@ if $ps_command | grep -q clickhouse; then
...
-    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:24.8.14.10459.altinitystable clickhouse
-    start_service_and_wait_ready clickhouse
-    $dc down clickhouse
-
-    echo "Upgrading clickhouse to 25.3"
-    $dcb $build_arg BASE_IMAGE=altinity/clickhouse-server:25.3.6.10034.altinitystable clickhouse
-    start_service_and_wait_ready clickhouse
   else
     echo "Detected clickhouse version $version. Skipping upgrades!"
   fi
```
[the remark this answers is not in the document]
aminvakil: This is tolerable in this project.

install/upgrade-clickhouse.sh
aminvakil: Exactly.

phacops (approved):

PR #4487, opened 2026-08-24
feat: Remove organizations:incidents flag

RudraPatel2003:
Cleanup that will be done after these 3 steps

Fixes ISWF-3270

image

### DIFF
--- a/sentry/sentry.conf.example.py
+++ b/sentry/sentry.conf.example.py
@@ -366,7 +366,6 @@ SENTRY_FEATURES.update(
             "organizations:discover",
             "organizations:global-views",
             "organizations:issue-views",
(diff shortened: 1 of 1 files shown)


aldy505:
@RudraPatel2003 Hi, can you provide any justification or why you're removing this flag?

RudraPatel2003:
The flag gates metric monitors, which are now available even to free-tier users. The dIscussion on removing this flag was started in the following PR

The flag checks have all been removed from the sentry codebase already. I am in the process of now removing the flag itself

getsentry/getsentry#20511

saponifi3d (approved):
the bot is generally correct -- but once we merge the open PRs this should be fine.

PR #4489, opened 2026-08-25
feat: install script to check for hard stops

aldy505:
The idea came from Discord, and Alex (stayalive) lay out a very good approach on this: https://discord.com/channels/621778831602221064/796028405833007104/1541789134006259814


aldy505:
@aminvakil apparently, `exit 0` terminates the install script. do you have any other suggestion for this? I want to avoid hadouken pattern.

EDIT; nevermind I just read this https://github.com/getsentry/self-hosted/pull/4489#discussion_r3854073595, sorry about that

BYK:
Just dropping this here: https://github.com/getsentry/self-hosted/pull/3878#pullrequestreview-3201466823

Overall, I think this is pretty good but we need a shared place and format to have the hard-stops so docs and the repo does not diverge. I'd propose JSON in a well-known location like the docs repo or this repo, and reading it with `jq` in the install script

aldy505:
> Just dropping this here: [#3878 (review)](https://github.com/getsentry/self-hosted/pull/3878#pullrequestreview-3201466823)
> 
> Overall, I think this is pretty good but we need a shared place and format to have the hard-stops so docs and the repo does not diverge. I'd propose JSON in a well-known location like the docs repo or this repo, and reading it with `jq` in the install script

@BYK Yeah I remember about your comment. I don't know where the "shared  place" should be. I don't want to have it on `sentry`, because I don't want to pull any image first, just for doing this. At the end of the day, I would think having these two separate would be good.

And yes, there will always be a maintenance burden for this one.

BYK:
>I don't want to have it on sentry, because I don't want to pull any image first, just for doing this

Not sure we are on the same page. My proposal is having this information in a separate, dedicated JSON file in _this_ repo and then the docs repo fetching it at build time.

aldy505:
> > I don't want to have it on sentry, because I don't want to pull any image first, just for doing this
> 
> Not sure we are on the same page. My proposal is having this information in a separate, dedicated JSON file in _this_ repo and then the docs repo fetching it at build time.

Aaaaaahhhhhh, that makes sense. I dunno how to do it on the docs repo, but making a JSON file here would be doable.

mhkarimi1383:
I think it's better to be on the Sentry's main source code in database migration part, not in install scripts, but it's better than nothing :)

aldy505:
> I think it's better to be on the Sentry's main source code in database migration part, not in install scripts, but it's better than nothing :)

Sentry's main source code does not have the concept of "hard stop", as they're always deploying code from the "master" branch every hour.

mhkarimi1383:
But hard stop should be handled in other installation methods (like k8s)

install/check-hard-stop.sh
```diff
@@ -0,0 +1,206 @@
...
+  else
+    echo ""
+  fi
+}
+
+# Compare two calver versions
+# Usage: compare_calver "1.2.3" "1.2.4"
+# Returns: -1 if first < second, 0 if equal, 1 if first > second
+#
+# This bit is written by Claude Haiku 4.5.
```
BYK: Haiku 4.5? What is this, the stone age? Use Luna at least 🤣

aminvakil:
> But hard stop should be handled in other installation methods (like k8s)

There is no other installation methods supported, k8s support is outside of scope of getsentry.

Also you should really read release notes before upgrading :wink:

install/check-hard-stop.sh
```diff
@@ -0,0 +1,206 @@
...
+  declare current_version=""
+  if [[ -f "$latest_version_file" ]]; then
+    current_version=$(cat "$latest_version_file")
+  fi
+
+  # We perform some checks if the `current_version` is not empty.
+  if [[ -n "$current_version" ]]; then
+    # We iterate over the list of hard stops, and check whether the current
+    # version is below any of them.
+    local _wrote_version=0
```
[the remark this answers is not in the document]
aminvakil: Correct.

install/check-hard-stop.sh
```diff
@@ -0,0 +1,206 @@
...
+    for hard_stop in "${hard_stops[@]}"; do
+      compare_result=$(compare_calver "$current_version" "$hard_stop")
```
aminvakil: I'd suggest using existing `vergte` function, also cursor and sentry bots about returning exit statuses are correct.

install/check-hard-stop.sh
```diff
@@ -0,0 +1,206 @@
...
+      if [[ "$compare_result" == 0 ]]; then
```
aminvakil: I'm not sure if I understood this correctly, but let's say we are going from 26.5.0 to 26.8.0 and 26.7.0 is a hard-stop.

Doesn't this break outside of loop as soon as it sees 26.6.0 which is fine and not a hard-stop? never reaching 26.7.0?

aminvakil (changes requested):
Would you please address bots comments and mention me again afterwards?

jamgregory:
> Also you should really read release notes before upgrading 😉

I'll just point out that I had read the release notes during our more recent upgrade and had still missed that 26.7.0 was a hard stop. It would still be preferable for the script to alert you to this, just in case you accidentally miss it.

PR #4497, opened 2026-08-28
Migrate fast-revert workflow to GitHub App token

joshuarli:
Update fast-revert to use actions/create-github-app-token with the shared GETSENTRY_FAST_REVERT_BOT credentials.

### DIFF
--- a/.github/workflows/fast-revert.yml
+++ b/.github/workflows/fast-revert.yml
@@ -10,28 +10,36 @@ on:
         required: true
         description: '`name <email>` for triggering user'
 
-# disable all permissions -- we use the PAT's permissions instead
+# Disable the default token permissions; the GitHub App token below supplies access.
 permissions: {}
 
 jobs:
   revert:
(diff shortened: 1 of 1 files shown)


.github/workflows/fast-revert.yml
```diff
@@ -10,28 +10,36 @@ on:
...
     runs-on: ubuntu-latest
+    environment: fast-revert
     if: |
       github.event_name == 'workflow_dispatch' || github.event.label.name == 'Trigger: Revert'
     steps:
+      - name: fast-revert bot token
+        id: fast-revert-bot-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          client-id: ${{ vars.GETSENTRY_FAST_REVERT_BOT_CLIENT_ID }}
```
joshuarli, line 27: need to set this up in here

PR #4498, opened 2026-08-31
feat(nginx): add an opt-in patch that re-resolves the upstream addresses

spawnia:
Adds an opt-in patch that makes nginx re-resolve `relay` and `web`, under `optional-modifications/patches/` next to `external-kafka`. Nothing changes for anyone who does not apply it.

@aminvakil, this is the shape you offered to review in https://github.com/getsentry/self-hosted/pull/4295#issuecomment-4325891462:

> I agree that this could be an opt-in behaviour, but I'll be happy to review your PR regarding that, but I think in current scope of project and `nginx.conf` being tracked by git, you should probably write a install script which checks an env or something like that

A patch avoids the install-script env check you disliked, and it is how this repository already handles opt-in edits to git-tracked files.

#4492 changed the default configuration and you rejected that. It stays closed. Its branch now carries a cleaned-up single commit for anyone who wants the same change without the patch indirection.

## Scope, corrected

I previously argued on #4295 that #3914 does not close the gap because our own outage came from a host-level Docker restart. That was wrong, and I have corrected it there. Our `docker inspect` evidence shows a plain `docker compose down` / `up -d`, which is exactly the case `restart: true` covers. We hit it only because our pinned version predates #3914.

So this patch is not for the case #3914 already handles. It is for address changes that no Compose operation causes: a crash restart, or a Docker daemon restart. The README says so, so a reader can tell whether it applies to them.

We also took an unclean VM hard reset on unpatched nginx and the failure did **not** reproduce, because a reboot starts containers rather than recreating them. I am not claiming evidence for the remaining gap beyond the mechanism.

## Answering the earlier objections

- **DNS cost.** `valid=10s` costs about six queries per minute per upstream against the engine's embedded DNS. A per-upstream timer drives it, so worker count and request volume do not affect it. Measured with 32 workers: 14 A queries in 70 s, zero AAAA.
- **Hardcoding `127.0.0.11`.** Dropped. It is Docker-only; Podman answers on the network gateway. The entrypoint reads the address from the container's own `/etc/resolv.conf` instead.
- **Zone size.** `512k`, not `64k`. nginx requires at least eight system pages, so `64k` refuses to start on 64 KiB-page hosts such as RHEL 9 on arm64.
- **Overriding `nginx.conf`.** The resolver lands in a separate include, so the `/etc/nginx/nginx.conf` mount target is unchanged and existing overrides keep working.

## Test plan

Applied against this branch on current `master`: all three hunks apply with no `.rej`, `docker compose config` passes, and the entrypoint renders `resolver <address> ipv6=off valid=10s` and `nginx -t` succeeds.

Behaviour verified on Docker with the patch applied: baseline 200; recreating `relay` onto a new address recovers in 4 s without touching nginx; `nginx -s reload` keeps serving; an upstream absent at startup yields 502 until it appears rather than refusing to start; two nameservers render and validate; no AAAA queries and no unreachable-peer errors.

Not tested on Podman. The patch reads the resolver from the container rather than assuming Docker's address, so it should work, but I have no Podman host to confirm on.

Credit to @ilias-115, whose #4295 established the mechanism.

### DIFF
--- a/optional-modifications/patches/resolve-upstreams/README.md
+++ b/optional-modifications/patches/resolve-upstreams/README.md
@@ -1,38 +1,17 @@
 # resolve-upstreams
 
-Make nginx re-resolve the `relay` and `web` container addresses instead of caching them for the
-life of a worker process.
+nginx resolves the names in an `upstream` block once, at startup, and keeps the result for the life
+of a worker. When a container is recreated onto a different address, nginx keeps using the old one
+and every request fails until nginx is restarted.
 
-## The problem
+The UI failing is obvious; ingest failing is not. `relay` serves `/api/store/`, `/api/<id>/` and
+`/api/0/relays/`, so events are dropped while the UI keeps answering and nothing reports an error.
 
-nginx resolves the names in an `upstream` block once, at startup, and keeps the result forever. If
-a container is recreated and comes back on a different address, nginx keeps sending traffic to the
-old one. Every request fails until nginx itself is restarted.
+This patch adds `resolve` to both upstreams, the `zone` that `resolve` requires, and a `resolver`
+directive rendered at container start from the container's own `/etc/resolv.conf` — the engine DNS
+address is `127.0.0.11` on Docker but the network gateway on Podman, so it is not hardcoded.
 
-The UI failing is obvious. Ingest failing is not: `relay` serves the `/api/store/`, `/api/<id>/`
-and `/api/0/relays/` routes, so events are dropped while the UI keeps working normally and nothing
-in the stack reports an error.
-
-`docker compose restart relay` alone does not trigger this — the address is usually reused. It
-takes a recreate (`docker compose up -d --force-recreate`, an image upgrade, a host reboot with a
-different container start order, or anything else that lets another container claim the old
-address) for the addresses to actually move.
-
-## What the patch changes
-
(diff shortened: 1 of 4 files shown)


aminvakil (changes requested):
Someone should maintain the patch to be correct, I don't want something unmaintainable with little value to be shipped here.

I still think this should be documented for users who want to try this and patch it locally without keeping it in self-hosted.

It's totally fine to have diff with self-hosted locally, I have `access_log off;` in `nginx.conf` and `maxmemory 16gb` in `redis.conf` for years, previously I had a custom `docker-compose-override.yml` for specific needs, I had `kafka` custom configurations locally.

spawnia:
Thanks for the feedback. I was planning on cleaning up the description and code here before marking ready for review btw, but seems like the notification reached you already.

From what the repo establishes, I thought patches were the way to go for such customizations. They do seem to strike a nice middle ground between pure docs prose and the complexity and maintenance weight of a built-in solution. Yes, it can break - but so can docs. At least with the patch, it is documented very precisely and able to be verified somewhat mechanically. Are you open to continuing work on this approach, or do think we should leave it here?

Agree on the local diff. We do maintain a few files that differ from upstream already and it has worked fine, but copying over the full nginx.conf seemed too much for me - which is why I tried to upstream a fix. Getting a second or third opinion on something I tried locally was already valuable for me, so I thank you for that.

PR #4510, opened 2026-09-14
fix(install): resolve geoip database paths relative to the repo, not cwd

### LINKED ISSUE #2239: Geoip file not found during installation
cp: cannot stat 'geoip/GeoLite2-City.mmdb.empty': No such file or directory

https://self-hosted.getsentry.net/organizations/self-hosted/issues/416/?project=3&query=is%3Aunresolved&referrer=issue-stream&stream_index=6

aguspatur22:
install/geoip.sh is documented (develop.sentry.dev/self-hosted/geolocation) as a standalone command people run later on their own to refresh the geolocation database, which in practice usually means a cron job. It computes script_dir from BASH_SOURCE at the top, but only uses that to find _detect-container-engine.sh. The actual mmdb/conf paths and the geoipupdate volume mount stay plain paths relative to whatever the caller's cwd happens to be.

When install.sh sources this from the repo root that's fine, but a standalone invocation from elsewhere (a crontab entry doesn't run with the repo as cwd unless it cds there first) hits the reported error on the very first line: cp: cannot stat 'geoip/GeoLite2-City.mmdb.empty': No such file or directory. The file has been in the repo since 2020; the script just isn't looking for it in the right place when invoked from outside the repo root.

Anchored mmdb, conf, and the geoipupdate volume mount to repo_root ($script_dir/..) instead of the bare relative paths.

Reproduced the reported error by running install/geoip.sh via absolute path from an unrelated directory, confirmed the fix resolves it in the same repro, then added a case to the existing _unit-test/geoip-test.sh that does the same from /tmp so it doesn't regress. Ran the full existing test file (all cases pass) and shfmt -d against both changed files (no diff, matches the repo's formatting).

Fixes #2239.


aminvakil (approved):

aldy505 (approved):

aldy505:
This PR looks like it was written by AI, but well, acceptable changes.

aminvakil:
@aldy505 Exactly!

For this I do not care if it's written by AI or not.
This is a good change, it improves the codebase.
Also I do not care for three comments in unit tests, that's fine.

